PR #60, opened 2016-01-13
Fix error messages

jougs:
This cleans up the wording of some error messages.


DimitriPlotnikov:
This code doesn't compile. Class names must comply to artifact names.

PR #126, opened 2016-04-12
fix sympy version check

Omniphage:
LooseVersion(sympy.\__version__) evaluates to LooseVersion("1.0") rather than LooseVersion("1.0.0") causing sympy check to fail when it should not.


DimitriPlotnikov:
What was the actual Problem?

PR #130, opened 2016-04-17
Invalid/Valid model pairs for every coco + integration into tests

Omniphage:
Why does aliasHasNoSetter mark its negative findings as warnings instead of errors?  (Kinda want to change it to make tests look more uniform)

Regard commit message ( tests re-enabled / removed ). And maybe double check that I did nothing wrong there.


DimitriPlotnikov:
There are conflicts in the pull request.

aliasHasNoSetter: a midterm plan is to compute a reverse function automatically. Otherwise, in many cases the modeler just wants to use aliases as a "shortcut" for a reusable part of expression. Thus, it makes no sense to force her or him to write a setter. But a control code should be put into the generated implemntation that indicates that setting a value to an alias without the corresponding setter doesn't work.

PR #133, opened 2016-04-22
Merge upstream and fix build error

Omniphage:
Updated pull request. Should work now.


DimitriPlotnikov:
Restructure models for context conditions. Each context condition has a positive and negative model.

PR #233, opened 2016-08-12
Rework expression type checking

Omniphage:
-switched to pseudo AG scheme for expression type calculation
-installed coco to assert presence of unit type information in literals, where required.


DimitriPlotnikov:
CI fails, please fix.

DimitriPlotnikov:
Tests sitll fail...

PR #245, opened 2016-08-31
Pretty Unit names and ODE Typechecking

Omniphage:
-Bugfixes/Cleanups in Unit name printing.
-actually use pretty names.
-general cleanup
-Refactor Expression type calculation (relocate to AST node)
-typechecking for ODEDeclarations


models/iaf_neuron.nestml
```diff
@@ -80,20 +80,20 @@ neuron iaf_neuron_nestml:
 
   # Declarative description of the updates rules
   equations:
-      shape G = (e/tau_syn) * t * exp(-1/tau_syn*t)
+      shape G = 1pA * (e/tau_syn) * t * exp(-1/tau_syn*t)
```
DimitriPlotnikov: The '1pA *' term is not really elegant, even it is probably provides some useful information. From the user perspective it is also not intuitive. I assume, at this stage you need the information, that the result of the function is a 'current'. Since a shape can only be used in an I_sum (current based case) or I_cond (conductance based case) you can get this information just by looking where the shape was used. C.f. NESTMLSymbolTableCreator#markConductanceBasedBuffers

PR #289, opened 2016-10-29
Refactor docker creation and run script

jougs:
This PR removes the individual scripts to create and run the docker container and adds a new script that can provision and run the container.

The script `nestml_docker.sh` takes a command (`provision` or `run`) and additional arguments. `provision` can take an optional argument `--dev` to provision the development container. If this is not given, the release container is created. `run` also takes the `--dev` argument, in which case it runs the corresponding container. Else the release container is run. Additionally, it takes one or more directories with NESTML models.

@DimitriPlotnikov, can you please test and merge? Thanks!


DimitriPlotnikov:
Hello @jougs,

I try to test your script, but it seems just to print the command and not to evaluate it. E.g. 
`sh nestml_docker.sh provision`

Outputs only the command it is going to evaluate, but not more:
```
Creating docker image 'nestml_release'
docker build -t nestml_release -f DockerfileRelease
```
Comparing the line 78 in nestml_docker.sh which is responsible for this output, I find only `echo $cmd` command. Can it be that it only prints the command and not to evaluate it?

docker/nestml_docker.sh
```diff
@@ -75,7 +75,7 @@ case $mode in
 	cmd="docker build -t $container -f $dockerfile"
```
DimitriPlotnikov: @jougs 

```
nestml@nestml:~/software/nestml/docker$ ./nestml_docker.sh run ../models/

Running docker image 'nestml_release'
Error: Unable to access jarfile /data/nestml/target/nestml.jar

nestml@nestml:~/software/nestml/docker$ ./nestml_docker.sh run --dev ../models/

Running docker image 'nestml_development'
Error: Unable to access jarfile /data/nestml.jar
```
Entrypoint script should either distinguish between development and release versions correclty or the nestml.jar should be stored at the same place (which can be done in corresponding  Dockerfile-artifacts)

jougs:
I've fixed the things from your comments. Sorry for not having tested again before pushing.

Your comment above seems to be cut off. Can you please check?

DimitriPlotnikov:
Works for me.

PR #317, opened 2016-12-06
minor fixes and changes to expression logic wrt units

PTraeder:
Make expression errors conform to format used in splErrorStrings

Better organization of testcases for units:
 - split up tests
 -added/improved postive tests

Warnings/Errors involving Units in expressions:
- Unit + primitive = Unit ; warning
- UnitA+UnitB = real ; warning

- true?Unit:primitive = primitive ; warning
- true?UnitA:UnitB = real ; warning

Grammar changes: 
-Allow declarations with unitless numerator, i.e:
         InverseVoltVariable 1/V = 10 [1/V]
-Implement CoCo to make sure that numerator literal is only ever '1'


Silmathoron:
Hi @PTraeder, could you provide a description of the rationale behind the modifications you propose?
I must say I'm not convinced by all I've seen, but since the NESTML workshop starts tomorrow, it would be a great opportunity to get the opinion of many users on the issues you tried to tackle.
I think it would also be a good time to discuss what should be treated as error and what should come out as a warning...
For instance, I would be in favor of treating all inhomogeneities as errors, since they are usually the sign of fundamental errors in the model.

DimitriPlotnikov:
@Silmathoron Yes you are generally right. However, since we are not fully convinced of the top quality of the unitsystem, we interpret unit errors, which don't lead to C++ compilation errors, as warnings. As soon as I become sure the unitframework is stable, I'll a) switch warning to errors. b) make a parameter in nestml tool that the user can define, how he would like see these issues (warning or error). And I'm not going to do it before tomorrow.

PTraeder:
Hi @Silmathoron. 

@DimitriPlotnikov was faster than me on this. But you hinted at other objections with my modifications? Can you elaborate? I know my commits have been a bit clumped in the past, but I am working on that.

DimitriPlotnikov:
@Silmathoron Could I calm you concerns?

Silmathoron:
Sorry for the delay,
actually, thinking about it a bit longer, since you want to switch it to Error/User choice afterwards, I think the choices made by @PTraeder are best:

* choosing real for conflicting units prevents arbitrary choice
* choosing unit for a combination of dimensionned and dimensionless parameters

This prevents the propagation of reals throughout the whole system while preventing obvious mistakes.

Issue #324, opened 2016-12-07
Explicit units for "sum" functions


Silmathoron:
Currently, there are some problems with ``_sum`` functions in NESTML:
*  ``spikes.get_sum()`` delivers ``pA``, whereas it should deliver ``real``
*  ``cond_sum`` seems not to deliver ``nS`` (I don't know what it does)
* neither does ``curr_sum()`` seem to deliver ``pA``

Maybe I should be more explicit about the first problem:
* we use normalization functions for the alpha model to make sure the peak is at ``1 pA`` times the "synaptic weight" (dimensionless)
* for ``_exp`` models, the dimensionless property can either be handled by the Unit+primitive=unit property of #317 or if we treat dimension mismatches as errors, the user will have to specify by hand ``g += 1 nS * spikes.get_sum()``

DimitriPlotnikov:
@PTraeder Can you check it?

PTraeder:
Fixed in #328

PR #328, opened 2016-12-08
Address #324 and Implement compound assignments

PTraeder:
Compound assignments construct a dummy '=' assignment to check instead.


src/main/java/org/nest/symboltable/predefined/PredefinedFunctions.java
```diff
@@ -134,13 +134,13 @@
     final MethodSymbol i_sum = createFunctionSymbol(CURR_SUM);
     i_sum.addParameterType(getType("pA"));
     i_sum.addParameterType(getBufferType());
-    i_sum.setReturnType(getType("pA"));
+    i_sum.setReturnType(getType("nS"));
```
Silmathoron: why is this ``nS``? Shouldn't this be real?

src/main/java/org/nest/symboltable/predefined/PredefinedFunctions.java
```diff
@@ -134,13 +134,13 @@
...
     name2FunctionSymbol.put(CURR_SUM, i_sum);
 
     final MethodSymbol cond_sum = createFunctionSymbol(COND_SUM);
     cond_sum.addParameterType(getType("nS"));
     cond_sum.addParameterType(getBufferType());
-    cond_sum.setReturnType(getType("pA"));
+    cond_sum.setReturnType(getType("nS"));
```
Silmathoron: See above

PTraeder: in #324 you stated:
cond_sum seems not to deliver nS (I don't know what it does)
neither does curr_sum() seem to deliver pA

from wich i deduced that you expected the curr and cond sum functions to return nS. 
Is the expected behaviour for both of those then to return real?

Silmathoron: Wow, you're right ! (partially)
I'm tired...
``CURR_SUM`` should return ``pA`` and ``COND_SUM`` should return ``nS``

PTraeder: Im having problems with some (generation) tests after changing it again. Not exactly sure where those come from. Might need a bit to fix.

Silmathoron: Which models are concerned? After my PR, the ``aeif_cond_alpha`` and ``aeif_cond_exp`` do not work because of the ``min`` function which breaks ``sympy``, so the fact that these models donot work is "normal" (read "needs fixing" ^^)

PTraeder: Yep those are concerned. I am just going to ignore those for the time being then.

DimitriPlotnikov:
@PTraeder 
I'm going to remove `get_sum` function. What is the type of the buffer itself? It should be also real. @Silmathoron isn't it?

Silmathoron:
Spike buffers should be ``real``, current buffers should be ``pA``

    spikesInh <- inhibitory spike [real]
    spikesExc <- excitatory spike [real]
    currents  <- current [pA]

DimitriPlotnikov:
@PTraeder @Silmathoron's proposal seems to be reasonable for me. It is already implemented that way? Are then model fixed regarding units?

PTraeder:
This doesnt seem all too trivial. I am going to look into implementing it now.

DimitriPlotnikov:
@PTraeder any updates?

PTraeder:
@DimitriPlotnikov 
Sorry for the delay, I had to take a bit of a hiatus. This changes the expression types generated from buffer variables according to @Silmathoron proposal. This implies that bufferType is not set in any expression anymore.

PTraeder:
Basically I am in need of some specifications here:
Do we want buffers to appear as real/pA in arithmetic expressions only, and otherwise as bufferType to protect predefined functions?

DimitriPlotnikov:
@PTraeder 
If I understand everything correctly, then current buffers remain `pA`. Spike buffers are just `real`. In order to keep type checking, all occurrences of of spike buffers must be multiplied either with `pA` (psc models) or `nS` (cond) in cases when it is necessary. @Silmathoron Did I get your idea right?

Silmathoron:
@DimitriPlotnikov yes, that's the idea!
@PTraeder could you not add a secondary type? They would all be ``bufferType`` and have a specific subtype (either ``real`` or ``pA``)

DimitriPlotnikov:
@Silmathoron 
Technically, it is not necessary to add a second type. It is a buffer symbol and it has information that is current or spike buffer as part of the symbol.

Silmathoron:
Ok, then what was the problem @PTraeder mentioned?

PTraeder:
My problem was that my changes might have done damage to @DimitriPlotnikov conception of the type system. Turns out it is still a workable solution. PR updated to merge conflicts.

src/test/java/org/nest/nestml/_symboltable/NESTMLSymbolTableCreatorTest.java
```diff
@@ -209,18 +209,18 @@ public void testPredefinedMethods() throws IOException {
         = resolveMethod("exp", Lists.newArrayList("real"), modelScope);
     assertTrue(withPredicate.isPresent());
 
-    Optional<MethodSymbol> I_sumOptional = resolveMethod(PredefinedFunctions.CURR_SUM, Lists.newArrayList("pA", "Buffer"), modelScope);
+    Optional<MethodSymbol> I_sumOptional = resolveMethod(PredefinedFunctions.CURR_SUM, Lists.newArrayList("pA", "real"), modelScope);
     assertTrue(I_sumOptional.isPresent());
     I_sumOptional = resolveMethod(PredefinedFunctions.CURR_SUM, Lists.newArrayList("nS", "Buffer"), modelScope);
```
Silmathoron: Why are there ``nS`` related to the ``CURR_SUM`` function?

src/test/java/org/nest/nestml/_symboltable/NESTMLSymbolTableCreatorTest.java
```diff
@@ -209,18 +209,18 @@ public void testPredefinedMethods() throws IOException {
...
     assertFalse(I_sumOptional.isPresent());
-    I_sumOptional = resolveMethod(PredefinedFunctions.CURR_SUM, Lists.newArrayList("pA", "real"), modelScope);
+    I_sumOptional = resolveMethod(PredefinedFunctions.CURR_SUM, Lists.newArrayList("pA", "Buffer"), modelScope);
     assertFalse(I_sumOptional.isPresent());
 
-    Optional<MethodSymbol> Cond_sumOptional = resolveMethod(PredefinedFunctions.COND_SUM, Lists.newArrayList("nS", "Buffer"), modelScope);
+    Optional<MethodSymbol> Cond_sumOptional = resolveMethod(PredefinedFunctions.COND_SUM, Lists.newArrayList("nS", "real"), modelScope);
     assertTrue(Cond_sumOptional.isPresent());
     Cond_sumOptional = resolveMethod(PredefinedFunctions.COND_SUM, Lists.newArrayList("pA", "Buffer"), modelScope);
```
Silmathoron: Why are there ``pA`` associated to the ``COND_SUM`` function?

src/main/java/org/nest/symboltable/predefined/PredefinedFunctions.java
```diff
@@ -147,13 +147,13 @@
 
     final MethodSymbol i_sum = createFunctionSymbol(CURR_SUM);
     i_sum.addParameterType(getType("pA"));
-    i_sum.addParameterType(getBufferType());
+    i_sum.addParameterType(getRealType());
```
Silmathoron: I'm quite out of my league here, so it's just a question to try to understand how the inner functions work: I thought you kept the ``BufferType``, so why are the functions now associated with ``RealType``?

DimitriPlotnikov: We check this with an extra different coco. It must be not only `BufferType`, but also a special type of the expression.

src/test/java/org/nest/nestml/_symboltable/NESTMLSymbolTableCreatorTest.java
PTraeder: Those are negative tests that are designed to fail (assertFalse)

src/test/java/org/nest/nestml/_symboltable/NESTMLSymbolTableCreatorTest.java
PTraeder: See above

src/main/java/org/nest/symboltable/predefined/PredefinedFunctions.java
Silmathoron: Ok, thanks!

DimitriPlotnikov:
@PTraeder Is that ready to review?

PTraeder:
Models folder seems fine. Tested using testModelsFolder test from NestmlFrontendIntegrationTest.
Only differences introduced by the patch are reduced warnings where expected types now match up with buffer types and the wording of warnings that were already present.

PR #329, opened 2016-12-09
Names corrections + work on Izhikevitch

Silmathoron:
* Normalized variables names to match ``nest::names`` (``Delta_T`` and ``I_syn_ex/in``)
* Corrected one spike buffer in ``hh_psc_alpha_neuron``
* Changed the Izhikevitch current units to ``pA``
* Added detailed and dimensionned ``izhikevitch_psc_alpha`` model.

### DIFF
--- a/models/izhikevich_psc_alpha.nestml
+++ b/models/izhikevich_psc_alpha.nestml
@@ -1,0 +1,115 @@
+/*
+Name: izhikevich_psc_alpha - Detailed Izhikevich neuron model with alpha-shaped
+                             post-synaptic current.
+
+Description:
+Implementation of the simple spiking neuron model introduced by Izhikevich
+[1]. The dynamics are given by:
+   C_m dV_m/dt = k (V-V_t)(V-V_t) - u + I + I_syn_ex + I_syn_in
+   dU_m/dt = a*(b*(V_m-E_L) - U_m)
+
+   if v >= V_th:
+     V_m is set to c
+     U_m is incremented by d
+
+   On each spike arrival, the membrane potential feels an alpha-shaped current
+   of the form:
+     I_syn = I_0 * t * exp(-t/tau_syn) / tau_syn.
+
+References:
+[1] Izhikevich, Simple Model of Spiking Neurons,
+IEEE Transactions on Neural Networks (2003) 14:1569-1572
+
+Sends: SpikeEvent
+
+Receives: SpikeEvent, CurrentEvent, DataLoggingRequest
+FirstVersion: 2009
+Author: Hanuschkin, Morrison, Kunkel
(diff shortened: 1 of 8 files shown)


models/izhikevich_psc_alpha.nestml
```diff
@@ -0,0 +1,115 @@
...
+      r -= 1
+    elif V_m >= V_peak:
+      V_m = c
+      U_m += d
+      emit_spike()
+      r = RefractoryCounts
+    end
+
+    I_syn_ex += spikesExc
+    I_syn_in += spikesInh
```
DimitriPlotnikov: Should be not I_syn_ex' += spikesExc*PSConInit_E?

Silmathoron: Very true! I see the neurocomputational part is starting to get to you ;)
I'll fix that quickly

DimitriPlotnikov: Just pattern matching:)

DimitriPlotnikov:
@Silmathoron 
I'm going to merge your PR. However, in the next step I'll replace the `max` function through predefined a `bound_max()` (do you have a better name?), since otherwise the symbolic analysis is not possible. During code generation, it will be mapped to `std::max`.  The advantage of the new function is that I can define the semantic of the function call (first argument is the variable to be bound, the second argument is the upper bound). The same for the `min` function.

I will also add a context condition that forbids the usage of the `min`, `max` in the `equation` block.

That test are still working is just an accident, actually, the `NestFrontendIntegrationTest` should fail (and it fails locally at my machine). It will be fixed either.

Silmathoron:
@DimitriPlotnikov I have no particular opinion on this...
I must say I liked the idea we talked about of bounding state variables through their declaration in the ``state`` block, with something like ``bounds [V_min, V_max], for instance, but using a function in ``equation`` would do, though I would probably advise to use ``maximum`` and ``minimum`` rather than ``bound_max``.

DimitriPlotnikov:
@Silmathoron 
Me either. I'm going to implement a pragmatic approach, s.t. all models in `models` are executable again. And then, I'll start a new issue regarding factoring out invariants and putting bounds where now

I don't like `maximum` because it would be confusing to me to distinguish between `max` and `maximum`.

Silmathoron:
> I don't like maximum because it would be confusing to me to distinguish between max and maximum.

I don't understand...

DimitriPlotnikov:
I need two different functions. A general purpose `max` and a special `bounded_max`. The `max` is not differentiable and cannot be used in ode block.

Silmathoron:
Why do you still need ``max``?
My point was just that ``maximum`` is more intuitive than ``bound_max`` for people...

PR #342, opened 2017-01-20
Created Sympy_Framework for evolving an ODE

ingablundell:
Added files that calculate the linear homogeneous ODE that a given function satisfies, calculates the propagator matrix and with that the stepping function for a given linear constant coefficient ODE with shapes of these function.

Misc.py contains a class "ShapeFunction" that assigns a linear homogeneous ODE to a function, that satisfies such an ODE, also it's initial values, the order of the equation, and the factor in the equation, e.g for I''=aI+ bI', the factors are [a,b]

ode_to_prop_matrix calculates the propagator matrix for a given linear constant coefficient ODE plus a list of shapes occurring in the ODE. Also it calculates the stepping function.

integration_test, tests the script from ODE plus shapes to stepping fuction for different shapes.

As reviewers I suggest @jougs @Silmathoron!


Silmathoron:
Hi Inga,
thanks for the work, it looks nice.

I skimmed through it, just to get a rough idea of what it's doing.
I'm not sure I'll have time to review it this week-end, but I'll try to have it under way next week.

Just to give me a bit of context:

* could you detail where this is supposed to be used in NESTML? (is it restricted to the spikes' shapes?)
* could you explain how things were done before and what has changed? (were you restricted to exp, delta and alpha before?) The pre-existing .ftl files look a bit cryptic to me ^^

ingablundell:
Hi Tanguy,
thanks!

- This is used only for equations of the form : V'=C*V+C_2*I where I is a (sum of) convolution of 'shapes', i.e. functions that satisfy a linear homogeneous ODE.
- It is supposed to differ from the preexisting code in the sense that it: now works for any 'shape' that satisfies such an ODE (before some cases were problematic), works for multiple different 'shape' functions in one ODE (e.g. I could contain a sum of convolutions for the alpha and the exp shape), before this was only possible for the same function with different parameters, is clearer and hopefully more understandable, and works in components that anyone could use independent of NESTML.

Thanks again!

Silmathoron:
Hi @ingablundell, sorry, the email address I'm using for GitHub has been blocked, so I did not receive the notification. I've been quite busy lately, so I probably would not have had time to check this anyway.
However, it's getting better now, so I hope I'll be able to have a look at this by the end of the week, or next week, at worse.
Cheers

ingablundell:
@Silmathoron Great, thanks!

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+            # natural numbers (differing in each row)
+            X = zeros(order)
+            
+            # `Y` will contain the derivatives of `order` of the natural
+            # number in the corresponding row of `X`
+            Y = zeros(order, 1)
+            
+            # It is possible that by choosing certain natural numbers,
+            # the system of equations will not be solvable, i.e. `X` is
+            # not invertable. This is unlikely but we check for
```
Silmathoron: invert**i**ble, and not invert**a**ble

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+            if not invertible:
+                raise Exception("Failed to find homogeneous linear ODE "
+                                "which shape satisfies, or shape does "
+                                "not satisfy any such ODE of order <= {}".format(MAX_ORDER))
+            
+            derivative_factors = X.inv() * Y
+            diff_rhs_lhs = 0
+            # Once we have 'derivative_factors' the
+         
+            # We calculated derivative_factors of the linear
```
Silmathoron: This should come before the previous explanation:

> If $f^{(order)}(t) = \sum_{i<order} C_i f^{(i)}(t)$, then we can find the
> $order-1$ derivative factors by evaluating the previous equation as a
> linear system of order $order-1$ such that Y = [f^{(order)}] = mat_X * [C_i]
> hence [C_i] can be found by inverting mat_X.

Or something like that (but the ipynb would be best I think)

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+        self.nestml_ode_form = "{} = {}".format(lhs, rhs)
+        self.derivative_factors = list(simplify(derivative_factors))
+        self.initial_values = [x.subs(t, 0) for x in derivatives[:-1]]
+
+
+
+class ShapeODE(object):
+    '''It is also possible to define a shape function by giving a
+    set of linear homogeneous ODEs which the shape functions satisfies 
+    and a set of initia values (for an ODE of order `order`, `order-1` many.)
```
Silmathoron: initial

src/main/resources/org/nest/sympy/integration_test.py
```diff
@@ -0,0 +1,42 @@
...
+        shape_alpha = misc.ShapeFunction("shape_alpha", "e / tau * t * exp(-t / tau)")
+        self.assertEqual(2, shape_alpha.order)
+
+        shape_exp = misc.ShapeFunction("shape_exp", "exp(-t/tau)")
+        self.assertEqual(1, shape_exp.order)
+
+        shape_1 = misc.ShapeFunction("shape_1",  "e / tau * t * t * exp(-t / tau)")
+        self.assertEqual(3, shape_1.order)
+        # shape_2 has order 4
+        # shape_2 = misc.ShapeFunction("shape_2",  "e / tau * t * t *t * exp(-t / tau)")
```
jougs: Why is this code commented out?

src/main/resources/org/nest/sympy/integration_test.py
```diff
@@ -0,0 +1,42 @@
...
+        shape_exp = misc.ShapeFunction("shape_exp", "exp(-t/tau)")
+        shape_sin = misc.ShapeFunction("shape_sinb", "sin(t)")
+        shapes = [shape_alpha, shape_exp, shape_sin]
+
+        ode_var = "V_m"
+        ode_rhs = "-1/Tau * V_m-1/C * (shape_alpha + shape_exp + shape_sin + currents + I_E)"
+        prop_matrices, const_term, step_const = otpm.ode_to_prop_matrices(shapes, ode_var, ode_rhs)
+
+        pm, ps = otpm.prop_matrix_to_prop_step(prop_matrices, const_term, step_const, shapes, ode_var)
+        # checks that the computation doesn't fail!
```
jougs: Which code does this comment refer to? Should it go above the code line above it?

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
+from sympy.parsing.sympy_parser import parse_expr
+from sympy import *
+
+from sympy.matrices import zeros
+
+# Define constants
+MAX_ORDER = 10
```
jougs: As far as I understood, it is anyway very unlikely to need 10 here. An explanatory comment should be added, though. The same applies to the line below.

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Convert the postsynaptic response shape from a function form  
+       shape = t * exp(t * a)
+    to an ODE form
+       shape'' = -a**2 * shape + 2*a * shape'
+    '''
+
+    def __init__(self, name, function):
+
+        self.name = parse_expr(name)
```
jougs: This will give us a SymPy symbol, which we need later on when differentiating. A corresponding comment should be added.

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Example:
+    ========
+
+    Convert the postsynaptic response shape from a function form  
+       shape = t * exp(t * a)
+    to an ODE form
+       shape'' = -a**2 * shape + 2*a * shape'
+    '''
+
```
jougs: I find this comment a bit strange here. The comment here should explain what the class `ShapeFunction` is about. The comment above should rather be a docstring for the module, i.e. go to the top of the file.

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+        # found_ode is true if we find a linear homogeneous ODE that
+        # `shape` satisfies
+        found_ode = False
+        
+        # First we check if `shape` satisfies a linear homogeneous ODE
+        # of order 1. `derivatives` is a list of all derivatives of `shape`
+        # up to the order we are checking (which we just call 'order')
+        derivatives = [shape, diff(shape, t)]
+        
+        # If `diff_rhs_lhs`, here shape'-derivative_factors*shape=0 for
```
jougs: I agree about splitting the comments.

src/main/resources/org/nest/sympy/ode_to_prop_matrix.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+
+
+    
+
+    
+
+    
+
```
jougs: Please remove all but one empty lines.

src/main/resources/org/nest/sympy/integration_test.py
ingablundell: I think, the reason was that I thought it didn't add any more information by multiplying with another $t$ (see shape alpha and shape 1). But then it makes more sense to delete it.

src/main/resources/org/nest/sympy/integration_test.py
ingablundell: I don't think so. I think it must be leftover from something. I will delete it.

src/main/resources/org/nest/sympy/misc.py
ingablundell: I honestly don't think it is necessary to be allowed to have anything above 10. That would just be crazy. But I will add comments.

src/main/resources/org/nest/sympy/misc.py
ingablundell: I'm not sure what you mean. What would an explanation of what the ShapeFunciton is about be?

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+        # linear homogeneous ODE (and does for t=1). Thereafter, `diff_rhs_lhs will` be the
+        # difference of the derivative of shape of order 'order' and
+        # the sum of all lower derivatives times their
+        # 'derivative_factors which is a list of the potential
+        # factors in the ODE from the factor of shape^(0) to
+        # shape^(order-1).
+        # In the case of the ODE of order 1 we have only one `derivative_factor`
+        # but in all other cases we have several. For unified handling we define
+        # `derivative_factors` as a list for all orders (also for order 1)
+        derivative_factors = (1/derivatives[0] * derivatives[1]).subs(t,1),
```
ingablundell: Yes you're right. Here we should try for other ts in the case that t=1 does not work.

src/main/resources/org/nest/sympy/misc.py
```diff
@@ -0,0 +1,178 @@
...
+        # order.
+        while not found_ode and order < MAX_ORDER:
+
+            # The potential order must be at least `order+1`
+            order += 1
+            # Add the next higher derivative to the list of
+            # derivatives of `shape`
+            derivatives.append(diff(derivatives[-1], t))
+            
+            # The goal here is to calculate the factors (which we call
```
ingablundell: I don't think an ipython notebook is particularly helpful in this instance but maybe something else, @jougs?

Silmathoron: You can display real LaTeX math with the notebook, which would allow you to write the system that supports your text explanation/code... otherwise it's really hard to follow.

src/main/resources/org/nest/sympy/misc.py
ingablundell: The comment seems to be inseparable to me. I don't know how to explain the one without the other.

ingablundell:
@Silmathoron @jougs I (hope) I have now addressed all your concerns. It would be great if you could have another quick look. Thank you!

src/main/resources/org/nest/sympy/prop_matrix.py
```diff
@@ -0,0 +1,142 @@
...
+    ========
+    shape_alpha = ShapeFunction("shape_alpha", "e / tau * t * exp(-t / tau)")
+    shape_exp = ShapeFunction("shape_exp", "exp(-t/tau)")
+    shape_sin = ShapeFunction("shape_sinb", "sin(t)")
+    shapes = [shape_alpha, shape_exp, shape_sin]
+
+    ode_var = "V_m"
+    ode_rhs = "-1/Tau * V_m-1/C * (shape_alpha + shape_exp + shape_sin + currents + I_E)"
+    prop_matrices, const_term, step_const = otpm.ode_to_prop_matrices(shapes, ode_var, ode_rhs)
+    '''
```
jougs: Very nice!

src/main/resources/org/nest/sympy/prop_matrix.py
```diff
@@ -2,19 +2,30 @@
...
-     for any given linear constant coefficient ODE with an inhomogeneous part
-     wich is a sum of shapes (These satisfy a linear hoomogeneous ODE). 
-     The idea is to reformulate the ODE or system of ODEs as the ODE
-     y'=Ay and to calculate A and then exp(Ah) to give an evolution of 
-     the system for a given timestep `h`.
-     Although we think of the equation as something like this:
+    '''The function `ode_to_prop_matrices` calculates a so called proagator
+    for any given linear constant coefficient ODE with an inhomogeneous part
+    wich is a sum of `shapes` (instaces of the class `ShapeFunction` or
+    ShapeODE; they satisfy a linear hoomogeneous ODE).
```
Silmathoron: `` `ShapeODE` `` + ho(**o**)mogeneous

src/main/resources/org/nest/sympy/shapes.py
```diff
@@ -1,34 +1,107 @@
...
-    '''The aim of this script is to check for any given function of
-    positive `t`, which we will call `shape`, if it satisfies a linear
-    homogeneous ODE. If it does, it will give it's order, initial
-    values, the equation itself and a list of all factors of
-    derivatives (which we will call `derivative_factors`) of the shape
-    function in the ODE, i.e. if shape''= a*shape+b*shape' the list
-    will be: [a,b].
-
+    '''Here we provide a class, `ShapeFunction` that can be called
+    with a chosen name of the shape and it's mathematical discription;
```
Silmathoron: chosen shape name and (**its**) mathematical d**e**scription`

src/main/resources/org/nest/sympy/integration_test.py
```diff
@@ -1,42 +1,38 @@
...
         self.assertEqual(2, shape_3.order)
 
-        shape_4 = misc.ShapeFunction("shape_4", "cos(t)")
+        shape_4 = ShapeFunction("shape_4", "cos(t)")
         self.assertEqual(2, shape_4.order)
 
-        shape_5 = misc.ShapeFunction("shape_5", "sin(t)")
-        self.assertEqual(2, shape_5.order)
+        shape_5 = ShapeFunction("shape_5", "sin(t)")
+        self.assertEqual(2, shape_5.order)
```
Silmathoron: Remove trailing whitespaces

src/main/resources/org/nest/sympy/shapes.py
```diff
@@ -155,13 +241,16 @@ def __init__(self, name, function):
...
+    defined with the name of the shape (i.e a function of `t`
+    that satisfies a certain ODE), the variables on the left handside
+    of the ODE system, the right handsides of the ODE systems
+    and the initial value of the function.
+        
+    Canonical calculation of the properties, `order`, `name`,
+    `initial_values` and the system of ODEs in matrix form are made.
     '''
-    
+
```
Silmathoron: Remove spaces

ingablundell:
@Silmathoron @jougs Thank you for your corrections, they have been addressed in my last commit.

PR #343, opened 2017-01-30
Model review

PTraeder:
Clear warnings in models folder. Some bugfixes. 
Some warnings in generation pipeline still exist because Units are not yet printed correctly for generation.

### DIFF
--- a/models/aeif_cond_exp_implicit.nestml
+++ b/models/aeif_cond_exp_implicit.nestml
@@ -115,9 +115,9 @@ neuron aeif_cond_exp_implicit:
       emit_spike()
     end
 
-    g_ex += spikeExc.get_sum()
-    g_in += spikeInh.get_sum()
-    I_stim = currents.get_sum()
+    g_ex += spikeExc.get_sum() * 1 nS
+    g_in += spikeInh.get_sum() * 1 nS
+    I_stim = currents.get_sum() * 1 pA
   end
 
 end
--- a/models/aeif_cond_exp_implicit.nestml
+++ b/models/aeif_cond_exp_implicit.nestml
@@ -115,9 +115,9 @@ neuron aeif_cond_exp_implicit:
       emit_spike()
     end
 
-    g_ex += spikeExc.get_sum() * 1 nS
-    g_in += spikeInh.get_sum() * 1 nS
-    I_stim = currents.get_sum() * 1 pA
+    g_ex += spikeExc
+    g_in += spikeInh
+    I_stim = currents
   end
 
 end
--- a/models/aeif_cond_exp_implicit.nestml
+++ b/models/aeif_cond_exp_implicit.nestml
@@ -115,8 +115,8 @@ neuron aeif_cond_exp_implicit:
       emit_spike()
     end
 
-    g_ex += spikeExc
-    g_in += spikeInh
+    g_ex += spikeExc * 1 nS
+    g_in += spikeInh * 1 nS
     I_stim = currents
   end
 
--- a/models/hh_cond_exp_traub.nestml
+++ b/models/hh_cond_exp_traub.nestml
@@ -37,12 +37,12 @@ neuron hh_cond_exp_traub_neuron:
     V_m mV = E_L #  Membrane potential
 
     # equilibrium values for (in)activation variables
-    function alpha_n_init real = 0.032 * ( 15. - V_m / 1 mV) / ( exp( ( 15. - V_m / 1 mV ) / 5. ) - 1. )
-    function beta_n_init real = 0.5 * exp( ( 10. - V_m / 1 mV) / 40. )
-    function alpha_m_init real = 0.32 * ( 13. - V_m / 1 mV) / ( exp( ( 13. - V_m / 1 mV) / 4. ) - 1. )
-    function beta_m_init real = 0.28 * ( V_m /1 mV  - 40. ) / ( exp( ( V_m / 1 mV - 40. ) / 5. ) - 1. )
-    function alpha_h_init real = 0.128 * exp( ( 17. - V_m / 1 mV) / 18. )
-    function beta_h_init real = 4. / ( 1. + exp( ( 40. - V_m / 1 mV ) / 5. ) )
+    function alpha_n_init 1/ms = 0.032/(1 ms* 1 mV ) * ( 15. mV - V_m) / ( exp( ( 15. mV - V_m) / 5. mV ) - 1. )
+    function beta_n_init 1/ms = 0.5 /1 ms * exp( ( 10. mV - V_m ) / 40. mV )
+    function alpha_m_init 1/ms = 0.32/(1 ms* 1 mV ) * ( 13. mV - V_m) / ( exp( ( 13. mV - V_m) / 4. mV ) - 1. )
+    function beta_m_init 1/ms = 0.28/(1 ms* 1 mV ) * ( V_m  - 40. mV ) / ( exp( ( V_m - 40. mV ) / 5. mV ) - 1. )
+    function alpha_h_init 1/ms = 0.128/1 ms * exp( ( 17. mV - V_m) / 18. mV )
+    function beta_h_init 1/ms = ( 4. / ( 1. + exp( ( 40. mV - V_m ) / 5. mV) ) ) / 1 ms
 
     Act_m real =  alpha_m_init / ( alpha_m_init + beta_m_init )
     Act_h real = alpha_h_init / ( alpha_h_init + beta_h_init )
@@ -67,16 +67,16 @@ neuron hh_cond_exp_traub_neuron:
 
     # channel dynamics
     function V_rel mV = V_m - V_T
-    function alpha_n real = 0.032 * ( 15. - V_rel / 1 mV ) / ( exp( ( 15. - V_rel / 1 mV ) / 5. ) - 1. )
-    function beta_n real = 0.5 * exp( ( 10. - V_rel / 1 mV) / 40. )
-    function alpha_m real = 0.32 * ( 13. - V_rel / 1 mV ) / ( exp( ( 13. - V_rel / 1 mV ) / 4. ) - 1. )
-    function beta_m real = 0.28 * ( V_rel / 1 mV - 40. ) / ( exp( ( V_rel / 1 mV - 40. ) / 5. ) - 1. )
-    function alpha_h real = 0.128 * exp( ( 17. - V_rel / 1 mV ) / 18. )
-    function beta_h real = 4. / ( 1. + exp( ( 40. - V_rel / 1 mV ) / 5. ) )
+    function alpha_n 1/ms = 0.032/(1 ms* 1 mV ) * ( 15. mV - V_m) / ( exp( ( 15. mV - V_m) / 5. mV ) - 1. )
+    function beta_n 1/ms = 0.5 /1 ms * exp( ( 10. mV - V_m ) / 40. mV )
+    function alpha_m 1/ms = 0.32/(1 ms* 1 mV ) * ( 13. mV - V_m) / ( exp( ( 13. mV - V_m) / 4. mV ) - 1. )
+    function beta_m 1/ms = 0.28/(1 ms* 1 mV ) * ( V_m  - 40. mV ) / ( exp( ( V_m - 40. mV ) / 5. mV ) - 1. )
+    function alpha_h 1/ms = 0.128/1 ms * exp( ( 17. mV - V_m) / 18. mV )
+    function beta_h 1/ms = ( 4. / ( 1. + exp( ( 40. mV - V_m ) / 5. mV) ) ) / 1 ms
 
-    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) / 1 ms
-    Act_h' = ( alpha_h - ( alpha_h + beta_h ) * Act_h ) / 1 ms
-    Inact_n' = ( alpha_n - ( alpha_n + beta_n ) * Inact_n ) / 1 ms
+    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) 
+    Act_h' = ( alpha_h - ( alpha_h + beta_h ) * Act_h ) 
+    Inact_n' = ( alpha_n - ( alpha_n + beta_n ) * Inact_n )
 
   end
 
--- a/models/hh_cond_exp_traub.nestml
+++ b/models/hh_cond_exp_traub.nestml
@@ -37,12 +37,12 @@ neuron hh_cond_exp_traub_neuron:
     V_m mV = E_L #  Membrane potential
 
     # equilibrium values for (in)activation variables
-    function alpha_n_init real = 0.032 * ( 15. - V_m ) / ( exp( ( 15. - V_m ) / 5. ) - 1. )
-    function beta_n_init real = 0.5 * exp( ( 10. - V_m ) / 40. )
-    function alpha_m_init real = 0.32 * ( 13. - V_m ) / ( exp( ( 13. - V_m ) / 4. ) - 1. )
-    function beta_m_init real = 0.28 * ( V_m - 40. ) / ( exp( ( V_m - 40. ) / 5. ) - 1. )
-    function alpha_h_init real = 0.128 * exp( ( 17. - V_m ) / 18. )
-    function beta_h_init real = 4. / ( 1. + exp( ( 40. - V_m ) / 5. ) )
+    function alpha_n_init real = 0.032 * ( 15. - V_m / 1 mV) / ( exp( ( 15. - V_m / 1 mV ) / 5. ) - 1. )
+    function beta_n_init real = 0.5 * exp( ( 10. - V_m / 1 mV) / 40. )
+    function alpha_m_init real = 0.32 * ( 13. - V_m / 1 mV) / ( exp( ( 13. - V_m / 1 mV) / 4. ) - 1. )
+    function beta_m_init real = 0.28 * ( V_m /1 mV  - 40. ) / ( exp( ( V_m / 1 mV - 40. ) / 5. ) - 1. )
+    function alpha_h_init real = 0.128 * exp( ( 17. - V_m / 1 mV) / 18. )
+    function beta_h_init real = 4. / ( 1. + exp( ( 40. - V_m / 1 mV ) / 5. ) )
 
     Act_m real =  alpha_m_init / ( alpha_m_init + beta_m_init )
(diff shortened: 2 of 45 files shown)


DimitriPlotnikov:
@PTraeder please, provide a more informative PR description.

@jougs and @Silmathoron are kindly invited to review the PR.

Silmathoron:
Hi there, sorry, the email address I'm using for GitHub has been blocked, so I did not receive the notification... I've been quite busy lately, but I hope I'll be able to have a look at this by the end of the week, or next week, at worse.

models/hh_cond_exp_traub.nestml
```diff
@@ -74,9 +74,9 @@ neuron hh_cond_exp_traub_neuron:
     function alpha_h real = 0.128 * exp( ( 17. - V_rel / 1 mV ) / 18. )
     function beta_h real = 4. / ( 1. + exp( ( 40. - V_rel / 1 mV ) / 5. ) )
 
-    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) * 1 [1/s]
-    Act_h' = ( alpha_h - ( alpha_h + beta_h ) * Act_h ) * 1 [1/s]
-    Inact_n' = ( alpha_n - ( alpha_n + beta_n ) * Inact_n ) * 1 [1/s]
+    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) / 1 ms
```
Silmathoron: [this is a memo] I need to check where this unit comes from

models/hh_psc_alpha_implicit.nestml
```diff
@@ -66,10 +66,10 @@ neuron hh_psc_alpha_implicit:
   equations:
     # synapses: alpha functions
     I_syn_ex'' = -I_syn_ex' / tau_syn_ex
-    I_syn_ex' = I_syn_ex' - ( I_syn_ex / tau_syn_ex ) * 1 [ms/s]
+    I_syn_ex' = I_syn_ex' - ( I_syn_ex / tau_syn_ex )
```
Silmathoron: [memo2] this is a highly disturbing way of writing this ODE, there has to be another way!

models/ht_neuron.nestml
```diff
@@ -99,7 +99,7 @@ neuron ht_neuron_nestml:
     recordable function I_h pA = -h_g_peak * Ih_m  * ( V_m - h_E_rev )
     # The spike current is only activate immediately after a spike.
     function I_spike mV = (g_spike) ? -( V_m - E_K ) / Tau_spike : 0
-    V_m'  = ( ( I_Na * 1 pA + I_K * 1 pA + I_syn + I_NaP + I_KNa + I_T + I_h + I_stim ) / Tau_m + I_spike * 1 [pA/(ms * mV)] ) * 1 [s/F]
+    V_m'  = ( ( I_Na * 1 pA + I_K * 1 pA + I_syn + I_NaP + I_KNa + I_T + I_h + I_stim ) / Tau_m + I_spike * 1 [pA/(ms * mV)] ) * 1 [Gs/F]
```
Silmathoron: How did this giga second get there? also the extraneous spaces should be removed

models/iaf_psc_delta.nestml
```diff
@@ -124,8 +124,8 @@ neuron iaf_psc_delta_neuron:
       # for decay until end of refractory period
       if with_refr_input:
         refr_spikes_buffer += spikes * exp(-r * h / tau_m) * 1 mV
-      else:
-        spikes.get_sum() # clear buffer entry, ignore spike
+      #else:
```
Silmathoron: is this not required? I think this should be the default behavior...

PTraeder: With the removal of get_sum(), I am not sure what NESTML construct would be used to clear the buffer. Hence my commenteing out the old, dysfunct code, pending review.

models/ht_neuron.nestml
PTraeder: I seem to have accidentally deleted my comment:
could have as well been s/nF, I merely dumped the necessary factor into time here.
LHS: mV/ms = V/s
RHS: (pA/ms)*(Gs/F) = (pA/ms) * (s/nF) = A/F = V/s

Silmathoron: RHS = mV/ms
I = [pA] = dq/dt = d(CV)/dt = C dV / dt = [pF] * [mV / ms], i.e. your missing factor is just **1pF**

PTraeder: Note that on the RHS we have pA/ms not just pA. Hence the missing factor is ms/pF = s/nF

jougs: @heplesser: in your role as the original author of the `ht_neuron` in NEST, can you please have a look at the use of units in the NESTML version? I have a feeling that some of the `/1ms` or `/1mV` that get rid of the units are not actually needed but stem from wrong units for some of the variables in the first place.

models/hh_cond_exp_traub_implicit.nestml
```diff
@@ -64,16 +64,16 @@ neuron hh_cond_exp_traub_implicit:
...
+    function beta_m real = 0.28 * ( V_rel / 1 mV - 40. ) / ( exp( ( V_rel / 1 mV - 40. ) / 5. ) - 1. )
+    function alpha_h real = 0.128 * exp( ( 17. - V_rel / 1 mV ) / 18. )
+    function beta_h real = 4. / ( 1. + exp( ( 40. - V_rel / 1 mV ) / 5. ) )
 
-    Act_m' = alpha_m - ( alpha_m + beta_m ) * Act_m
-    Act_h' = alpha_h - ( alpha_h + beta_h ) * Act_h
-    Inact_n' = alpha_n - ( alpha_n + beta_n ) * Inact_n
+    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) * 1 [1/ms]
+    Act_h' = ( alpha_h - ( alpha_h + beta_h ) * Act_h ) * 1 [1/ms]
+    Inact_n' = ( alpha_n - ( alpha_n + beta_n ) * Inact_n ) * 1 [1/ms]
```
heplesser: Similar to above, but I find the notation here more confusing; `/ 1 ms` would probably be easier, but I think no unit manipulation should happen here at all.

models/ht_neuron.nestml
```diff
@@ -99,21 +99,20 @@ neuron ht_neuron_nestml:
     recordable function I_h pA = -h_g_peak * Ih_m  * ( V_m - h_E_rev )
     # The spike current is only activate immediately after a spike.
     function I_spike mV = (g_spike) ? -( V_m - E_K ) / Tau_spike : 0
-    V_m'  = ( I_Na + I_K + I_syn + I_NaP + I_KNa + I_T + I_h + I_stim ) / Tau_m + I_spike
-
+    V_m'  = ( ( I_Na * 1 pA + I_K * 1 pA + I_syn + I_NaP + I_KNa + I_T + I_h + I_stim ) / Tau_m + I_spike * 1 [pA/(ms * mV)] ) * 1 [s/nF]
```
heplesser: Why are some currents multiplied with units and others not? This seems to point to inconsistencies and errors elsewhere. All currents should "arrive" here with proper units, including I_spike. I would also suggest that, for easier comparison with the equation in the paper, we should have `I_spike/tau_spike` here, instead of incorporating the time constant in `I_spike`; as it is now, `I_spike` looks like a current but isn't, and that is going to cause misunderstandings later.

Note one challenge with the Hill-Tononi model: In that model, all conductances are dimensionless, which also means that currents are measured in mV. Especially, the `1 s/nF` term is not needed.

models/ht_neuron.nestml
```diff
@@ -99,21 +99,20 @@ neuron ht_neuron_nestml:
...
+    function m_inf_T real = 1.0 / ( 1.0 + exp( -( V_m / 1 mV + 59.0 ) / 6.2 ) )
+    function h_inf_T real = 1.0 / ( 1.0 + exp( ( V_m / 1 mV + 83.0 ) / 4 ) )
+    function tau_m_h real = 1.0 / ( exp( -14.59 - 0.086 * V_m / 1 mV ) + exp( -1.87 + 0.0701 * V_m / 1 mV ) )
     # I_KNa
     function D_influx_peak real = 0.025
     function tau_D real = 1250.0 # yes, 1.25s
     function D_thresh real = -10.0
     function D_slope real = 5.0
-    function D_influx real = 1.0 / ( 1.0 + exp( -( V_m - D_thresh ) / D_slope ) )
+    function D_influx real = 1.0 / ( 1.0 + exp( -( V_m / 1 mV- D_thresh ) / D_slope ) )
```
heplesser: D_thresh and D_slope have units mV.

models/iaf_psc_exp.nestml
```diff
@@ -85,7 +85,7 @@ neuron iaf_psc_exp_neuron:
     shape I_shape_in = exp(-1/tau_syn_in*t)
     shape I_shape_ex = exp(-1/tau_syn_ex*t)
     function  I_syn mV = (curr_sum(I_shape_in, in_spikes) + curr_sum(I_shape_ex, ex_spikes) + I_e + currents)
-    V_abs' = -1/tau_m * V_abs + 1/C_m * I_syn
+    V_abs' = -V_abs/tau_m + (I_syn/C_m) * 1 [s**3 * A**2 /(kg * m**2)] * 1 [s/Gs]
```
heplesser: The units here seem very confusing. I would expect `I_syn` and `C_m` to have their units defined elsewhere, so that the second term can be written without explicit units just as the first. And if units are needed, they should be given as high-level units with suitable prefactors (e.g. mV, pF) instead of basic MSKA and a `G` prefactor.

PTraeder: Yes, this one is complete nonsense. It is likely an aggregate of several fixes. This all equates to a factor of 1 nS. Fixed in update.

models/aeif_cond_exp_implicit.nestml
```diff
@@ -115,9 +115,9 @@ neuron aeif_cond_exp_implicit:
       emit_spike()
     end
 
-    g_ex += spikeExc.get_sum()
-    g_in += spikeInh.get_sum()
-    I_stim = currents.get_sum()
+    g_ex += spikeExc * 1 nS
+    g_in += spikeInh * 1 nS
+    I_stim = currents
```
PTraeder: Spike buffers are typed as real. g_ex and g_in are both nS. 
Current buffers are already typed as pA (same as I_stim in the model) so no factor is necessary.

DimitriPlotnikov: currents are per construction always `pA`, since it models an external current which has by definition `pA` as its dimension. In contrast to current buffers, spike buffers may be either current-based or conductance based. Since it is not always possible (or it is not transparent for the modeler) we decided to make spike buffers unitless. Thus, the modeler is forced to state the desired type of the input by multiplying it with a corret unit literal (1pA in the case of current based models, 1nS in the case of the conductance based models).

models/hh_cond_exp_traub.nestml
```diff
@@ -37,12 +37,12 @@ neuron hh_cond_exp_traub_neuron:
...
-    function alpha_m_init real = 0.32 * ( 13. - V_m ) / ( exp( ( 13. - V_m ) / 4. ) - 1. )
-    function beta_m_init real = 0.28 * ( V_m - 40. ) / ( exp( ( V_m - 40. ) / 5. ) - 1. )
-    function alpha_h_init real = 0.128 * exp( ( 17. - V_m ) / 18. )
-    function beta_h_init real = 4. / ( 1. + exp( ( 40. - V_m ) / 5. ) )
+    function alpha_n_init real = 0.032 * ( 15. - V_m / 1 mV) / ( exp( ( 15. - V_m / 1 mV ) / 5. ) - 1. )
+    function beta_n_init real = 0.5 * exp( ( 10. - V_m / 1 mV) / 40. )
+    function alpha_m_init real = 0.32 * ( 13. - V_m / 1 mV) / ( exp( ( 13. - V_m / 1 mV) / 4. ) - 1. )
+    function beta_m_init real = 0.28 * ( V_m /1 mV  - 40. ) / ( exp( ( V_m / 1 mV - 40. ) / 5. ) - 1. )
+    function alpha_h_init real = 0.128 * exp( ( 17. - V_m / 1 mV) / 18. )
+    function beta_h_init real = 4. / ( 1. + exp( ( 40. - V_m / 1 mV ) / 5. ) )
```
DimitriPlotnikov: It should not be necessary to points everywhere (but I have to check what happens if it translated to C++). Yes you can use 15mV as an unit literal.

models/hh_cond_exp_traub.nestml
```diff
@@ -67,16 +67,16 @@ neuron hh_cond_exp_traub_neuron:
...
+    function beta_m real = 0.28 * ( V_rel / 1 mV - 40. ) / ( exp( ( V_rel / 1 mV - 40. ) / 5. ) - 1. )
+    function alpha_h real = 0.128 * exp( ( 17. - V_rel / 1 mV ) / 18. )
+    function beta_h real = 4. / ( 1. + exp( ( 40. - V_rel / 1 mV ) / 5. ) )
 
-    Act_m' = alpha_m - ( alpha_m + beta_m ) * Act_m
-    Act_h' = alpha_h - ( alpha_h + beta_h ) * Act_h
-    Inact_n' = alpha_n - ( alpha_n + beta_n ) * Inact_n
+    Act_m' = ( alpha_m - ( alpha_m + beta_m ) * Act_m ) / 1 ms
+    Act_h' = ( alpha_h - ( alpha_h + beta_h ) * Act_h ) / 1 ms
+    Inact_n' = ( alpha_n - ( alpha_n + beta_n ) * Inact_n ) / 1 ms
```
PTraeder: Those rates are defined as real, not as 1/ms. I was not aware that this should be the case when I modified the models. Changing the definition of alpha and beta rates from real type to 1/ms would indeed make these extra factors unnecessary.

models/hh_cond_exp_traub_implicit.nestml
PTraeder: See above

models/hh_cond_exp_traub.nestml
PTraeder: I changed it in both hh_con_exp_traub models.

models/ht_neuron.nestml
PTraeder: They would arrive there with proper units if they were defined with proper units. 
Fixed in update

models/ht_neuron.nestml
PTraeder: see above

PTraeder:
@heplesser @Silmathoron @jougs I have adressed several requests regardeing this PR. Can you look over the changes/current state of the models and give your Ok/further concerns for the PR?

Silmathoron:
I'm sorry, I cannot look at this before 3/23... I will as soon as possible after that date.

models/aeif_cond_exp_implicit.nestml
DimitriPlotnikov: @heplesser 
We will tackle explicit typing of ports in the upcoming release.

DimitriPlotnikov:
I'm going to merge this PR today evening, if there are no strong concerns regarding aginst this. Generally, this PR enforces type soundness of SI types. There is still lot room for improvement, but it already increases the quality of models significantly.

DimitriPlotnikov:
I'm merging the PR. Of course, the general discussion is not over.

models/iaf_chxk_2008_implicit.nestml
```diff
@@ -73,13 +73,13 @@ neuron iaf_chxk_2008_implicit:
   internals:
     # Impulse to add to DG_EXC on spike arrival to evoke unit-amplitude
     # conductance excursion.
-    PSConInit_E real = 1.0 * e / tau_syn_ex
+    PSConInit_E real = 1.0 ms * e / tau_syn_ex
```
Silmathoron: I'm very sorry this comes so late (I must say I completely forgot this PR) but I just noticed the combination of lines 76/89 and [120](https://github.com/PTraeder/nestml/blob/c0a589c34b2c7109689d6fc44c96ddd0c2fd1943/models/iaf_chxk_2008_implicit.nestml#L120)/121 is redundant (therefore less clear) in terms of units: explicit units are used twice because ``PSConInit_I/E`` is ``real``, whereas it could directly be ``nS``. 
This would lead to:

    PSConInit_E nS = 1.0 nS * e / tau_syn_ex
    PSConInit_I nS = 1.0 nS * e / tau_syn_ex

then

    g_ex' += spikesExc * PSConInit_E
    g_in' += spikesInh * PSConInit_I

models/iaf_cond_beta.nestml
```diff
@@ -65,11 +65,11 @@ neuron iaf_cond_beta_neuron:
   internals:
     # Impulse to add to GE' on spike arrival to evoke unit-amplitude
     # conductance excursion.
-    PSConInit_E real = 1.0 * e / tau_syn_rise_E #
+    PSConInit_E real = 1.0 * e / tau_syn_rise_E * 1 ms
```
Silmathoron: Same as before: typing this as ``nS`` would make things simpler

models/iaf_cond_exp_implicit.nestml
```diff
@@ -61,11 +61,11 @@ neuron iaf_cond_exp_implicit:
   internals:
     # Impulse to add to g_ex' on spike arrival to evoke unit-amplitude
     # conductance excursion.
-    PSConInit_E real = 1.0 * e / tau_synE
+    PSConInit_E real = 1.0 ms * e / tau_synE
```
Silmathoron: See above

models/terub_neuron_gpe.nestml
```diff
@@ -150,8 +150,8 @@ neuron terub_neuron_gpe:
   end
 
   internals:
-    PSCurrInit_E real = 1.0 * e / tau_syn_ex
-    PSCurrInit_I real = 1.0 * e / tau_syn_in
+    PSCurrInit_E real = 1.0 ms * e / tau_syn_ex
```
Silmathoron: See above but with ``pA``

models/iaf_chxk_2008_implicit.nestml
DimitriPlotnikov: @Silmathoron  it sounds reasonalbe. @PTraeder  what is your opinion?

models/iaf_cond_exp_implicit.nestml
PTraeder: In this model, the PSConInit variables are not used. Is this intended?

PTraeder:
@Silmathoron I implemented your propositions in #346

PR #344, opened 2017-02-27
Add reference documentation

jougs:
This is the first version of the reference manual for NESTML.

The PR is not (yet) meant for merging, but more as a basis for discussion about the structure and content. Please feel free to add comments and ideas!


doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+
+Data types define types of variables as well as parameters and return values of functions. NESTML provides the following primitive types and physical data types:
+
+#### Primitive data types
+* `real` corresponds to the `double` data type in C++. Example literals are: `42.0`, `-0.42`, `.44`
+* `integer` corresponds to the `long` data type in C++. Example literals are: `42`, `-7`
+* `boolean` corresponds to the `bool` data type in C++. Its only literals are `true` and `false`
+* `string` corresponds to the `std::string` data type in C++. Example literals are: `"Bob"`, `""`, `"Hello World!"`
+* `void` corresponds to the `void` data type in C++. No literals are possible and this can only be used in the declaration of a function without return value.
+
```
jougs: @DimitriPlotnikov: do we really need the `void` data type? If we can establish that function return types can always be deducted from the type of the returned expression, I think we can also drop this. I think that `void` is a very low-level concept that should not be present in a high-level language as NESTML.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+|Quantity | Unit name | Definition | NESTML/SI unit |
+|---|---|---|---|
+| capacitance            | farad   | A<sup>2</sup> s<sup>4</sup> kg<sup>−1</sup> m<sup>−2</sup> | F |
+| electrical conductance | siemens | A<sup>2</sup> s<sup>3</sup> kg<sup>−1</sup> m<sup>−2</sup> | S |
+| electric charge        | coulomb | A s                                                        | C |
+| electric potential     | volt    | kg m<sup>2</sup> A<sup>−1</sup> s<sup>−3</sup>             | V |
+| electric resistance    | ohm     | kg m<sup>2</sup> A<sup>−2</sup> s<sup>−3</sup>             | ohm/&‌#x03A9; |
+| frequency              | hertz   | s<sup>−1</sup>                                             | Hz |
+
```
jougs: @abigailm, @DimitriPlotnikov: I think these units suffice for writing neuron models. I'd rather leave out the ones below. We can always add more if people request them. At least we should not have them in the reference manual.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+|Factor | SI Name | NESTML prefix | Factor | SI Name | NESTML prefix
+|---|---|---|---|---|---|
+|10<sup>-1</sup>  | deci  | d  | 10<sup>1</sup>  | deca  | da |
+|10<sup>-2</sup>  | centi | c  | 10<sup>2</sup>  | hecto | h |
+|10<sup>-3</sup>  | milli | m  | 10<sup>3</sup>  | kilo  | k |
+|10<sup>-6</sup>  | micro | mu | 10<sup>6</sup>  | mega  | M |
+|10<sup>-9</sup>  | nano  | n  | 10<sup>9</sup>  | giga  | G |
+|10<sup>-12</sup> | pico  | p  | 10<sup>12</sup> | tera  | T |
+|10<sup>-15</sup> | femto | f  | 10<sup>15</sup> | peta  | P |
+
```
jougs: @abigailm, @DimitriPlotnikov: I think we can safely leave out the ones below and just keep the ones above.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+Simple unit literals are composed of a number and a type name (with or without a whitespace inbetween the two). Complex unit literals are composed of a number and a type inside of square brackets. Simple units can also be put inside square brackets.
+```
+<number> <unit_type>
+<number> [<unit_type>]
+
+a mV = 1mV
+aa mV = 1[mV]
+b mV/ms = 1[mV/ms]
+```
+
```
jougs: @DimitriPlotnikov: I don't understand the distinction between simple and complex literals from the point of the user. Do we really need this distinction (internally) and do we have to distinguish between them in the frontend? Also, I think we should either omit the term `literal` here or explain it a bit more. This is computer science terminology which many of our customers probably won't understand.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+  <statements>
+  if 2 < 3:
+    <statements>
+  end
+end
+```
+
+### Expressions and operators
+
+Expressions in NESTML can be specified in a recursive fashion.
```
jougs: @DimitriPlotnikov: this sentence is a bit lonely here. What does it mean and why is this relevant to the user? Can you please elaborate a bit on this?

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+
+Each neuron model can only send a single type of event. The type of the event has to be given in the `output` block. Currently, however, only spike output is supported.
+```
+output: spike
+```
+
+Please note that this block is *not* terminated with the `end` keyword.
+
+## Synaptic input
+
```
jougs: @ingablundell: I think that we really should have some examples in this section to demonstrate the use of the functions. We should also have links to the models in the repository that use these concepts.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+* `resolution` returns the current resolution of the simulation in ms. This can be set by the user using the PyNEST function `nest.SetKernelStatus({"resolution": ...})`.
+* `steps` takes one parameter of type `ms` and returns the number of simulation steps in the current simulation resolution.
+
+These functions can be used to implement custom buffer lookup logic, but should be used with care.
+
+## Equations
+
+### Synaptic response
+
```
jougs: @ingablundell: this section would benefit a lot by having an example and also a figure that demonstrates what such a response looks like.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+> `integrate_odes`: Explain the current behavior
+
+> `process_spikes()`: Do we still have this function?
+
+`emit_spike()`: calling this function in the `update`-block results in firing a spike to all target neurons and devices time stamped with the current simulation time.
+
+### Solver selection
+
+Currently there is support for GSL and exact integration.
```
jougs: @ingablundell: this section needs to be written and possibly renamed to something more sensible.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+    end
+  else:
+    ref_counts  -= 1 # neuron is refractory
+  end
+
+end
+```
+
+## Setting and retrieving model properties
+
```
jougs: This requires more examples and some better explanations.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+* All variables in the `state` and `parameters` blocks are added to the status dictionary of the neuron.
+* Values can be set using `nest.SetStatus(<gid>, <variable>, <value>)` where `<variable>` is the name of the corresponding NESTML variable.
+* Values can be read using `nest.GetStatus(<gid>, <variable>)`. This call will return the value of the corresponding NESTML variable.
+
+## Recording values with devices
+
```
jougs: This requires more examples and some better explanations.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+* All values in the `state` block are recordable by a `multimeter` in NEST.
+* The `recordable` keyword can be used to also make variables in other blocks (`parameters, internals`) available to recording devices.
+```
+parameters:
+  recordable t_ref ms = 5ms
+end
+```
+
+## Guards
+
```
jougs: This requires a bit more motivation.

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+The content of spike and current buffers can be used by just using their plain names. NESTML takes care behind the scenes that the buffer location at the current simulation time step is used.
+
+## Dynamics and time evolution
+
+`update:` inside this block the current time can be accessed via the variable `t`
+
+> `integrate_odes`: Explain the current behavior
+
+> `process_spikes()`: Do we still have this function?
```
jougs: Do we still have this function?

doc/syntax.md
```diff
@@ -0,0 +1,476 @@
...
+```
+has to be stated in the `state` block. If the initial values are not defined in `state` it is assumed that they are zero and unit checks are no longer possible.
+
+The content of spike and current buffers can be used by just using their plain names. NESTML takes care behind the scenes that the buffer location at the current simulation time step is used.
+
+## Dynamics and time evolution
+
+`update:` inside this block the current time can be accessed via the variable `t`
+
+> `integrate_odes`: Explain the current behavior
```
jougs: Explain the current behavior of this function and the variants with one or more equations as argument.

doc/syntax.md
Silmathoron: Actually this one is something I proposed. The idea was to avoid having the user deal with spikes at the wrong time in the `update` block, but I don't know what its status is.
Moreover, since we talked about the possibility of making the ``update`` block optional by default, it might not be necessary in this case...

DimitriPlotnikov: `process_spikes` is removed. the docu will be updated accordingly to explain what and why happens under the hood.

DimitriPlotnikov:
Documentation is better than no documentation. 

As discussed I'll merge the PR and start new dedicated well focused issues with incremental improvements.

PR #347, opened 2017-04-16
Unit addition

PTraeder:
Allow Addition/Assignments etc between two units of the same base type, but differing magnitudes.


PTraeder:
Will have to look over this some more

PR #351, opened 2017-05-03
Don't print unit names in codegeneration

PTraeder:
Use LegacyExpressionPrinter in place of ExpressionsPrettyPrinter in neuron code generation.


DimitriPlotnikov:
PrettyPrinter ommits units in unit literals and solves the broken C++ code.

DimitriPlotnikov:
PrettyPrinter ommits units in unit literals and solves the broken C++ code.

PR #366, opened 2017-06-26
Coco no assignment to parameters

PTraeder:
Install IllegalAssignment Coco that throws an error when a parameter, equation, internals, or alias symbol is assigned a value outside of its declaration.

Adapt models to those changes

Add test for this coco to NestmlCoCosTest

Replace i_stim with currents buffer in models


DimitriPlotnikov:
Can you extend the Coco to forbid also assignments to:
a) internals
b) ode-functions 
```
neuron test:
  internals:
   a real = 0
  end
  equations:
    function b real = 0
  end
  update:
    a = 2. # error a)
    b = 2. # error b)
  end
end
```

Merge all these checks with the `AssignmentToAlias`-Coco and give it a proper name (e.g. FobidenAssignment)

PTraeder:
@DimitriPlotnikov implemented your suggestions

PR #374, opened 2017-08-24
Add GPLv2 license and static code analysis tool

gtrensch:
This PR adds:
- NEST compliant GPLv2 license
- shell scripts to perform VERA++, CLANG-FORMAT and CPPCHECK static code analysis

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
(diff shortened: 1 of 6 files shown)


DimitriPlotnikov:
@gtrensch 
great contribution!

However, I have several questions wrt checks.
1) Do these checks run automatically in travis-ci?
2) What is the purpose of the vera++? I assume it is a tool for C++ code, but we do have here a Python project.

gtrensch:
@DimitriPlotnikov 
Not yet, but the static_code_analysis.sh script was written to be called either from Travis or from the check_code_style.sh script for manual usage. For an automated run on Travis we additionally need a logparser which can also be taken then from Nest and adapted to our needs. However, automated checks on Travis should be the goal.  It think it is fine to implement that in a later stage of the project. But manual checks before pushing changes is something we should have from the beginning.
In general the tool is able to handle C/C++ and Python (I forgot to mention PEP8 in the list above.). Does the project will be pure Python? I was not aware of that. I thought we generate C++ code and need to format and check for Nest coding style compliance.

DimitriPlotnikov:
Yes, indeed we generate the C++ code at the end. And I also agree that it is a good idea to check the generated C++ code.

The project itself is a pure python project (with a little extension where Java is used to process the grammar. However this step is executed manually and seldom).

PR #376, opened 2017-09-07
Adds PyNestML Grammar

kperun:
Grammar added and ready for review.


DimitriPlotnikov:
Can you provide a better name for the PR?

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,329 @@
...
+         | left=expression (timesOp='*' | divOp='/' | moduloOp='%') right=expression
+         | left=expression (plusOp='+'  | minusOp='-') right=expression
+         | left=expression bitOperator right=expression
+         | left=expression comparisonOperator right=expression
+         | logicalNot='not' term=expression
+         | left=expression logicalOperator right=expression
+         | condition=expression '?' ifTrue=expression ':' ifNot=expression
+         | simpleExpression
+         ;
+
```
DimitriPlotnikov: Did you check that the priorities of the operators are comparable to the Python ones?

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,329 @@
...
+
+  /*********************************************************************************************************************
+  * Equations-Language
+  *********************************************************************************************************************/
+  /** ASTOdeDeclaration. Represents a block of equations and differential equations.
+
+    @attribute Equation      List with equations, e.g. "I = exp(t)" od differential equations.
+  */
+  odeDeclaration  : (equation | shape | odeFunction | NEWLINE)+;
```
DimitriPlotnikov: We don't need this production. You can embed the rhs where the `odeDeclaration` was used.

pynestml/src/main/grammars/org/Tokens.g4
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+  * Numeric literals. We allow integers as well as floats. Moreover, we ensure that values are either == 0 or
+  * do not start with 0, e.g., 01221.012, where the leading 0 does not make sense.
+  * Examples:
+  *  (1) 1
+  *  (2) 3.14
+  */
+  NUMERIC_LITERAL :  ([1-9][0-9]* | '0' ) ('.' [0-9]*)?;
+
```
DimitriPlotnikov: What I don't understand is how do you want do distinguish floating point literals and integers?

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,329 @@
...
+    @attribute Body    The body of the neuron, e.g. internal, state, parameter...
+  */
+  neuron : 'neuron' NAME body;
+
+  /** ASTBody The body of the neuron, e.g. internal, state, parameter...
+  */
+  body : BLOCK_OPEN
+           (NEWLINE | var_Block | dynamics | equations | inputBuffer | outputBuffer | function)*
+         BLOCK_CLOSE;
+
```
DimitriPlotnikov: I would use different names. E.g.:
`(NEWLINE | BlockWithVariables | UpdateBlock | EquationsBlock | InputBlock | OutputBlock | Function)*`

kperun:
@DimitriPlotnikov Regarding the format I use (in general): All tokens are completely upper-case with underlines as separators, e.g. NUMERIC_LITERAL (tokens have to start by upper case as required by Antlr).  Rules are starting with a lower-case (as required by Antlr) and separation of words is achieved by camel case, e.g., ifClause.  I will rework the grammar according to this specification.

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: This part has been reworked according to your feedback.

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: This part has been reworked according to your feedback.

pynestml/src/main/grammars/org/Tokens.g4
kperun: This is an important point of discussion. While it is not difficult to distinguish if it is a float or an integer, it is currently not possible to write numbers in the scientific representation, e.g. 10E2. However, the original version as used in NESTML uses a cascade of interfaces to represent arbitrary numbers. I will come back to this one later on.

DimitriPlotnikov:
@kperun 
Take a look on how it is handled in the Python grammar:
E.g. https://github.com/antlr/grammars-v4/tree/master/python3

Indeed it is not the most important feature at the moment.

DimitriPlotnikov:
@gtrensch, @jougs 
I need one more thumb-up.

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: Yes, I used the following reference to check if the priorities of the operators are correct * [link](https://www.tutorialspoint.com/python/operators_precedence_example.htm).  However, it states that comparison operators have a higher priority than equality operators. In my opinion, this is not relevant for us, since expression like 5 < 6 != V_m do not make any sense.

kperun:
@DimitriPlotnikov I am currently dealing with the following question: Does it make sense to allow models to contain numbers with leading 0's, i.e., something like 05.114 . In my opinion, we should prevent such declarations. However, in the case of the python grammar as provided above, this case is not regarded, thus such definitions allowed. What do you think?

pynestml/src/main/grammars/org/Tokens.g4
kperun: This part is now fixed. The following grammar is now used for this part:

```
 INTEGER : NON_ZERO_INTEGER
          | '0';

  fragment NON_ZERO_INTEGER : [1-9][0-9]*;

  /**
  * The following declaration originates from Antrl4 Python Grammar definition as distributed under the MIT license.
  * link: https://github.com/antlr/grammars-v4/blob/master/python3/Python3.g4
  */
  /*
  * A flaot can be a point float, e.g., 10.05 or 0.1, or an exponent float, e.g. 10E10.
  */
  FLOAT : POINT_FLOAT | EXPONENT_FLOAT;

  fragment POINT_FLOAT : (NON_ZERO_INTEGER |'0')? FRACTION
                       | (NON_ZERO_INTEGER |'0') '.'
                       ;

  fragment EXPONENT_FLOAT: ( NON_ZERO_INTEGER | POINT_FLOAT ) EXPONENT ;

  /**
  * The exponent is introduced by e or E, the signum and an integer.
  */
  fragment EXPONENT: [eE] [+-]? (NON_ZERO_INTEGER |'0');

  fragment FRACTION: '.' [0-9]+;
```

kperun: Now, it still remains to discuss if we want to include other bases. Currently, only base 10 is used. What do you think @DimitriPlotnikov ?

kperun:
An updated version of the grammar has been pushed. All remarks, except for the questions which are still open, have been implemented.

kperun:
Addtionally, I also modified the following part of the grammar:
```
 simpleExpression : functionCall
                   | BOOLEAN_LITERAL // true & false;
                   | (INTEGER|FLOAT) (variable)?
                   | isInf='inf'
                   | variable;
```
where previously:
```
 simpleExpression : functionCall
                   | BOOLEAN_LITERAL // true & false;
                   | (INTEGER|FLOAT) (variable)?
                   | isInf='inf'
                   | Name // <---- no longer used
                   | variable;
```
The reason is, that it is not required to distinguish between variables and names. Each name can be seen as a variable. @DimitriPlotnikov , what do you think? Is there a special reason for this design decision?

DimitriPlotnikov:
Indeed, you don't need both: `name` and `variables`.

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,320 @@
...
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author dplotnikov, kperun
```
DimitriPlotnikov: Remove this entry everywhere. The authors are not important.

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,320 @@
...
+/**
+  Grammar representing the Simple Programming Language (SPL). It is easy to learn imperative
+  language which leans on the Python syntax.
+*/
+grammar PyNESTML;
+
+  import Tokens;
+
+  nestmlCompilationUnit : (neuron | NEWLINE )* EOF;
+  /*********************************************************************************************************************
```
DimitriPlotnikov: Add a blank-line

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,320 @@
...
+  unitType : leftParentheses='(' compoundUnit=unitType rightParentheses=')'
+           | base=unitType powOp='**' exponent=INTEGER
+           | left=unitType (timesOp='*' | divOp='/') right=unitType
+           | unitlessLiteral=INTEGER divOp='/' right=unitType
+           | unit=NAME;
+
+  /*********************************************************************************************************************
+  * Expressions-Language
+  *********************************************************************************************************************/
+  /**
```
DimitriPlotnikov: Blank line

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,320 @@
...
+
+  /**
+    ASTFunctionCall Represents a function call, e.g. myFun("a", "b").
+    @attribute name The (qualified) name of the fucntions
+    @attribute args Comma separated list of expressions representing parameters.
+  */
+  functionCall : calleeName=NAME '(' (args=arguments)? ')';
+
+  arguments : expression (',' expression)*;
+
```
DimitriPlotnikov: Remove blank

pynestml/src/main/grammars/org/PyNESTML.g4
```diff
@@ -0,0 +1,320 @@
...
+  forStmt : 'for' var=NAME 'in' vrom=expression '...' to=expression 'step' step=signedNumericLiteral BLOCK_OPEN block BLOCK_CLOSE;
+
+  whileStmt : 'while' expression BLOCK_OPEN block BLOCK_CLOSE;
+
+  signedNumericLiteral : (negative='-'?) (INTEGER|FLOAT);
+
+  /*********************************************************************************************************************
+  * Nestml-Language
+  *********************************************************************************************************************/
+  /** ASTNeuron represents neuron.
```
DimitriPlotnikov: Add blank

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: This issue has already been documented and resolved!

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: Fixed!

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: Fixed!

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: Fixed!

pynestml/src/main/grammars/org/PyNESTML.g4
kperun: Fixed!

DimitriPlotnikov:
Regarding blanks: I just highlighted some of "issues". I assume, you reworked the whole grammar definition.

pynestml/src/main/grammars/org/Tokens.g4
```diff
@@ -0,0 +1,84 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+lexer grammar Tokens;
+
+  SL_COMMENT : ('#' (~('\n' |'\r' ))*|
```
gtrensch: What is about multi-line comments?

gtrensch:
I approve this if my comment is addressed. :+1:

pynestml/src/main/grammars/org/Tokens.g4
kperun: The second alternative ensures that comments can be multiline:
```  
SL_COMMENT : ('#' (~('\n' |'\r' ))*| 
               '/*' .*? '*/' | '"""' .*? '"""')
               -> channel(HIDDEN)
             ;
```
Here, we allow Java style comments /* ... */ and Python style comments """...""" .

kperun: However, currently, comments are not further regarded. This is a part of future work which will deal with those parts.

DimitriPlotnikov: Hi, but it would be better, that you either have 2 productions `SL_COMMENTS` and `ML_COMMENTS` or only one which is `COMMENTS`.

kperun: Assuming SL stands for Single-Line, then you are right =).

kperun:
Grammar has been updated according to the feedback. Moreover, following parts have been changed:
Previously:
```
 block : (stmt | NEWLINE)?
```
now:
```
  block : ( smallStmt | compoundStmt | NEWLINE )*;
```
The rule for stmt has been removed since it does not have any benefits. 
Moreover, the rule "derivative" has been removed since it corresponds exactly to the "variable" one. The "parameters" rule has also been removed since it does not make any sense encapsulate everything in an intermediate rule. Several minor changes have also been included. Finally, multiline and single line comments are now two separated tokens.

DimitriPlotnikov:
I disagree regarding the block-rule change. Now, at the AST level you have an ASTBlock class with two lists for `smallStmts` and `compoundStmt`s. However, at that point the oder of the statements in the source model is lost. For example, how are you going to pretty print the model (without doing black magic with original line numbers)?

In my opinion `stmt` rule as it was before is crucial for the modeltransforamtions and codegeneration.

kperun:
@DimitriPlotnikov Although it is quite easy to get the next statement during the parsing step, I think you are right. I will fix this asap.

PR #377, opened 2017-09-13
Adds ASTClasses, Parser, Tests with reference neuron models

kperun:
The following concepts have been implemented in this pull request:
* Lexer/Parser
* ASTClasses
* A visitor to create the internal representation
* Several tests to check the creation of ASTs and the functionality of the parser


DimitriPlotnikov:
1) Tests fail
2) In Python file the copyright file must be changed. a) there are no authors entry. b) Take a look at the "reference comment" in https://github.com/nest/nest-simulator/blob/master/extras/parse_travis_log.py

kperun:
@DimitriPlotnikov As already discussed, the tests themselves do not fail, but the checking as executed by travis does. In order to use the AST classes, the install file must be executed. This creates a new module in the python distro dir on your system, which can then be included in the code by: pynestml.main..... However, this concept somehow does not work with travis. It just can't find those modules.

kperun:
Regarding the copyright header: I will change this asap.

setup.py
```diff
@@ -0,0 +1,39 @@
...
+    description='NESTML is a domain specific language that supports the specification of neuron models in a'
+                ' precise and concise syntax, based on the syntax of Python. Model equations can either be given'
+                ' as a simple string of mathematical notation or as an algorithm written in the built-in procedural'
+                ' language. The equations are analyzed by NESTML to compute an exact solution if possible or use an '
+                'appropriate numeric solver otherwise. PyNESTML represents a toolchain migrated from Java to Python.',
+    license='GNU General Public License v2.0',
+    url='https://github.com/nest/nestml/tree/PyNestML',
+    packages=find_packages(),
+    install_requires=['numpy >= 1.8.2',
+                      'sympy >= 1.0',
```
DimitriPlotnikov: Actually, you need a version at least of `1.0.1`

kperun: I will fix this one in the next push.

DimitriPlotnikov: there is 1.1.1 in the repository. I'm testing it.

DimitriPlotnikov: It works, take the `1.1.1`

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.7"
```
DimitriPlotnikov: Why did you remove the 2.x Python. I thought we can do both.

kperun: Somehow, it is currently a bug I try to deal with: Enums are not directly available in python 2.7. Still, in my master branch [link](https://github.com/kperun/nestml/blob/PyNestML/.travis.yml), it works quite well, but here, we the exactly the same setup, it results in strange import errors. So yes, we will provide 2.7 support, but currently, the are some bugs.

DimitriPlotnikov: Ok, take a look on:
https://stackoverflow.com/questions/26828206/importerror-no-module-named-enum
https://stackoverflow.com/questions/39670197/ubuntu-16-04-python-2-7-importerror-no-module-named-enum

kperun: Fixed, thanks for the information. I used 
```
pip install --upgrade pip enum34
```
as an additional pre-installation step and it seems to solve the problem. Still, on my master branch, it works quite well without this. I will have to take a look at this matter.
All test work now fine and python 2.7. has been integrated for testing.

gtrensch:
@kperun many thanks for the work! :+1:  Looked together with @DimitriPlotnikov at this.
Just a few comments / questions:
- Did you copy the latest version of doc/syntax.md? The formatting of some of the ASCII-tables look corrupted.
- There are a couple of generated files. Is it possible to move those files, and only those files, to a separate sub-folder with a descriptive name, e.g. /generated or something like that?
- Additionally a generated file should contain the copyright notice from the file it is generated from. In our case the ANTLR copyright notice.
- requirements.txt refers to SymPy version 1.0 or greater, but setup.py specifies version > 1.1.1. As per @DimitriPlotnikov > 1.1.1 is needed. Thus, requirements.txt needs to be corrected.

I approve this pull request if the above is addressed.

kperun:
@gtrensch Issue 1 and 4 have been resolved.
Regarding issue 2:
It is easily possible to switch the dir in which all generated files are located. However, one of the following pull requests will more or less restructure the whole project to make it more "pythonic". I would, therefore, like to postpone this issue to one of the future releases in order to avoid making the work twice.
Regarding issue 3:
Here, we are dealing with two problems at the same time, being
a) The generated files should actually not be included in the repo, however, should we still include them in order to enable the user to use the system out of the box without antlr4 lexer/parser generator and only the runtime?
b) In antlr, I did not find a parameter which can be used to add a custom header. So here, we can at most include it by hand. In this case, we would also have to include all generated files (extended by the corresponding copyright header) to the repo.
I think we should include all generated files and modify their headers by hand with the corresponding copyright. Here, the sole problem would be to adjust those headers whenever we generate new artifacts.
I am also not quite sure what you mean by "...should contain the copyright notice from the file it is generated from". Are you talking about the grammar file? Or the actual antlr copyright?

DimitriPlotnikov:
`syntax.md`: hm, it wasn't ready, it is not update, ... , remove it.

kperun:
@DimitriPlotnikov Fixed. I will update this file in a future pull request as soon as a complete version is available.

DimitriPlotnikov:
Currently, there are conflicts which must be solved. Other than that, I'll be ready to merge it.

PR #380, opened 2017-10-03
Adds Symbol Table, CoCos, Visitors, Unit System and more

kperun:
This pull request includes the following features:
* Symbol Table and all CoCos as required to ensure correctness of the model
* A complete rework of the project to make its structure phytonic
* A new concept for visitors, the NESTML Visitor
* A Unit System which enables checks for correctness of expressions
* A set of helper functions and a logger for precise logging of messages
* The grammar as adjusted to the latest version of buffer definitions (spike buffers with data types) and the initial block 
* Many bugfixes
and many more.


DimitriPlotnikov:
2 Additions from my side:
1) Every text that you log in Logger.logMessage in context condition must be collected in a single file (e.g. coco_constants). Since these are texts which are visible for the PyNESTML applicant they must be easy to review without navigated trough multiple files.
2) For the cocos the rule is to create an invalid model (that shows the error violation). And an invalid model that shows a corrected version. Valid models are placed in a folder `valid`, invalid models are placed in a folder `invalid`.

DimitriPlotnikov:
3) There are conflicts in the PR.

kperun:
Issue 1 has been implemented and will be pushed soon. Now, the 'store_log' parameter can be provided to the frontend which indicates that all messages shall be stored in a file in the target directory. This file can then be reviewed for errors.

DimitriPlotnikov:
Ok,
1) What I meant, was that context conditions produce some textual output. All strings which are used in context conditions must be placed in a single class/file/module (pick the most obvious python alternative, e.g. module) s.t. somebody can review them.
2) yes, also log can be stored in a file. Use json output. The current format looks like this:
```
[{
  "filename" : "iaf_cond_exp.nestml",
  "neuronName" : "iaf_cond_exp_implicit",
  "severity" : "WARNING",
  "code" : "NESTML_POST_PROCESSING_ERROR",
  "row" : 130,
  "col" : 4,
  "message" : "The type of (derived) variable V_m' is: 1 / MH. This does not match Type of RHS expression: A / F"
},]
```

kperun:
@DimitriPlotnikov A new concept for the handling of error messages has been implemented. Now, a file containing all messages as well as the corresponding message number has been integrated and is used across the project. Moreover, the log can now be stored as a single JSON file. 
As mentioned in the last PR, a new folder "generated" has been created which contains all files which are created by tools ( in our case lexer/parser).
CoCos have now a valid and an invalid example in the corresponding subfolder and correctness in both cases for all cocos is tested.

DimitriPlotnikov:
Why are templates stored in files with an extension `html`?

kperun:
@DimitriPlotnikov Regarding the first issue: There is no default file-extension for jinja files and according to this discussion [link](https://stackoverflow.com/questions/29590931/is-there-an-idiomatic-file-extension-for-jinja-templates), there won't be any. Jinja was developed mostly for generation of html files and uses across the official tutorial the file ending .html, so I, therefore, used this file ending. If you have any suggestions (like .jtf  for Jinja Template File), feel free to tell me.
Regarding the second issue: I will create a new PR as soon as the generator and all its parts are completely implemented. So each PR represents a completed part but may contain something which is required later on.

DimitriPlotnikov:
1) Ok, my rule for the templates extension: not `html` and should not start with `j`. `.template`, `cgt` (code generation template), `gt`, `tf`, ...

2) Fine for me.

kperun:
@DimitriPlotnikov I have tried to find a better file extension guideline for our jinja templates. Candidates: .j2, .jinja2, or .jtf (jinja template file).  Unfortunately, all of those are not that good and start with .j . But even after searching for quite some time, I did not find any suggestions besides html and jinja2. @gtrensch any suggestions maybe?

DimitriPlotnikov:
`jinja2` is actually fine for me.

kperun:
@DimitriPlotnikov I have adjusted the file extension of the templates in the latest push. Please review it and accept the PR.

PR #392, opened 2017-11-09
Unit addition

PTraeder:
Implement casting between units of matching physical type but differing magnitude. 
This allows assignments and expressions such as:

```
smallV mV = 5 mV                   
bigV V = 1000 mV                    #assigns V/mv * 1000 mV 
bigV = bigV + smallV               #assigns bigV + V/mV * smallV 
smallV = bigV - smallV             #assigns mV/V * bigV - smallV
smallV += bigV                        #handles smallV = smallV + bigV instead !!
```

Installs the necessary ASTNodes to implicitly form a valid expression as hinted in the "comments" above.
Whenever this is done an "implicit cast" warning is dropped.

@DimitriPlotnikov Can you please have a look at the last commit in the PR titled "experimental fix"? It adresses the resolving error I encountered but it seems a dodgy fix to me.


models/iaf_cond_beta.nestml
```diff
@@ -98,8 +98,8 @@ neuron iaf_cond_beta_neuron:
 
     end
 
-    g_ex' += spikeExc * PSConInit_E
-    g_in' += spikeInh * PSConInit_I
+    g_ex' = g_ex' + spikeExc * PSConInit_E
+    g_in' = g_in' + spikeInh * PSConInit_I
```
DimitriPlotnikov: Can we leave the short notation?

PR #394, opened 2017-11-18
Model cleanup

PTraeder:
Clean warnings from models in the models/ folder. 
This PR is built on top of #392 since that one is still pending review.


src/main/java/org/nest/nestml/_cocos/IllegalExpression.java
```diff
@@ -71,15 +74,19 @@ public void check(final ASTAssignment node) {
 
       ASTExpr.Builder exprBuilder = ASTExpr.getBuilder();
       if(node.isCompoundProduct()){
+        node.setCompoundProduct(false);
```
DimitriPlotnikov: Why do we need set  it manually?

src/main/java/org/nest/utils/AstUtils.java
```diff
@@ -353,7 +354,7 @@ public static String getNameOfDerivedVariable(ASTShape astShape) {
 
   public static String convertDevrivativeNameToSimpleName(final ASTVariable astVariable) {
```
DimitriPlotnikov: Do we need this function at all? I mean, the `toString` has the same functionality.

PR #410, opened 2018-02-08
Mypy travis

PTraeder:
Integrate mypy test into Travis CI:

This PR adds dependencies for mypy for python 3.
In python 3 environments, it executes mypy before the rest of the test suite. 

At this time it is intended to error on Travis since the actual mypy tests fail. 
To remedy the errors, some refinement of the mypy cmd line call and some initial annotation in the code base is required.


jougs:
@PTraeder: Just in case you don't know yet: if you want to test the TravisCI setup without creating public pull requests, you can enable TravisCI for your own forks on https://travis-ci.org/profile/PTraeder.

jougs:
@PTraeder: In addition, you should rebase-squash these kind of experiments in order to avoid commit messages like the ones above in the global public history.

PTraeder:
@jougs Allright, will do. Thanks wrt Travis.

DimitriPlotnikov:
pr doesn't run. if it will not be fixed be the next week, I'll colse it.

PR #411, opened 2018-02-26
Make different but compatible units play well together & massive Refactoring

PTraeder:
Conversion between compatible SI  units is achieved by noting the conversion factor in a new field of expression classes, namely `implicitConversionFactor`. The type of the entire resulting expression (`factor * original expression`) is saved as the type of the `original expression`

**Declarations/Assignments**
`foo mV = 10V` is **treated** as `foo mV = 1000.0mV/V * (10V) `
`foo V = 1000mV` is **treated** as `foo V = 0.001V/mV * (1000mV) `

**Compound Assingment**
`foo mV += 10V` is **treated** as `foo mV += 1000.0mV/V * (10V) `
`foo V += 1000mV` is **treated** as `foo V += 0.001V/mV * (1000mV) `

**RHS Addition**
`somevar V = 1V + 1000 mV`  is **treated** as `somevar V = 1V + 0.001V/mV(1000mV)`

**Functions**
let:
```
function bar(param A)mA:
  return param
end
```

then bar()'s body is **treated** as:
```
function bar(param A)mA:
  return 1000.0mA/A * (param)
end
```

also:
a call to `bar(10mA)`  is treated as `bar(0.001 * (10mA))`

**Refactorings**

- ASTExpression and ASTSimpleExpression now share a common superclass in ASTExpressionElement.
Reduces code duplication.

- `TypeSymbol `is now abstract superclass to specific symbol type classes for NESTML types. i.e. `IntegerTypeSymbol`, `RealTypeSymbol`,`UnitTypeSymbol` etc.
This allows us to get rid of the clunky distinction through various boolean attributes, and with overloaded operators the handling of (expression)-type arithmatic is vastly improved.

- `ErrorTypeSymbol` replaces the Either- error handling for expression types.

- Various variables and funcions renamed to fit pep8

- Rename `ASTElement `to `ASTNode`

- Rename `ASTExpressionElement `to `ASTExpressionNode`

@DimitriPlotnikov


DimitriPlotnikov:
@PTraeder 
The issue why Python2 fails seems to be described in:
https://stackoverflow.com/questions/38963018/typeerror-super-takes-at-least-1-argument-0-given-error-is-specific-to-any

PTraeder:
@DimitriPlotnikov Yes i only just got to fixing it.

kperun:
@DimitriPlotnikov An Ok from my side. Minor bug:
`-  ForStmt = "for" var:NAME "in" vrom:Expression "..." to:Expression "step" step:SignedNumericLiteral
 +  ForStmt = "for" var:NAME "in" from:Expression "..." to:Expression "step" step:SignedNumericLiteral
              BLOCK_OPEN
               Block
              BLOCK_CLOSE;
`
was changed in PyNestML.mc4. Now the "vrom" is called extra in this way since MC generates an artifact with an attrbute which is then called vrom. If we use from, then it will clash with Python's keyword. Since we only use the .g4 one, you can fix it later.

DimitriPlotnikov:
@kperun 
I not really understand what do you mean with `vrom`. What I expect to see is `from`. If somehow there is a clash with Python, the concrete syntax should use a different name, which is readable.

DimitriPlotnikov:
@PTraeder 
Where do I find tests to the conversion functionality? @kperun 's  proposal with startFrom,endWith is also fine with me.

PTraeder:
@DimitriPlotnikov  tests/UnitSystemTest.py holds tests for the conversions. It looks at the NestPrinter output for the respective AST nodes.

PTraeder:
renamed according to @kperun 's suggestions and cleaned up the typesymbol copying issue

PR #426, opened 2018-09-18
Small updates to the installation and usage documentation

clinssen:
* Small fixes like nest.install() -> nest.Install()

* addressed the fact that module names have to end in "module" if NEST is expected to pick them up for linking (see https://nest.github.io/nest-simulator/extension_modules). This also means that we cannot have a module that's just called "module", so I (confusingly) called it "nestmlmodule" by default; other suggestions welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ cmake -Dwith-nest=<nest_install_dir>/bin/nest-config .
 make all
 make install
 ```
-where `<nest_install_dir>` is the installation directory of NEST (e.g. `/home/nest/work/nest-install`). Subsequently, the module can either be linked into NEST (see [Writing an extension module](https://nest.github.io/nest-simulator/extension_modules)), or loaded dynamically using the `Install` API function. For example, to dynamic load a module with `module_name` = `mymodelsmodule` in PyNEST:
+where `<nest_install_dir>` is the installation directory of NEST (e.g. `/home/nest/work/nest-install`). Subsequently, the module can either be linked into NEST (see [Writing an extension module](https://nest.github.io/nest-simulator/extension_modules)), or loaded dynamically using the `Install` API function. For example, to dynamically load a module with `module_name` = `mymodelsmodule` in PyNEST:
 ```py
 nest.Install("mymodelsmodule")
 ```
@@ -82,7 +82,7 @@ Subsequently, it is possible to call PyNestML from other Python tools and script
 ```py
 to_nest(path, target, dry, logging_level, module_name, store_log, dev)    
 ```
-This operation expects the same set of arguments as in the case of command line invocation. The following default values are used corresponding to the command-line defaults; note that only `path` is mandatory:
+This operation expects the same set of arguments as in the case of command line invocation. The following default values are used, corresponding to the command line defaults. Possible values for `logging_level` are the same as before ('INFO', 'WARNING', 'ERROR', 'NO'). Note that only the `path` argument is mandatory:
 
 | Argument | Type | Default |
 |---       |---   | --- |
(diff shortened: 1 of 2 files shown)


clinssen:
N.B. to see the rendered version of the README, navigate to: https://github.com/clinssen/nestml

README.md
```diff
@@ -18,121 +14,101 @@ an exact solution if possible or use an appropriate numeric solver otherwise.
...
+```py
+from pynestml.frontend.pynestml_frontend import to_nest, install_nest
 ```
+Subsequently, it is possible to call PyNestML from other Python tools and scripts via:
+```py
 to_nest(path, target, dry, logging_level, module_name, store_log, dev)    
 ```
-This operation expects the same set of arguments as in the case of the shell/CMD call,
-with the following default values being used, where only the __path__ is mandatory:
+This operation expects the same set of arguments as in the case of command line invocation. The following default values are used corresponding to the command-line defaults; note that only `path` is mandatory:
```
jougs: Make spelling of `command line` consistent. There's on with and one without hyphen in this sentence.

README.md
```diff
@@ -18,121 +14,101 @@ an exact solution if possible or use an appropriate numeric solver otherwise.
...
+```py
+from pynestml.frontend.pynestml_frontend import to_nest, install_nest
 
-to_nest(path="/home/nest/work/pynestml/models", target="/home/nest/work/pynestml/target",dev=True)
+to_nest(path="/home/nest/work/pynestml/models", target="/home/nest/work/pynestml/target", dev=True)
 
 install_nest("/home/nest/work/pynestml/target", "/home/nest/work/nest-install")
 
-nest.install(<the module name>)    
+nest.Install("mymodelsmodule")
```
jougs: Use name `nestmlmodule` here, as that is the default.

README.md
```diff
@@ -18,121 +14,101 @@ an exact solution if possible or use an appropriate numeric solver otherwise.
...
 make install
 ```
-where _nest\_install\_dir_ points to the installation directory of NEST (e.g. work/nest-install). Subsequently, PyNEST can be used to set up and execute a simulation.
-
-PyNestML is also available as a component and can therefore be used from within 
-other Python tools and scripts. After the PyNestML has been installed,
-the following modules have to be imported:
-```
-from pynestml.frontend.pynestml_frontend import to_nest,install_nest
+where `<nest_install_dir>` is the installation directory of NEST (e.g. `/home/nest/work/nest-install`). Subsequently, the module can either be linked into NEST (see [Writing an extension module](https://nest.github.io/nest-simulator/extension_modules)), or loaded dynamically using the `Install` API function. For example, to dynamic load a module with `module_name` = `mymodelsmodule` in PyNEST:
```
jougs: "dynamic load" -> "dynamically load"

README.md
```diff
@@ -18,121 +14,101 @@ an exact solution if possible or use an appropriate numeric solver otherwise.
...
-with the following default values being used, where only the __path__ is mandatory:
+This operation expects the same set of arguments as in the case of command line invocation. The following default values are used corresponding to the command-line defaults; note that only `path` is mandatory:
 
 | Argument | Type | Default |
 |---       |---   | --- |
-| path     | string | -no default- |  
+| path     | string | _no default_ |
 | target   | string | None |
 | dry      | boolean | False |
 | logging_level | string | 'ERROR' |
```
jougs: Maybe change this to `verbosity` and also mention all possible values?

jougs:
@clinssen: thanks for taking care of this. I'm happy to merge once my minor comments are addressed.

clinssen:
Thanks for your comments, @jougs! The extra commits should address all of the points you raised.

PR #427, opened 2018-09-20
minor updates to antlr4 parser/lexer build script

clinssen:
Minor updates: file name change `script` -> `generate_lexer_parser`

Should we also include a paragraph in the readme to point the user to this script?


clinssen:
I cannot find an obvious place in the README right now to add this information, so I would suggest to go ahead as-is.

clinssen:
Scratch that---found a good place to mention the script, and fixed (all) links inside the documentation to the .g4 grammar file (as it has been renamed)

PR #428, opened 2018-09-20
updates to grammar to successfully generate lexer/parser/visitor with Antlr4.7.1

clinssen:


clinssen:
OK, so this is not exactly the latest version of Antlr4, but the one from the Ubuntu package repository--the actually latest Antlr4 (4.7.1) fails to parse the grammar, and complains about "custom channels are not supported in combined grammars". This might be a separate issue that we have to figure out.

clinssen:
A slight change in the Antlr4 grammar (PyNestML.g4) seems to have fixed the issue.

clinssen:
I filed this PR because I couldn't generate the lexer/parser/visitor.py from the PyNESTML grammar with antlr4.7.1.

In general, Antlr recommends that you separate your lexer grammar and your parser grammar. The lexer grammar is the old `Tokens.g4`, but another (new?) requirement of Antlr is that you explicitly define all your literals. That is, you cannot use '=' in the parser, but have to use, say, `EQUALS`, with `EQUALS : '='` defined in the lexer. See e.g. https://github.com/antlr/antlr4/issues/37

This PR implements the necessary changes. Note that the target language is set to Python2; this is also Python3 compatible (but not vice versa).

I would still like to include lexer/parser/visitor generation as part of the Travis CI build script.

clinssen:
With the latest commit, lexer/parser/visitor generation has been included in the Travis CI build script. This means that during installation, the `pynestml/generated` directory is cleared, and antlr4 is invoked to generate lexer/parser/visitor.py, which are then used by pynestml in the further tests.

Issue #429, opened 2018-10-01
stale version number in message string


clinssen:
https://github.com/nest/nestml/blob/a53867178246ca58fbfed8f9942259f1020b0d45/pynestml/frontend/frontend_configuration.py#L75

clinssen:
@jougs: I know you did some magic extracting the always-latest NEST version number for use in this kind of situation. Unfortunately most of the magic seems to be written in SLI. Is there a design pattern, or similar best practice that we can follow?

jougs:
The code that does the magic is actually a pretty plain bash script: [extras/create_rcsinfo.sh](https://github.com/nest/nest-simulator/blob/master/extras/create_rcsinfo.sh). We set the version to a fixed string and then try to come up with a git tag or latest commit that we can use instead, if the sources are not from a tarball.

clinssen:
What we could do is the following:
- Maintain a NESTML version number in no more than two places, e.g. in the README and in one of the Python files (perhaps as `pynestml.__version__`)
- I will file a PR to set these strings to "3.0-dev" (3.0 is the latest release according to the GitHub releases page). The "dev" ensures that people do not pull the master branch and mistake it for the 3.0 release.
- When we do a release, we file a PR to put the new version number in (e.g. "3.1"), do the release, then file a second PR to add the "-dev" suffix back into master.

The last step is obviously where things get a little tedious, but let me know if you think this is a workable solution, or if you have a much better suggestion altogether!

jougs:
I like your suggestion. Please feel free to implement it. I'll happily review and accept a corresponding PR.

clinssen:
Fixed by PR #433

PR #433, opened 2018-10-23
updated the versioning scheme

clinssen:
Versioning according to PEP 440 (https://www.python.org/dev/peps/pep-0440/).

This fixes issue #429.

PEP 440 defines a lexicographical ordering on version number strings. There are rather a few special cases (such as "dev" and "post") that affect this order. To obtain a version that sits lexicographically _after_ the 3.0 release, but _before_ identifying the version number of the next release, because we don't know yet what that is going to be (3.1? 3.0.1?), we have to use the _post_ keyword:

    >>> from pkg_resources import parse_version as V
    >>> V("3.0") < V("3.0-post")
    >>> True

To make it abundantly clear to the user that they are dealing with a development branch, I've also added "-dev" after the "-post". Technically, this means that this is a pre-release (3.0-post-dev) of the post-release (3.0-post):

    >>> V("3.0") < V("3.0-post-dev") < V("3.0-post")
    >>> True

Versioning upon a release will follow the plan as set out in #429 and treat "-post-dev" as a single keyword.

### DIFF
--- a/pynestml/__init__.py
+++ b/pynestml/__init__.py
@@ -17,3 +17,5 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+__version__ = "3.0-post-dev"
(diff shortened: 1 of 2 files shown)


clinssen:
For future reference: note that the version number also needs to be updated in the `setup.py` script.

PR #436, opened 2018-11-07
target platform command line option

clinssen:
This PR adds a `--target NAME` command line option to PyNESTML, which allows the user to request targets other than NEST, e.g. SpiNNaker, simply from the command line.

This replaces the "dry run" command line parameter: a dry run (no code generated) is now specified by giving "none" or the empty string as a target. The default target is NEST.

Note that this pattern would also allow us to target different installations of NEST, e.g. an HPC versus desktop version by specifying "NEST-HPC" as a target, which might use `TargetIdentifierIndex` instead of `TargetIdentifierPtrRport` as index type.

Code-wise, I've implemented a class called `CodeGenerator` that is subclassed by `NESTCodeGenerator` (and, in the future, other classes for new targets). There is also a static method in `CodeGenerator` called `generate_code()` that takes a `target` string as parameter and instantiates the generator as a local variable.

Much of the transformation that is done between reading the input/solving the equations and the code generation is left in the `NESTCodeGenerator` class as opposed to factoring it out into separate "Transformer"-type classes. This is, for now, by design; see https://github.com/nest/nestml/blob/master/doc/impl/middle.md: "it may be beneficial to regard model transformations as a part of the target-format generating backend and encapsulate all components required for a specific target in a single subsystem." If there should later turn out to be significant overlap between these transformation steps for the NEST target, and e.g. the SpiNNaker target, we can decide to factor it out at that stage.

Minor: this PR also changes all full-word command line options from the "-word" format into the "--word" format to match GNU convention.


clinssen:
Please note that I based this off of the `error_strategy` branch, so the diff will also show changes between that branch and the current master.

clinssen:
Hi @sebastianberns, I did this to disambiguate the target path from the target platform. Aside from having to rename the parameters in your API calls, are you running into any issues with this changeset?

sebastianberns:
Yes, see #442

PR #444, opened 2018-12-10
Removed _nestml suffix from neuron names in models

clinssen:


clinssen:
Hi @jougs, thanks for your review and comment. There is indeed a large amount of duplication in these comment blocks. This is a consequence of the doxygen-style comments layout: an almost-identical neuron, once in "normal" form and once in "implicit" form, is defined twice (two "neuron" blocks in NESTML), and each can be associated with only a single comments block.

I will run through everything again to fix the inconsistencies between name in the comment, and name in the definition. Another thing we could decide to change, is that some neuron names end in "_neuron" (but not others). So we might have a definition in the form of "neuron XXX_neuron:". Shall we consistently eliminate the _neuron suffix from all models?

clinssen:
Note that even when eliminated in the model files, the _neuron suffix still appears in a few filenames (e.g. iaf_neuron.nestml). After merging the synapses feature into NESTML, I propose to add subdirectories `neurons` and `synapses` to the `models` directory. It should then be clear that, e.g. `hill_tononi.nestml` refers to a neuron by virtue of being in the `models/neurons` directory. All the _neuron suffixes can then be removed from the filenames. But we can also eliminate these suffixes right now instead of waiting for the synapses feature. Please let me know which option has your preference.

jougs:
1. IIRC, the difference between the "normal" and the "implicit" versions is that the postsynaptic shapes are expressed as functions of time in one and as ODE in the other. The name implicit was never a good choice due to the potential for confusion with implicit solvers for differential equations. Probably the difference should be rather explained in some piece of documentation and only the more expressive version should be kept in the repository.
1. The suffix `_neuron` is only there to distinguish the models from the ones already present in NEST. That suffix should go away.
1. The `iaf_neuron` is a simplified version of the `iaf_psc_alpha` with a single synaptic time constant instead of one for inhibitory and one for excitatory synapses. That model was removed from NEST in  https://github.com/nest/nest-simulator/pull/526, so I guess it can also be removed from NESTML.
1. I am in favor of renaming the files already in this PR.

clinssen:
Thanks again @jougs for your detailed comments. The new commits implement points 2 through 4. Point 1 ("implicit" versions of models) might require some further discussion, and we can perhaps address this best in a separate PR. (I will put this point on the agenda.)

PR #445, opened 2018-12-13
Refactor unit tests to use the pytest framework

### LINKED ISSUE #443: Travis CI builds skip some unittest tests
The Travis CI builds call the unit tests (in `tests/`) by invoking them one-by-one on the command line. This requires each `test_XXX.py` file to contain the section:

    if __name__ == '__main__':
        unittest.main()

Some are missing this code, therefore not running any test at all, and just exiting with 0 (no error), so Travis reports no build errors.

This caused the regression that was just now fixed by #442.

clinssen:
This fixes #443.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - echo >> antlr4
   - chmod +x antlr4
   - export PATH=$PATH:`pwd`
-  - pip install --upgrade pip enum34
+  - pip install --upgrade pip enum34 pytest
   - pip install -r requirements.txt
   #- pip install matplotlib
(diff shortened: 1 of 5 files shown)


clinssen:
This PR is best summarized by looking at the changeset for .travis.yml:

https://github.com/nest/nestml/compare/master...clinssen:pytest#diff-354f30a63fb0907d4ad57269548329e3

The resulting test files are not yet "pure" pytest: many XxxTest classes still inherit from unittest.TestCase. We make use of pynest's backward compatibility with unittest. The refactoring ensures that all tests are discoverable by pynest and that test setup/teardown is done appropriately.

PR #446, opened 2019-01-07
automated documentation generation for NESTML models

clinssen:
At the time of filing, this PR implements a working example of automated documentation generation. A new code generation target named "autodoc" is defined, that will generate HTML code output for any neuronal model in NESTML format.

This PR will serve as a central place to track and discuss the automated documentation generation feature of NESTML. Upcoming discussion should cover topics such as MarkDown/reStructuredText support, styles, cross-references, etc. Feedback is, as always, welcome!


clinssen:
An example run for the Izhikevich neuron (https://github.com/nest/nestml/blob/master/models/izhikevich.nestml) produces the following HTML output:

autodoc_target_example

clinssen:
After meeting with @jessica-mitchell, we decided to unabashedly support the reStructuredText syntax throughout the NEST and NESTML codebases, when parsing comments. The latest commits change the output format of the "autodoc" target to reStructuredText instead of HTML. The Izhikevich model documentation comment block was edited to follow the rst syntax.

In general, and following [PEP8](https://www.python.org/dev/peps/pep-0008/), two comment styles will be supported:
1. In-line comments, e.g. 

    ``a real = 0.02       # time scale of recovery variable``

2. Block comments, e.g. 

    ``# time scale of recovery variable``
    ``a real = 0.02``

The current format for the main documentation comment block is along the lines of doxygen: a /* ... */-enclosed block just before the `neuron` keyword. It would be more Pythonesque to move this to just inside the neuron definition.

Other outstanding issues include linking from one .rst file to the other.

Example output for the Izhikevich neuron:

autodoc_target_example2

jougs:
What do you think about adding dedicated language constructs for the different parts of the documentation? Maybe a toplevel `description` block with sub-blocks for `related`, `references`, and so on. This would allow to parse and check each of them against the specific syntax expected inside of it.

clinssen:
The last series of commits improves the equation (expression) pretty-printing in LaTeX style. The following is an example when run on the (unmodified) `terub_stn.nestml` model. Some heuristics were programmed in to make the output look better, including support for the Greek alphabet and variable subscripts (`I_syn_in` → `I_{syn,in}`). Physical units (that are included explicitly in the specification of this particular model) are typeset in a grey colour to make the equations easier to read.

Screenshot_20190506_183850

clinssen:
N.B. see comments by @jougs on ReST-style "magic keywords", which could be replaced by the corresponding elements from the NESTML description by the NESTML tooling, in #513.

jougs:
Merging after live review.

PR #447, opened 2019-01-28
add context condition checker for ODE declaration type consistency

clinssen:
Presently, there is no (context condition) check for ODE definitions to verify the consistency of physical units between a left-hand side ODE variable and its right-hand side defining expression. For example, `V_m mV, V_m' = 1.` is presently allowed, even though the rhs is of type "Real".

This PR implements the necessary check. Type errors are now generated for many existing models, because all built-in functions (e.g. max(x, 0)) will return "Real"s, and all their parameters are also expected to be of type Real. For the latter, this is not a problem, because any unit can be cast to a Real. A Real, however, cannot be cast to any other physical Unit. Example output after the patch:

	[9,aeif_cond_alpha, ERROR, [51:4;51:95]]: ODE definition for 'd V_m / dt' has inconsistent units: expected 'mV / s', got 'pA / pF'

In this case, the type system does not deduce that these units are equivalent. This should also be fixed.

N.B. a simple workaround for these typing issues is to allow the "Real" type to be cast to any other unit.


clinssen:
The latest commit (494bfa9) uses astropy.units to provide type equivalence checking. This enables us to write things like `V_m = 1 mV/ms`. NESTML now also understands that 'mV / s' and 'pA / pF' are equivalent, save for a constant.

PR #452, opened 2019-01-31
reduce the number of errors generated during model generation

clinssen:


clinssen:
Make tests more verbose: replace logging level "NO" with "INFO" in all tests, and add `-s` (disable stdout/stderr capturing) to pytest in `.travis.yml`. This is a big help when inspecting the log of a run on Travis CI. This does not change the behaviour of the tests. Tests can still be run locally with low verbosity by simply omitting the `-s` parameter.

clinssen:
As for the design rationale for using a `skip_post_ode_specification_checks_` private member field in the AST visitor class: see The Definitive Antlr4 Reference, Terence Parr, Chapter 7.5: Sharing Information Among Event Methods. This chapter suggests the use of a stack member, but in our case we make do with just a boolean flag.

PR #459, opened 2019-02-19
Synapses feature

clinssen:
This PR adds synapse modeling support to NESTML by adding new grammar rules to the language, and the corresponding infrastructure changes and enhancements to PyNESTML. This builds the basic infrastructure for future plasticity extensions.

This PR includes documentation (see `doc/synapses_in_nestml.rst`) and integration tests (see `tests/static_synapse_test.py` and figure for output).

static_synapse_test

There are a few ancillary changes involved with the synapses feature, some of which have been split off into separate PRs where possible, to keep the interests separated (see #458, #457 and #456). Additional changes:

(50320edf) Avoid calling deepcopy on the log in `pynestml/utils/model_parser.py`. This is an issue when we refer to NESTML keywords in the PyNestMLLexer class (e.g. `PyNestMLLexer.RECORDABLE_KEYWORD`), beause the (generated) PyNestMLLexer class by default directs output to stdout, which leads to deepcopy() trying to copy the stdout file descriptor and failing.


clinssen:
Certain features brought in by this PR, such as the ``get_post_trace()`` predefined function, were inflexible, and the new co-generation approach in #582 avoids the need for such a function entirely.

PR #466, opened 2019-03-06
Add process exit code to PyNESTML

clinssen:
This PR adds an exit code to PyNESTML, such that the process returns 0 for "no errors occurred" and 1 for "errors occurred".

Because normal processing raises a number of errors (inappropriately, see #452), this presently causes the tests to fail. The inappropriately raised errors will be addressed in a future PR, after which this PR can be merged in.


Silmathoron:
@clinssen would it also be possible to include an exit message together with the exit code? Something like "All code was successfully generated, please have fun with you new model!" so that we know unambiguously whether the code was successful
(on python2 most crashes are silent so it's difficult to differentiate between error and success)

clinssen:
HI @Silmathoron, in general I agree with your suggestion. Two small points:
- Python2 will no longer be supported going forward.
- Even if one model gave errors, the module code (which registers each model with the simulation kernel) can still be generated for the remaining models. So we need some additional logic to see if at least one model was successfully generated, or zero, in which case it does not make sense to generate the module code. In practice it's confusing to see "module code successfully generated!" if there are actually zero models that made it.

Silmathoron:
@clinssen ok, I'm must say not supporting Py2 is fine with me ^^
Then I guess a brief summary leading to:

- all models successful
- N models successful, M errors
- code generation failed

could be useful?

clinssen:
@Silmathoron: a brief follow-up: it is difficult to make a unified summary statement at the end of processing, because each NESTML file may contain more than one neuron definition. If a syntax error is present, the file is never fully parsed so we do not know the name(s) of the model(s) that failed. So if we have a summary/overview message at the end, it would have to be in terms of file names, rather than model names.

clinssen:
Note to self: merge #482 first.

Silmathoron:
@clinssen a summary with filenames sounds good enough to me...

jougs:
Can't we either a) say that there only can be one model definition per file (which most sane people will make so anyway) or b) look for further top-level blocks in the remainder of the file in case of errors and resume parsing there? I assume we also resume parsing other files in case we get an error in another file, don't we?

PR #473, opened 2019-03-19
Add parameter and return type templating for predefined functions

clinssen:


clinssen:
Previously, predefined functions such as `max(x, y)` would discard the type of `x` and `y`, converting both to a real value and returning a real. This PR adds a templating feature such that the returned type of the function can be made equal to the type of any parameter.

Each template parameter is defined by a unique nonnegative integer. For example, for the `pow` function, the type for the base is TemplateSymbol(0), the type of the exponent is a real, and the return type is also TemplateSymbol(0), indicating that whatever type will be derived for the base parameter (e.g. mV) is also the type of the return value.

A context condition checker was added to verify that all of the same template parameters (formal type) are consistently typed when the actual type is derived.

pynestml/symbols/predefined_functions.py
```diff
@@ -168,9 +168,9 @@ def __register_exponent_function(cls):
         Registers the exponent (e(X)) function.
         """
         params = list()
-        params.append(PredefinedTypes.get_real_type())  # the argument
+        params.append(PredefinedTypes.get_template_type(0))  # the argument
         symbol = FunctionSymbol(name=cls.EXP, param_types=params,
-                                return_type=PredefinedTypes.get_real_type(),
+                                return_type=PredefinedTypes.get_template_type(0),
```
Silmathoron: in the case of the exponential function (and almost all non-polynomial function), it is correctly defined only if both the entry type and the return type are real.
I.e. in this case the previous code should be kept.

pynestml/symbols/predefined_functions.py
```diff
@@ -246,9 +246,9 @@ def __register_exp1_function(cls):
         Registers the alternative version of the exponent function, exp1.
         """
         params = list()
-        params.append(PredefinedTypes.get_real_type())  # the argument
+        params.append(PredefinedTypes.get_template_type(0))  # the argument
         symbol = FunctionSymbol(name=cls.EXPM1, param_types=params,
-                                return_type=PredefinedTypes.get_real_type(),
+                                return_type=PredefinedTypes.get_template_type(0),
```
Silmathoron: see EXP

pynestml/symbols/predefined_functions.py
```diff
@@ -271,10 +271,10 @@ def __register_max_function(cls):
         Registers the maximum function.
         """
         params = list()
-        params.append(PredefinedTypes.get_real_type())
-        params.append(PredefinedTypes.get_real_type())
+        params.append(PredefinedTypes.get_template_type(0))
+        params.append(PredefinedTypes.get_template_type(0))
         symbol = FunctionSymbol(name=cls.MAX, param_types=params,
-                                return_type=PredefinedTypes.get_real_type(),
+								return_type=PredefinedTypes.get_template_type(0),
```
Silmathoron: here it is fine because ``max`` is a special function

pynestml/symbols/predefined_types.py
```diff
@@ -225,6 +226,15 @@ def get_integer_type(cls):
         """
         return cls.name2type[cls.INTEGER_TYPE]
 
+    @classmethod
+    def get_template_type(cls, i):
+        """
+        Returns a new type symbol for argument type templating. The templates are uniquely identified with an integer number `i`.
```
Silmathoron: I don't understand this integer number part, could you add some additional explanation in the docstring?

pynestml/symbols/real_type_symbol.py
```diff
@@ -105,6 +105,8 @@ def __sub__(self, other):
         return self.binary_operation_not_defined_error('-', other)
 
     def is_castable_to(self, _other_type):
+        if super(RealTypeSymbol, self).is_castable_to(_other_type):
+            return True
```
Silmathoron: again, this looks weird to me...

pynestml/symbols/string_type_symbol.py
```diff
@@ -44,4 +44,6 @@ def __add__(self, other):
         return self.binary_operation_not_defined_error('+', other)
 
     def is_castable_to(self, _other_type):
+        if super(StringTypeSymbol, self).is_castable_to(_other_type):
+            return True
```
Silmathoron: even worse, what does it mean to cast a string to millivolt?

pynestml/symbols/template_type_symbol.py
```diff
@@ -0,0 +1,49 @@
...
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+from pynestml.symbols.type_symbol import TypeSymbol
+
+
+class TemplateTypeSymbol(TypeSymbol):
+    def __init__(self, i):
+        super(TemplateTypeSymbol, self).__init__(name='_template_' + str(i))
+        self.i = i
```
Silmathoron: I would make this name more explicit (and potentially "private", i.e. ``_integer_type`` with an ``@property`` function)

pynestml/symbols/type_symbol.py
```diff
@@ -191,7 +191,10 @@ def is_castable_to(self, _other_type):
         :return: True if castable, otherwise False
         :rtype: bool
         """
-        pass
+        from pynestml.symbols.template_type_symbol import TemplateTypeSymbol
+        if isinstance(_other_type, TemplateTypeSymbol):
+            return True
```
Silmathoron: why not call the method from ``_other_type``? They are not necessarily castable but the relationship is commutative

pynestml/symbols/unit_type_symbol.py
```diff
@@ -156,14 +154,16 @@ def get_conversion_factor(cls, to, _from):
         factor = (_from / to).si.scale
         return factor
 
-    def is_castable_to(self, other_type):
+    def is_castable_to(self, _other_type):
+        if super(UnitTypeSymbol, self).is_castable_to(_other_type):
+            return True
         from pynestml.symbols.real_type_symbol import RealTypeSymbol
-        if other_type.is_instance_of(RealTypeSymbol):
+        if _other_type.is_instance_of(RealTypeSymbol):
```
Silmathoron: again, this is very surprising to me... a real should not be castable to anything but a real

tests/invalid/CoCoFunctionCallNotConsistentWrongArgNumber.nestml
```diff
@@ -28,7 +28,6 @@
 neuron CoCoFunctionCallNotConsistentWrongArgNumber:
     state:
         test integer = max(1,2,3) # wrong number of args
-        tes2 integer = max(10mV,2) # wrong type of args
```
Silmathoron: This should be checked and it should crash

tests/resources/FunctionParameterTemplatingTest.nestml
```diff
@@ -0,0 +1,49 @@
...
+
+    parameters:
+    end
+
+    update:
+        force = max(1 N, 10 N)
+        force = max(1 A, 10 N)
+        force = max(1 N, 10 A)
+        force = max(1 A, 10 A)
+        foo = max(1, 3.14)
```
Silmathoron: I'm unsure about where this is actually tested...
For me this file should generate:

- "fine" for the first and last lines
- "wrong" for all others but with different reasons

Silmathoron:
Just to make sure that my previous comment was clear enough:

- power functions ``x^n`` lead to a unit ``[X]^n``
- non-polynomial functions (``exp``, ``log``...) are dimensionless (they expect dimensionless arguments and their output is also dimensionless)
- special functions (``min``, ``max``) take arguments of the same dimension and return this same dimension

clinssen:
Thanks for your comments!

There is one snag when it comes to the pow() function: in the current framework the return type is specified statically, whereas it should actually be derived dynamically, based on the (semantic) value of the exponent. One potential solution would be to eliminate the pow() function altogether as a predefined function, and instead exclusively use the `**` operator.

Silmathoron:
Right... that sounds indeed reasonable !

jougs:
+1 for the removal of `pow()` in favor of just the `**` operator.

jougs:
Thanks for the work!

PR #474, opened 2019-03-20
Explicitly disable ode-toolbox stiffness testing

clinssen:
Adds installation of GSL and PyGSL to the Travis CI environment, enabling the automated stiffness checker from ode-toolbox.

Unfortunately, stiffness checking has to remain disabled because of implementation gaps in the NESTML/ode-toolbox coupling. These will be addressed in a separate PR. For the time being, an extra boolean parameter is passed to `analysis()` to indicate whether or not stiffness checking should be skipped. This first requires that the necessary support is merged in on the side of ode-toolbox, namely PRs https://github.com/jougs/ode-toolbox/pull/5 and https://github.com/nest/ode-toolbox/pull/12/, in that order.

The upshot of this PR is to explicitly disable stiffness checking, instead of depending on the non-presence of GSL/PyGSL in the runtime environment. These two libraries are optional dependencies of NESTML. If indeed installed, they would otherwise cause the end user to run into exceptions thrown from within ode-toolbox during a NESTML code generation run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,15 @@ before_install:
   - export PATH=$PATH:`pwd`
   - pip install --upgrade pip enum34 pytest
   - pip install -r requirements.txt
+ 
+  # gsl library to support ode-toolbox stiffness testing
+  - wget http://ftp.wrz.de/pub/gnu/gsl/gsl-2.5.tar.gz
+  - tar -xvzf gsl-2.5.tar.gz
+  - cd gsl-2.5
+  - ./configure && make && sudo make install
+  - cd ..
+  - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+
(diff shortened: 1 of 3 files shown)


jougs:
Are the failures caused by the merge of nest/ode-toolbox#12?

clinssen:
Negative: it was a pre-existing issue that was revealed only by updating the NESTML runtime environment to include GSL/PyGSL. This PR fixes things so that even with GSL/PyGSL installed, the stiffness test is skipped when NESTML invokes ode-toolbox. The ultimate underlying issue lies in an incomplete NESTML/ode-toolbox interface and will be addressed in a future NESTML PR.

Silmathoron:
Ok, now it seems to run fine, thanks guys!
Also as a side note, it would be nice to have an "all went well" message at the end of the code generation ;)

PR #479, opened 2019-03-27
add prefix to function calls to permit proper resolution in C++

clinssen:
When using NESTML functions that are called from within the ODE stepping function, the calls could not be resolved because the stepping function is declared as an `extern "C"` and the functions are generated as part of the neuron class. This PR adds the necessary `node.` prefix.

This PR also declares each of these functions are "const", i.e. they should not modify the state of the neuron. I'm not sure if this is warranted, but otherwise we get complaints about violating const correctness elsewhere.

### DIFF
--- a/pynestml/codegeneration/expressions_pretty_printer.py
+++ b/pynestml/codegeneration/expressions_pretty_printer.py
@@ -47,7 +47,20 @@ class ExpressionsPrettyPrinter(object):
             self.types_printer = TypesPrinter()
 
     def print_expression(self, node, prefix=''):
-        # type: (ASTExpressionNode) -> str
+        """Print an expression.
+
+        Parameters
+        ----------
+        node : ASTExpressionNode
+            The expression node to print.
+        prefix : str
+            *See documentation for the function print_function_call().*
+
+        Returns
+        -------
+        s : str
+            The expression string.
+        """
         if node.get_implicit_conversion_factor() is not None:
             return str(node.get_implicit_conversion_factor()) + ' * (' + self.__do_print(node) + ')'
         else:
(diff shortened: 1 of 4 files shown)


pynestml/codegeneration/resources_nest/directives/GSLDifferentiationFunction.jinja2
```diff
@@ -13,19 +13,19 @@ extern "C" inline int {{neuronName}}_dynamics(double, const double ode_state[],
     {%- for function in utils.get_alias_symbols(ode) -%}
       {%- if not function.is_equation() -%}
         {% set declaringExpression = function.get_declaring_expression() -%}
-  double {{names.name(function)}} = {{printerGSL.print_expression(declaringExpression)}};
+  double {{names.name(function)}} = {{printerGSL.print_expression(declaringExpression, prefix="node.")}};
```
Silmathoron: I think this is what makes the compilation crash (that or the next one): one needs to differentiate between predefined function and the class methods

pynestml/codegeneration/expressions_pretty_printer.py
```diff
@@ -73,48 +73,48 @@ def __do_print(self, node):
...
-                if_not = self.print_expression(node.if_not)
+                condition = self.print_expression(node.get_condition(), prefix=prefix)
+                if_true = self.print_expression(node.get_if_true(), prefix=prefix)
+                if_not = self.print_expression(node.if_not, prefix=prefix)
                 return self.reference_converter.convert_ternary_operator() % (condition, if_true, if_not)
         else:
             raise RuntimeError('Unsupported rhs in rhs pretty printer!')
 
-    def print_function_call(self, function_call):
+    def print_function_call(self, function_call, prefix=""):
```
Silmathoron: or one could check here if the function is a predefined function and ignore the prefix in that case (either here or on [L76](https://github.com/nest/nestml/pull/479/files#diff-e2c25bbf6644b9a4c049b155c155df87R76)

PR #482, opened 2019-04-02
Names reference conversion fix

### LINKED ISSUE #481: Detect incorrect variable name
I spent a whole afternoon debugging a very stupid problem (which was therefore very difficult to find): NESTML has a weird behavior when a variable name corresponds to a unit or magnitude prefix.

For instance I was using `k` as function argument and it was replaced by `1.0` in the generated C++ code.

I think it would be useful to check variable names to make sure that they do not correspond to either a unit or a magnitude (or any other predefined symbol), and raise an error if it does.

clinssen:
This fixes #481.

@Silmathoron : could you please check if this works for you?

### DIFF
--- a/pynestml/cocos/co_co_variable_once_per_scope.py
+++ b/pynestml/cocos/co_co_variable_once_per_scope.py
@@ -52,10 +52,10 @@ class CoCoVariableOncePerScope(CoCo):
         for sym1 in scope.get_symbols_in_this_scope():
             if sym1.get_symbol_kind() != SymbolKind.VARIABLE or sym1.is_predefined:
                 continue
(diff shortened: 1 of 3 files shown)


Silmathoron:
@clinssen not at all I'm afraid, but it's become a bit complicated with all the PR going on, so I think it would be better to first correct and merge #479 and #473 
Currently this seems to crash all the units but again, it might come from interferences on my branch from the other PRs

clinssen:
The unit test was set up so that even function parameters and function-local variables were resolvable from the global (neuron) scope. This is no longer the case, which seems appropriate, but I have to double-check the literature on this pattern.

PR #487, opened 2019-04-15
Corrected error message in install_nest

Silmathoron:
There is a copy-paste mistake in the error message of the ``install_nest`` function, which this PR corrects.


clinssen:
merci!

PR #496, opened 2019-06-24
Update setup script and requirements to facilitate distribution

clinssen:
This was done to facilitate distribution of NESTML as a PyPi package: users can now simply run `pip install nestml` to take care of installating NESTML.


jougs:
LGTM. Thanks for the update.

PR #497, opened 2019-06-24
Update setup script and requirements to facilitate distribution

clinssen:


README.md
```diff
@@ -1,97 +1,111 @@
...
+
 ```
-pip install -r requirements.txt
+pip install nestml --user
+```
+
+Optionally, you can run `pip` as root to install NESTML system-wide:
+
 ```
+sudo pip -H install nestml
```
terhorstd: would suggest to write `sudo -H pip install nestml` since `-H` is a flag to sudo, not pip.

README.md
```diff
@@ -1,97 +1,111 @@
...
+git clone https://github.com/nest/nestml
 ```
-Or for Python 3:
+
+First, install the requirements:
+
 ```
-python3 setup.py install --user
+cd nestml
+pip install -r requirements.txt
```
terhorstd: is this really required?! shouldn't this be listed in the setup.py and be installed automatically...?

setup.py
```diff
@@ -17,25 +17,27 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
 from setuptools import setup, find_packages
 import sys
 
+assert sys.version_info.major >= 3, "Python 3 is required to run PyNESTML"
+
```
jougs: Failing on Python < 3 also makes out Python 2.7 build on TravisCI fail. That failure should be counted as a success.

jougs:
All good. Thanks! Can you please also update the PyPI package?

Silmathoron:
I don't think the ``sudo -H`` pqrt was corrected @jougs

jougs:
@Silmathoron, you are right, my bad. @clinssen: Can you please file another PR for the `sudo` part and also clarify the separate installation of the requirements mentioned by @terhorstd?

README.md
clinssen: It looks like running `pip` as root is actively discouraged, because of obvious security risks: anyone can upload malicious code to PyPI, which will then (in `setup.py`) be run as root. Shall I remove the suggestion to `sudo pip`?

jougs: Yes, remove it. I guess that people will find out anyway if they need to.

PR #512, opened 2019-08-06
enhance NEST integration test by comparing NESTML-generated against existing NEST C++ models behaviourally

clinssen:


clinssen:
Testing is not exhaustive but presently covers 22 neuron models. For each neuron model, two spikes are generated as input stimulus. The resulting membrane potential waveform is then compared to the reference, known-good model in native NEST simulator C++. The test fails if the difference in voltage between top and bottom traces exceeds a threshold.

For example, for the aeif_cond_exp model:

nestml_nest_integration_test_ aeif_cond_exp _ aeif_cond_exp_nestml

clinssen:
N.B. code in the file `tests/nest_tests/nest_integration_test_multisynapse.py` is retained for posterity but not presently functional, while we work on fixing multisynapse support.

Issue #368, opened 2017-06-30
Discussion of possible types for buffers


PTraeder:
We are planning to add typing to buffers.
Current buffers are intended to carry the type `pA`.
Concerning spike buffers, we would like to have feedback for types you consider probable/useful.
For example, would you like to see just a specific subset of units? All of the "native" NEST units? 
Compound units such as `mV/ms`? Primitive types?

Additionally, your ideas about possible syntax for these types would be welcome.

Currently buffers are defined thusly:
```
bufferName <- (inhibitory | exitatory)* (spike | current) 
```
My proposition, in accordance with suggestions made by @heplesser, would be to simply add the respective type after the buffer name. Similarly to variable definitions.
```
someBuffer1 mV  <- exitatory spike 
someBuffer2 nS  <- exitatory spike 
```
for a buffer of type ms.

PTraeder:
@heplesser @jougs @Silmathoron @DimitriPlotnikov

heplesser:
I assume that the buffers you mention here are the `RingBuffer`s for spikes and currents in the models.

1. Exactly what these buffers store depends on the model, so the user needs to be able to choose the unit freely.
1. For currents, pA is the most likely unit; I cannot immediately think of another unit that would be reasonable here.
1. For spikes, it depends a bit on how NESTML interprets these buffers. 
    1. If they are interpreted as pure spike storage, then they contain spike counts in bins of a given width, the simulation resolution, and the buffers would be unitless.
    1. If they are interpreted as the buffers are implemented in NEST, i.e., a summed input (at each point in time) weighted with synaptic weights, then the buffers have the same unit as the synaptic weights.
    1. The unit of the synaptic weight is determined by the model: 
        - For `_psc_` models, it is pA
        - Except for `_psc_delta` models, where it is mV
        - For `_cond_` models it is nS
        - Other units may apply to other models, especially binary neurons.

Silmathoron:
Just in case, I'm also referencing the [post that summarizes a long discussion on how we might simplify the interaction between spikes and state variables](https://github.com/nest/nestml/issues/330#issuecomment-266700499), just so that the unit discussion does take into account the way the buffer might be used.

Because of this, I would be in favour of dimensionless buffers where the unit is applied depending on the target variable on which it will be applied, plus potentially through the effect of a normalization factor.

DimitriPlotnikov:
[...]

PR #515, opened 2019-08-19
Improved analysis and code generation for dynamical equations (ODEs and shapes)

### LINKED ISSUE #453: Make shape/ODE transformer more generic
Currently, the `transform_shapes_and_odes()` function has several special cases, for example, the one where there is exactly one shape with an exponentially decaying kernel, in combination with precisely one linear ODE with constant coefficients:

https://github.com/nest/nestml/blob/3204d21aeb8ee8cba290812c3891c4068b13a5c2/pynestml/codegeneration/nest_codegenerator.py#L282

Can we make this more generic, e.g. when two "delta" shapes are supplied, or another shape or ODE is present?

Secondly, is the name "delta" appropriate here, as it seems to refer to an exponential decay kernel with given time constant? For a usage example, see:

https://github.com/nest/nestml/blob/3204d21aeb8ee8cba290812c3891c4068b13a5c2/models/iaf_psc_delta.nestml#L82

### LINKED ISSUE #478: Shapes and functions
I am currently trying to understand why, now that ``function``s exist, we still need the ``shape`` object.

I see two main issues with it:

- it is less intuitive than ``function`` for new users
- it does not accept a return type

This is typically problematic when used with the ``convolve`` function because types are inconsistent unless the proper unit is added.

Shouldn't we just get rid of the ``shape`` and replace it with ``function`` altogether?
(I guess we can still implement the proposed template in #373 to functions instead of shapes)

### LINKED ISSUE #489: Handling adaptive LIF model
I come across a bug in generating an adaptive LIF model with delta synapses (see NESML file attached). 
The neuron model I am trying to get is a simplified version of adEx. 
V = -V/tau_m + 1/C_m * (I_spikes+ I_e + currents - w)
w' = -w/tau_w 
where **w** is increased by **b** after each spike. 

In the cpp file, NESTML outputs lines of code that initialize GSL integration. Yet, in the update step, there is an attempt to solve the equations analytically ignoring **w**. Also, NESTML substitutes **b** with ones. 

I use NESTML v.3, commit a41eb31849b107b5dafa51fe1d1408791bbc25b7

We went through the code with @Silmathoron and he confirms that it's a bug.
[adapt_lif.txt](https://github.com/nest/nestml/files/3144393/adapt_lif.txt)

### LINKED ISSUE #492: Handling delta function properly
The NESTML predefined `delta` function, which is used in convolve calls, has two parameters: time and a time constant. It does not make sense for a delta function to be parameterised by a time constant. This constant is apparently assumed to be equal to the time constant of the membrane potential ODE in https://github.com/nest/nestml/blob/73f484e260d8026796c610e21f727b6c52ac360d/pynestml/solver/solution_transformers.py#L76

`ode_var_update_instructions` is e.g. 
```py
V_abs = V_abs * exp(-__h / tau_m)
V_abs += (tau_m / C_m - tau_m * exp(-__h / tau_m) / C_m) * (I_e + currents - w)
V_abs += spikes
```
Note the `tau_m` in the first line, which comes from the NESTML model specification `shape G = delta(t, tau_m)` in iaf_psc_delta (which is apparently the only model included that uses the delta function).

### LINKED ISSUE #510: Make ``internals`` block parsed first?
Currently, when using the ``internals`` block, the block must be placed above the first block which makes use of the variables declared in it, so my guess is that blocks are parsed sequentially.
However, the ``internals`` block is not very important and it's annoying to have it before more relevant blocks... how difficult would it be to parse it first?
(if I remember correctly we're parsing tree so it should be trivial, but I just want to make sure)

### LINKED ISSUE #564: Issues in gsl_reference_converter script
On latest master (4fa0c499f1379cda89722b3dcf55dfc75a9c19f3), there are several issues in the GSL script at ``pynestml/codegeneration/gsl_reference_converter.py``:

* it is missing ``from pynestml.utils.logger import Logger, LoggingLevel``
* nonexisting ``variable`` is accessed on [lines 69-70](https://github.com/nest/nestml/blob/master/pynestml/codegeneration/gsl_reference_converter.py#L69); I would change it to ``ast_variable`` I think but I'm not sure...

EDIT: after testing, ``ast_variable`` seems to be the way to go @clinssen do you want me to make a PR or are you already working on this file?

clinssen:
The branch [ode-toolbox:merge_shape_ode_concepts-dev](https://github.com/nest/ode-toolbox/compare/master...clinssen:merge_shape_ode_concepts-dev) contains a major refactoring of ode-toolbox, improving the API and making it more generic. This PR proposes the corresponding changes in NESTML, to work with the new ode-toolbox. Most of the changes pertain to processing of ODEs and shapes, and processing the (analytic and numeric) solvers returned by ode-toolbox.

Fixes #492.
Fixes #478: "shapes" will remain separate entity from "function" for now, with improvements in documentation describing differences and use cases.
Fixes #453.
Fixes #489.
Fixes #510.
Fixes #564.

Removes `curr_sum` and `cond_sum` predefined functions, as they are not used anywhere in the NESTML models library, and seem to have been succeeded by the `convolve` function. Discussion welcome on this point.


models/iaf_chxk_2008.nestml
```diff
@@ -27,7 +27,8 @@ neuron iaf_chxk_2008:
   initial_values:
     V_m mV = E_L   # membrane potential
     G_ahp nS = 0 nS      # AHP conductance
-    G_ahp' nS/ms = e / tau_ahp * nS   # AHP conductance
+    G_ahp' nS/ms = 0 nS/ms   # AHP conductance
+    #G_ahp' nS/ms = e / tau_ahp * nS   # AHP conductance
```
Silmathoron: why is this commented? if there is a reason for keeping it, then an explanatory comment would be nice, otherwise I think it should just be removed

models/iaf_cond_alpha.nestml
```diff
@@ -31,7 +31,7 @@ Author: Schrader, Plesser
 
 SeeAlso: iaf_cond_exp, iaf_cond_alpha_mc
 */
-neuron iaf_cond_alpha:
+/*neuron iaf_cond_alpha:
```
Silmathoron: wouldn't it be better to separate the models into two files and keeping them both available?

models/iaf_psc_alpha.nestml
```diff
@@ -84,32 +84,32 @@ neuron iaf_psc_alpha:
   end
 
   equations:
-    shape I_shape_in = pA * (e/tau_syn_in) * t * exp(-1/tau_syn_in*t)
-    shape I_shape_ex = pA * (e/tau_syn_ex) * t * exp(-1/tau_syn_ex*t)
+    shape I_shape_in = (e / tau_syn_in) * t * exp(-t / tau_syn_in)
```
Silmathoron: so do we now enforce the fact that ``shape`` objects have no dimension? if so it should appear in the documentation (and be properly checked)

doc/nestml_language.md
```diff
@@ -459,7 +463,7 @@ These functions can be used to implement custom buffer lookup logic but should b
 
 ## Equations
 
-### Synaptic response
+### Shape functions
```
Silmathoron: Details required regarding the dimension of shape objects (later modification to NESTML code are ambiguous)

models/iaf_psc_alpha.nestml
```diff
@@ -84,32 +84,32 @@ neuron iaf_psc_alpha:
...
     RefractoryCounts integer = steps(t_ref) # refractory time in steps
   end
 
   input:
-    ex_spikes pA  <- excitatory spike
-    in_spikes pA  <- inhibitory spike
-    currents    <- current
+    ex_spikes pA <- excitatory spike
+    in_spikes pA <- inhibitory spike
+    currents <- current
```
Silmathoron: unrelated, but since I started talking about dimensions, it would be nice if a buffer called ``current`` had one (e.g. ``pA``) ;)

models/iaf_psc_delta.nestml
```diff
@@ -79,8 +78,8 @@ neuron iaf_psc_delta:
   end
 
   equations:
-    shape G = delta(t, tau_m)
-    V_abs' = -1/tau_m * V_abs + 1/C_m * (convolve(G, spikes) + I_e + currents)
+    shape G = 42 * delta(t)
```
Silmathoron: x)

models/lorenz_attractor.nestml
```diff
@@ -0,0 +1,31 @@
...
+parameters:
+  sigma real = 10
+  beta real = 8/3
+  rho real = 28
+end
+
+input:
+end
+
+output: spike
```
Silmathoron: is it possible to make a model without output? would be cleaner

models/lorenz_attractor.nestml
```diff
@@ -0,0 +1,31 @@
...
+  integrate_odes()
+end
+
+parameters:
+  sigma real = 10
+  beta real = 8/3
+  rho real = 28
+end
+
+input:
```
Silmathoron: does this need to be specified? (I guess so)
it's a bit ugly so it would be nice if we could remove it and make it empty by default

pynestml/cocos/co_co_convolve_cond_correctly_built.py
```diff
@@ -47,13 +47,12 @@ def check_co_co(cls, node):
...
     """
-    Visits a function call and checks that if the function call is a cond_sum,cur_sum or convolve, the parameters
-    are correct.
+    Visits a function call and checks that if the function call is a convolve, the parameters are correct.
     """
 
     def visit_function_call(self, node):
         func_name = node.get_name()
-        if func_name == 'convolve' or func_name == 'cond_sum' or func_name == 'curr_sum':
+        if func_name == 'convolve':
```
Silmathoron: I guess I would add a test here to check that if ``symbol_var.is_shape()`` then it is ``real``

pynestml/cocos/co_co_sum_has_correct_parameter.py
```diff
@@ -57,8 +57,7 @@ def visit_function_call(self, node):
         :type node: ast_function_call
         """
         f_name = node.get_name()
-        if f_name == PredefinedFunctions.CURR_SUM or \
-                f_name == PredefinedFunctions.COND_SUM or f_name == PredefinedFunctions.CONVOLVE:
+        if f_name == PredefinedFunctions.CONVOLVE:
```
Silmathoron: or do the check for dimensionless ``shape`` here...

pynestml/codegeneration/nest_codegenerator.py
```diff
@@ -269,172 +380,335 @@ def is_functional_shape_present(self, shapes):
...
-    def transform_shapes_and_odes(self, neuron, shape_to_buffers):
+    def ode_toolbox_analysis(self, neuron, shape_buffers):
         # type: (ASTNeuron, map(str, str)) -> ASTNeuron
         """
         Solves all odes and equations in the handed over neuron.
 
         Precondition: it should be ensured that most one equations block is present.
 
         :param neuron: a single neuron instance.
-        :param shape_to_buffers: Map of shape names to buffers to which they were connected.
```
Silmathoron: don't remove but update the parameter description

pynestml/meta_model/ast_neuron.py
```diff
@@ -745,3 +764,31 @@ def equals(self, other):
...
+        return None
+
+    def get_shape_by_name(self, shape_name):
+        assert type(shape_name) is str
+        shape_name = shape_name.split("__X__")[0]
+        for decl in self.get_equations_block().get_declarations():
+            if type(decl) is ASTOdeShape and decl.get_variable().get_name() == shape_name:
+                print("Is shape " + str(shape_name) + "? YES")
+                return decl
+        print("Is shape " + str(shape_name) + "? NO")
```
Silmathoron: Are these prints necessary?

pynestml/meta_model/ast_ode_shape.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, lhs, rhs, source_position):
         self.lhs = lhs
         self.rhs = rhs
         return
-
+
```
Silmathoron: remove whitespaces

pynestml/solver/transformer_base.py
```diff
@@ -67,23 +67,23 @@ def add_declaration_to_internals(neuron, variable_name, init_expression):
     return neuron
 
 
-def add_declarations_to_initial_values(neuron, declarations):
+def add_declarations_to_initial_values(neuron, variables, initial_values):
     # type: (ASTNeuron, map(str, str)) -> ASTNeuron
     """
     Adds a single declaration to the initial values block of the neuron.
     :param neuron: a neuron
     :param declarations: a single
```
Silmathoron: update parameters

pynestml/solver/transformer_base.py
```diff
@@ -232,20 +171,22 @@ def add_declaration_to_update_block(declaration, neuron):
...
+    stmt.update_scope(neuron.get_update_blocks().get_block().get_scope())
     return neuron
 
 
-def add_state_updates(state_shape_variables_updates, neuron):
+def add_state_updates(neuron, update_expressions):
     # type: (map[str, str], ASTNeuron) -> ASTNeuron
     """
     Adds all update instructions as contained in the solver output to the update block of the neuron.
     :param state_shape_variables_updates: map of variables to corresponding updates during the update step.
```
Silmathoron: update parameters

pynestml/symbol_table/scope.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, scope_type, enclosing_scope=None, source_position=None):
         self.scope_type = scope_type
         self.enclosing_scope = enclosing_scope
         self.source_position = source_position
+        print("------- CREATING SCOPE " + str(self) + " ---------")
+        #import pdb;pdb.set_trace()
```
jougs: This seems to be only for debugging. I suggest to remove it.

pynestml/symbols/variable_symbol.py
```diff
@@ -241,6 +244,9 @@ def is_shape(self):
         :return: True if part of a shape definition, otherwise False.
         :rtype: bool
         """
+        #print("is_shape(" + self.name + ") ? --> " + str(self.variable_type == VariableType.SHAPE))
+        #if not(self.variable_type == VariableType.SHAPE):
+            #import pdb;pdb.set_trace()
```
jougs: Please remove.

pynestml/utils/messages.py
```diff
@@ -257,6 +257,7 @@ def get_no_variable_found(cls, variable_name):
         assert (variable_name is not None and isinstance(variable_name, str)), \
             '(PyNestML.Utils.Message) Not a string provided (%s)!' % type(variable_name)
         message = 'No variable \'%s\' found!' % variable_name
+        #import pdb;pdb.set_trace()
```
jougs: Remove.

pynestml/utils/messages.py
```diff
@@ -822,6 +823,7 @@ def get_could_not_resolve(cls, name):
         """
         assert (name is not None and isinstance(name, str)), \
             '(PyNestML.Utils.Message) Not a string provided (%s)!' % type(name)
+        import pdb;pdb.set_trace()
```
jougs: Remove.

pynestml/utils/ode_transformer.py
```diff
@@ -29,19 +29,18 @@ class OdeTransformer(object):
...
-        :param _ast: a single node
-        :type _ast: ASTNode
-        """
+    #@classmethod
+    #def refactor_convolve_call(cls, _ast):
+        #"""
+        #Replaces all `convolve` calls in the handed over node.
+        #:param _ast: a single node
+        #:type _ast: ASTNode
+        #"""
```
jougs: Remove.

pynestml/utils/ode_transformer.py
```diff
@@ -29,19 +29,18 @@ class OdeTransformer(object):
...
 
-        function_calls = cls.get_sum_function_calls(_ast)
-        for call in function_calls:
-            cls.replace_function_call_through_first_argument(_ast, call)
+        #function_calls = cls.get_sum_function_calls(_ast)
+        #for call in function_calls:
+            #cls.replace_function_call_through_first_argument(_ast, call)
```
jougs: Remove.

pynestml/visitors/ast_symbol_table_visitor.py
```diff
@@ -58,11 +58,8 @@ def visit_neuron(self, node):
         code, message = Messages.get_start_building_symbol_table()
         Logger.log_message(neuron=node, code=code, error_position=node.get_source_position(),
                            message=message, log_level=LoggingLevel.INFO)
-        # before starting the work on the neuron, make everything which was implicit explicit
-        # but if we have a model without an equations block, just skip this step
-        if node.get_equations_blocks() is not None:
-            make_implicit_odes_explicit(node.get_equations_blocks())
         scope = Scope(scope_type=ScopeType.GLOBAL, source_position=node.get_source_position())
+        print("In ASTSymbolTableVisitor::visit_neuron(): neuron = " + node.get_name() + ", scope = " + str(scope))
```
jougs: Tracing? Remove!

models/lorenz_attractor.nestml
Silmathoron: @jougs fair point about it being a bad idea if it could happen by accident...

models/lorenz_attractor.nestml
Silmathoron: @jougs what is the output of rate neurons in NEST? (my guess is that it is a continuous variable)
We might want to keep ``output`` for later but make it optional, with ``spike`` as the default output

jougs: Sorry for being unclear (or too short). My complete idea was to
1. remove the `output` block
2. add functions `emit_rate()`, `emit_current()` and similar once we need them
3. forbid the use of more than one of them in a single model
4. infer the type of output from the `emit_` function used
5. create the corresponding code for the connection handshake
6. probably warn if none of the `emit_` functions is used.

clinssen:
@Silmathoron : I'm trying to reverse engineer the following ODE notation for synaptic conductances in the iaf_cond_beta and hill_tononi neuron models for a bi-exponential ("beta") postsynaptic conductance:

```
  equations:
    g'' = -g' / tau_syn_rise
    g' = g' - g / tau_syn_decay
  end
  
  update:
    g' += spikes * normalisation_const
    [...]
  end
```

The left-hand sides of the top two equations do not really make sense to me, as it should logically be the case that `(d/dt) * g_in' =  g_in''`, but this equality does not seem to hold for the given definitions.

I expect that these equations came about by rewriting the second order ODE into a system of two first-order equations. However, then the logical notation to me would be to introduce a new variable, say `g' = h` which implies `g'' = h'` (a notation also used by Van Rossum in the cited book chapter), as follows:

```
  equations:
    g' = -g / tau_syn_decay + h
    h' = -h' / tau_syn_rise
  end
  
  update:
    h += spikes * normalisation_const
    [...]
  end
```

Right now my plan is to disallow the first notation and rewrite the models to use the second, but perhaps I'm missing something and it would be great to get your input on this.

Silmathoron:
Hey @clinssen to be honest, I've always hated this implementation which I think looks terribly misleading, but it is indeed what is written in the Roth and Rossum book.
The proper formulation is indeed as you wrote in the last code block, however the change of variable is much more complicated: ``h = g' + g/tau_d``
Then, deriving both expressions for ``h`` and stating that they must be equal, you get the proper 2nd order equation:
``g'' = -g' * (1/tau_r + 1/tau_d) - g / (tau_r*tau_d)``

clinssen:
After discussion with @jougs, we noticed that there is one thing missing from the examples above: the "shape" keyword. There are three cases that we should support: direct definition of the function in time, second-order differential equation, and system of first-order differential equations. (See below for examples.) In the latter case, we need a system for figuring out which variables belong "together" to one shape. That is, if we define `g` and `h` as above, we do not initially know that a convolution with `g` means that also the `h` variable has to be processed in relation to the spike buffer.

Because of this reason, and because we now have a simple and clear definition of that a "shape" is (a kernel that can be used as an argument in a convolution function call), I propose the following syntax, which allows multiple declarations to follow a single "shape" keyword, acting as a binding mechanism that indicates that the variables that follow (here, `g` and `h`) belong to the same kernel shape.

Direct function of time
--------------------
```
shape g_gap = g_gap_const * (exp(-t/tau_gap_decay) - exp(-t/tau_gap_rise))
```

Second-order differential equation
-----------------------------
As per your derivation:
```
shape g'' = -g' * (1/tau_r + 1/tau_d) - g / (tau_r*tau_d)
```

Coupled system of first-order equations
-------------------------

```
shape g_ex' = -g_ex / tau_syn_decay_E + h_ex,
      h_ex' = -h_ex / tau_syn_rise_E
```

The user can use the dollar symbol "$" to pick names for new variables introduced while rewriting the higher-order equation to a system of first-order equations. This is purely a convention. (We note that the dollar symbol is already a legal character in variable names according to the NESTML grammar.)

```
shape g_in' = -g_in / tau_syn_decay_I + g_in$,
      g_in$' = -g_in$ / tau_syn_rise_I
```

Silmathoron:
@clinssen true, I like the proposed solution (though I'm not a huge fan of the dollar variant)
Quick question: is odetoolbox actually able to handle a 2nd order ODE like beta function?

doc/nestml_language.md
```diff
@@ -532,6 +523,15 @@ has to be stated in the `initial_values` block. If the initial values are not de
...
 
+### Inline expressions
+
+In the `equations` block, inline expressions may be used to reduce redundancy, or improve legibility in the model code. An inline expression is a named expression, that will be "inlined" (effectively, copied-and-pasted in) when its variable symbol is mentioned in subsequent ODE or shape expressions. In the following example, the inline expression `h_inf_T` is defined, and then used in an ODE definition:
+
+```nestml
+inline h_inf_T real = 1 / (1 + exp((V_m / mV + 83) / 4))
+IT_h' = (h_inf_T * nS - IT_h) / tau_h_T / ms
+```
+
```
jougs: :+1:

pynestml/cocos/co_co_convolve_cond_correctly_built.py
clinssen: Right now, in an expression ``convolve(g, spikes_in)``, where ``g`` is the shape and ``spikes_in`` is the spike input port, the type of the expression is equal to the type of the second argument. The type is thus "carried" by the input port, for example if we have a definition ``spikes_in nS <- spike`` then any convolution with ``spikes_in`` will have type nS. The type of the first argument is ignored.

This is suboptimal, and we could change this around to use the type of the first argument (the shape variable) instead, while leaving spike input ports untyped (you would just write ``spikes_in <- spike``). The type of the convolution is then that of the shape state variable, e.g. nS if we define our shape as follows:

```
g nS = 0 nS
g' nS/ms = g_E_const * (1 / tau_syn_rise_E - 1 / tau_syn_decay_E) * nS

shape g'' = -(g + g' * (tau_syn_decay_E + tau_syn_rise_E)) / (tau_syn_decay_E * tau_syn_rise_E)
```

clinssen:
Notes for reviewers:
- Shapes and shape variables should be of type "real" (or 1/s, or 1/s^2 etc. for differential definitions). This is now checked by CoCoShapeType. See https://github.com/nest/nestml/issues/368 for earlier discussion.
- The directory ``pynestml/solver`` has been removed as all of the functionality was taken over by ODE-toolbox. Some helper functions were moved to ``pynestml/codegeneration/ast_transformers.py``, but some of these could also be written as ASTNode (mostly, ASTNeuron) methods instead.

pynestml/cocos/co_co_all_variables_defined.py
```diff
@@ -76,8 +76,8 @@ def check_co_co(cls, node):
                  and symbol.block_type != BlockType.INPUT_BUFFER_SPIKE \
                  and not symbol.get_referenced_object().get_source_position().is_added_source_position():
                     # except for parameters, those can be defined after
-                    if (not symbol.get_referenced_object().get_source_position().before(var.get_source_position()) and
-                            symbol.block_type != BlockType.PARAMETERS):
+                    if ((not symbol.get_referenced_object().get_source_position().before(var.get_source_position()))
+                     and (not symbol.block_type in [BlockType.PARAMETERS, BlockType.INTERNALS])):
```
Silmathoron, line 80: isn't this indentation a bit weird?
Does this also fix #510?

pynestml/cocos/co_co_convolve_cond_correctly_built.py
```diff
@@ -26,13 +26,13 @@
 
 class CoCoConvolveCondCorrectlyBuilt(CoCo):
     """
-    This coco ensures that convolve and cond/curr sum are correctly build, i.e.,
-    that the first argument is the variable from the initial block and the second argument an input buffer.
+    This coco ensures that convolve function call correctly built, i.e., that the first argument is the variable from the initial block and the second argument an input buffer.
```
Silmathoron, line 29:
```suggestion
    This coco ensures that ``convolve`` is correctly called, i.e. that the first argument is the variable from the initial block and the second argument is an input buffer.
```

pynestml/cocos/co_co_convolve_cond_correctly_built.py
```diff
@@ -26,13 +26,13 @@
...
+
     Allowed:
-        function I_syn_exc pA =   convolve(g_ex, spikesExc) * ( V_bounded - E_ex )
-    Not allowed:
-        function I_syn_exc pA =   convolve(g_ex, g_ex) * ( V_bounded - E_ex )
+        inline I_syn_exc pA = convolve(g_ex, spikesExc) * ( V_m - E_ex )
 
+    Not allowed:
+        inline I_syn_exc pA = convolve(g_ex, g_ex) * ( V_m - E_ex )
```
Silmathoron, line 35:
```suggestion
        inline I_syn_exc pA = convolve(g_ex, g_ex) * ( V_m - E_ex )
        inline I_syn_exc pA = convolve(spikesExc, g_ex) * ( V_m - E_ex )
```

pynestml/frontend/pynestml_frontend.py
```diff
@@ -123,6 +123,7 @@ def main():
         print('Not a valid path to model or directory: "%s"!' % FrontendConfiguration.get_path())
         return 1
     # after all argument have been collected, start the actual processing
+    sys.setrecursionlimit(10000)
```
Silmathoron, line 126: why is that necessary? (if it really is, then please add an explanatory comment)

pynestml/generated/PyNestMLLexer.interp
```diff
@@ -0,0 +1,270 @@
+token literal names:
+null
```
Silmathoron, line 2: why is ``null`` repeated?

pynestml/generated/PyNestMLLexer.py
```diff
@@ -1,16 +1,15 @@
...
 from __future__ import print_function
 from antlr4 import *
 from io import StringIO
 import sys
 
 
-
 def serializedATN():
     with StringIO() as buf:
         buf.write(u"\3\u608b\ua72a\u8133\ub9ed\u417c\u3be7\u7786\u5964\2")
```
Silmathoron, line 11: I don't think the unicode marker is required anywhere

models/hh_cond_exp_destexhe.nestml
```diff
@@ -137,9 +137,9 @@ neuron hh_cond_exp_destexhe:
   end
 
   input:
-      spikeInh nS <- inhibitory spike
-      spikeExc nS <- excitatory spike
-      currents pA <- current
+    spikeInh nS <- inhibitory spike
+    spikeExc nS <- excitatory spike
+    I_stim pA <- current
```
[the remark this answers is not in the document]
clinssen: I had to double-check the diff of #526, but it looks like we removed ``currents`` in favour of ``I_stim`` instead!

pynestml/cocos/co_co_all_variables_defined.py
clinssen: Are you calling my personal indentation style weird? :)

I think your suggestion of a (partial) PEP8 enforcement would definitely be good, but I would propose creating a separate issue for that, because it's not conceptually related to this PR.

You're right, this fixes #510 also!

Silmathoron: > Are you calling my personal indentation style weird? :)

;P

> I think your suggestion of a (partial) PEP8 enforcement would definitely be good, but I would propose creating a separate issue for that, because it's not conceptually related to this PR.

Agreed

> You're right, this fixes #510 also!

Great!

models/hh_cond_exp_destexhe.nestml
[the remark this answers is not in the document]
Silmathoron: OK, sorry, I got confused...
Checking again the files, there are then a couple of instances to cleanup to ``I_stim``:
* [doc/nestml_language.rst](https://github.com/nest/nestml/pull/515/files#diff-000fcb2a2f06ecb7bc4104fab6456b82R686)
* [pynestml/grammars/PyNestMLParser.g4](https://github.com/nest/nestml/pull/515/files#diff-7afef42ed256049b86abd2684be2f454R252)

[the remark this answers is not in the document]
clinssen: Addressed by 4aa8fbc9a40ac0f7016986e667d56940ceb19d81.

pynestml/frontend/pynestml_frontend.py
clinssen: The default Python recursion limit is 1000, which might not be enough in practice if you are running an AST visitor on a deep tree, e.g. containing an automatically generated expression. I will add this as a comment.

pynestml/generated/PyNestMLLexer.interp
clinssen: No idea, this file is auto-generated by Antlr.

pynestml/generated/PyNestMLLexer.py
clinssen: This file is also auto-generated by Antlr.

clinssen:
Please see all but the last two commits in #562 for the diffs related to making the update block (and with that, all blocks) optional.

clinssen:
Remaining issues that will be turned into GitHub Issues after merge:
- Rename shape to kernel
- Minimal PEP8/linting in CI (#559)
- Some more cleanup of the jinja2 templates/generated code
- Add/fix remaining models in NEST integration test
- Add coco to log an error if equations defined but integrate_odes() never called
- Add unit test for "recordable inline"

PR #516, opened 2019-09-03
Rename natural logarithm function to ln() and disambiguate base-10 log function by renaming to log10()

clinssen:
Includes an integration test, i.e. target code is generated for the NESTML functions `ln()` and `log10()`, which is then built on the target platform, and numerically compared to a reference timeseries. Includes an expected-to-fail integration test.

### DIFF
--- a/pynestml/symbols/predefined_functions.py
+++ b/pynestml/symbols/predefined_functions.py
@@ -32,7 +32,8 @@ class PredefinedFunctions(object):
         PRINT                 The callee name of the print function.
         PRINTLN               The callee name of the println function.
         EXP                   The callee name of the exponent function.
-        LOG                   The callee name of the logarithm function.
+        LN                    The callee name of the natural logarithm function, i.e. the logarithm function of base :math:`e`.
+        LOG                   The callee name of the logarithm function of base 10.
         COSH                  The callee name of the hyperbolic cosine.
         SINH                  The callee name of the hyperbolic sine.
         TANH                  The callee name of the hyperbolic tangent.
@@ -57,7 +58,8 @@ class PredefinedFunctions(object):
     PRINT = 'print'
     PRINTLN = 'println'
     EXP = 'exp'
-    LOG = 'log'
+    LN = 'ln'
+    LOG10 = 'log10'
     COSH = 'cosh'
     SINH = 'sinh'
     TANH = 'tanh'
@@ -88,7 +90,8 @@ class PredefinedFunctions(object):
         cls.__register_print_function()
         cls.__register_print_ln_function()
         cls.__register_exponent_function()
-        cls.__register_log_function()
+        cls.__register_ln_function()
+        cls.__register_log10_function()
         cls.__register_cosh_function()
         cls.__register_sinh_function()
         cls.__register_tanh_function()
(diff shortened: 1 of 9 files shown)


Silmathoron:
Shouldn't we use ``log`` by default for the Neperian logarithm, as seems to be standard in all computational toolboxes? (numpy, C++ std...)

clinssen:
I had never heard of the Napierian logarithm before, interesting Wikipedia hole... but could it be that you meant the natural logarithm instead (to base *e*)?

My reason for the change is to make the predefined functions more self-documenting. The use of just "log" for a name will always be ambiguous and require a documentation lookup to find the intended base (often, 10 or *e*--see e.g. Google for the query "log(10)"). This PR disambiguates the intended logarithm completely on the NESTML syntax end (forcing the use of `log10()` and `ln()`), instead of leaving it up to the platform implementation to define which base is intended (removing `log()`).

Silmathoron:
@clinssen the Napierian logarithm is the "natural" logarithm (basis e), the decimal one was introduced later by Henry Briggs.
And yes, that is what I am referring to and its standard notation is "log", see for instance:

- [numpy.log](https://docs.scipy.org/doc/numpy/reference/generated/numpy.log.html)
- [std::log](https://en.cppreference.com/w/cpp/numeric/math/log)

Therefore, I would rather go for the standard notation which is ``log`` for natural logarithm and ``log10`` for decimal: we have a python-like syntax and are converting into C++, with both having the same standard, I am not sure it is wise to define our function differently, especially since people using NEST are usually familiar with either or both.

clinssen:
My TI-83 and Google use base 10. :) 

The thing is that the base to use for `log()` is defined by convention, it's not universally defined. `log10()` and `ln()` are completely unambiguous to everyone.

Napierian log also seems to have two definitions, one being `ln(x)`, the other `-10^7 ln(x / 10^7)`

Silmathoron:
> My TI-83 and Google use base 10. :)

I strongly object against using google or TI as reference! xD

I did not know about this weird definition of the logarithm...

Anyway, I still think we should stick to the standards of the communities that are closest to us, but given that a priori nobody will use the logarithm in a model, it's also not a big issue...
I'll let you see with @jougs whether he has an opinion on this ;)

jougs:
@Silmathoron: I was hoping that I don't have to have an opinion on this and you two can disentangle this on your own ;-)

That said, I lean slightly towards using the definitions as used in Python/C++ with the same justification as put forward by @Silmathoron.

pynestml/symbols/predefined_functions.py
```diff
@@ -32,7 +32,8 @@ class PredefinedFunctions(object):
         PRINT                 The callee name of the print function.
         PRINTLN               The callee name of the println function.
         EXP                   The callee name of the exponent function.
-        LOG                   The callee name of the logarithm function.
+        LN                    The callee name of the natural logarithm function, i.e. the logarithm function of base :math:`e`.
+        LOG                   The callee name of the logarithm function of base 10.
```
Silmathoron:
```suggestion
        LOG10                 The callee name of the logarithm function of base 10.
```

PR #521, opened 2019-10-29
Point out difference between "ODE functions" and regular functions in documentation

clinssen:
I'm not really 100% happy about this state of affairs, but at least this PR brings some documentation about the fact that we have two different types of function indicated by the "function" keyword in NESTML. Any language design comments are welcome.

### DIFF
--- a/doc/nestml_language.md
+++ b/doc/nestml_language.md
@@ -181,7 +181,17 @@ Examples for valid assignments for a numeric variable `n` are
 
 #### Functions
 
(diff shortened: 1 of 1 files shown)


clinssen:
Additional notes: "ODE functions" can only be defined within the equations block. Investigate the possibility of removing the "function" keyword here entirely, then a "function" unambiguously refers to a callable function, and the concept of an "ODE function" would disappear entirely.

clinssen:
ODE functions have been renamed to inline expressions in commit https://github.com/nest/nestml/pull/515/commits/e7fbaa5a6a53968374bfe6f9f104910390cd0368 of #515, superseding this PR. Please continue all further discussion about ODE functions/inline expression on that PR's thread.

PR #525, opened 2019-11-22
Syntax highlighting for pandoc

terhorstd:
These files can be used for syntax highlighting in KDE based environments like Kate and for `pandoc` markup conversion e.g. towards reveal.js presentations.


Silmathoron:
Nice! Is there a way to add some instructions to sphinx to use it for syntax highlighting in the docs?
Could you also add a note in the Readme to tell users where they need to put the xml file to have it the their KDE apps config?

terhorstd:
I don't have KDE on any of my machines, so I could only paste internet search results, sorry.
Sphinx uses Pygments, which doesn't seem that difficult. Should go to a different PR though.

PR #526, opened 2019-11-29
Require type/physical unit specification on nonspiking input ports

### LINKED ISSUE #449: Misleading implementation of `I_stim`
If I understand the model files correctly, the input current from e.g. `step_current_generator` in NEST is currently passed to the `I_stim` variable

    # Input current injected by CurrentEvent.
    # This variable is used to transport the current applied into the
    # _dynamics function computing the derivative of the state vector.
    I_stim pA = 0pA

However, I believed that the injected current would actually be the ``curents`` variable defined in the ``input`` block

    input:
      spikesInh nS  <- inhibitory spike
      spikesExc nS  <- excitatory spike
      currents <- current
    end

Yet, I also noticed that this has no unit defined, which seems surprising...

So, in a nutshell, which is which and how do they differ?

### LINKED ISSUE #450: "current" input is missing physical units
e.g.:

    input:
        AMPA nS  <- spike
        currents <- current
    end

clinssen:
Make "current" type input ports require a physical unit. Usually, this unit is pA, but e.g. nA and real are now also made possible in a generic manner.

A context condition check is included, that will log an error if no type is specified. No integration tests are included. Under the hood, a conversion to target-specific "neuroscience units" is made (for NEST, pA) when printing variables during code generation. This makes use of the existing infrastructure for converting to target platform specific units (see UnitConverter).

The above fixes #450.

Fixes #449: Current input ports on all models are now consistently named "I_e", while the constant-zero parameter by the same name is removed.  

Additionally, it turned out that "ODE functions" (see #521) are not checked for consistency of units. This PR also fixes that.


models/aeif_cond_alpha.nestml
```diff
@@ -99,7 +93,7 @@ neuron aeif_cond_alpha:
   input:
     spikesInh nS  <- inhibitory spike
     spikesExc nS  <- excitatory spike
-    currents <- current
+    I_e pA <- current
```
jougs: See above.

models/aeif_cond_alpha.nestml
```diff
@@ -74,11 +73,6 @@ neuron aeif_cond_alpha:
     tau_syn_ex ms = 0.2 ms    # Synaptic Time Constant Excitatory Synapse
     E_in mV = -85.0 mV        # Inhibitory reversal Potential
     tau_syn_in ms = 2.0 ms    # Synaptic Time Constant for Inhibitory Synapse
-
-    # Input current injected by CurrentEvent.
-    # This variable is used to transport the current applied into the
-    # _dynamics function computing the derivative of the state vector.
-    I_stim pA = 0 pA
```
jougs: I think this removal is correct and in line with what I said about the other current inputs. The name and comment of this one suggests that this was actually meant to be what *currents* was made to do? Maybe *I_stim* is a good name for the current buffer in the `input` block?

models/aeif_cond_alpha.nestml
```diff
@@ -177,12 +171,12 @@ neuron aeif_cond_alpha_implicit:
...
 
     # Add aliases to simplify the equation definition of V_m
-    function exp_arg real = (V_bounded-V_th)/Delta_T
+    function exp_arg real = (V_bounded-V_th)/Delta_T / (mV/ms)
     function I_spike pA = g_L*Delta_T*exp(exp_arg)
     function I_syn_exc pA =   convolve(g_ex, spikesExc) * ( V_bounded - E_ex )
     function I_syn_inh pA =   convolve(g_in, spikesInh) * ( V_bounded - E_in )
 
-    V_m' = ( -g_L*( V_bounded - E_L ) + I_spike - I_syn_exc - I_syn_inh - w + I_e + currents ) / C_m
+    V_m' = ( -g_L*( V_bounded - E_L ) + I_spike - I_syn_exc - I_syn_inh - w + I_e ) / C_m
```
jougs: See above.

models/aeif_cond_alpha.nestml
```diff
@@ -218,7 +211,7 @@ neuron aeif_cond_alpha_implicit:
   input:
     spikesInh nS  <- inhibitory spike
     spikesExc nS  <- excitatory spike
-    currents <- current
+    I_e pA <- current
```
jougs: See above.

models/aeif_cond_alpha.nestml
```diff
@@ -193,7 +187,6 @@ neuron aeif_cond_alpha_implicit:
     V_reset mV = -60.0 mV     # Reset Potential
     g_L nS = 30.0 nS          # Leak Conductance
     E_L mV = -70.6 mV         # Leak reversal Potential (aka resting potential)
-    I_e pA = 0 pA             # Constant Current
```
jougs: See above.

clinssen:
Thanks for the comments! If `I_e` has to stay by that name and refer to the parameter-based current, then I would suggest `I_inj` for the device-based current injection. `I_stim` implies that something is being stimulated, whereas (for e.g. a negative injected current) the purpose might be to inhibit the cell. Alternative suggestions are of course welcome.

models/aeif_cond_alpha.nestml
```diff
@@ -59,7 +59,6 @@ neuron aeif_cond_alpha:
...
```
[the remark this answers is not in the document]
Silmathoron: we can make it ``external_current`` or ``current_input``?

models/aeif_cond_alpha.nestml
Silmathoron: proposing here again ``external_current`` or ``current_input``

pynestml/cocos/co_co_only_spike_buffer_data_types.py
```diff
@@ -25,15 +25,15 @@
 
 class CoCoOnlySpikeBufferDataTypes(CoCo):
     """
-    This coco ensures that all spike buffers have a data type and all current buffers no data type stated.
+    This coco ensures that all spike and current buffers have a data type stated.
     Allowed:
         input:
-            spikeIn integer <- inhibitory spike
+            spikeIn pA <- inhibitory spike
```
Silmathoron: I would leave the previous example and add ``current pA <- current`` (or whatever we decide to name it), making it 2 lines inside of the input block. I would also leave spike as integer

pynestml/meta_model/ast_input_line.py
```diff
@@ -48,14 +48,14 @@ class ASTInputLine(ASTNode):
         signal_type = None
     """
 
-    def __init__(self, name=None, size_parameter=None, data_type=None, input_types=list(), signal_type=None,
+    def __init__(self, name=None, size_parameter=None, data_type=None, input_types=None, signal_type=None,
```
Silmathoron: is it just me or are we missing an initialization of ``input_types`` to ``[]`` at the beginning of the function now?

clinssen: This is a common Python pitfall/antipattern; see e.g. https://nikos7am.com/posts/mutable-default-arguments/

clinssen: Misread your comment; you are right! Thanks for catching this.

clinssen:
For consistency with existing NEST terminology, we went with `I_stim` for the external input current port.

Silmathoron:
OK but why do we have to decide on it and cannot let it be user-defined? I mean is there any technical reason?
Cannot re-review right know but will try to do it this week

clinssen:
There's no technical reason why they should be there, but because most NEST neuron models have an `I_e` parameter, users may expect a parameter with this name and behaviour to exist in the generated model. It is just a convention, of course you could name these in any way you want when writing a model. (We already had long discussions about naming :) )

Silmathoron:
Haha, sorry, unclear comment!
I know why we have ``I_e``.
What I was asking is whether there is any technical reason why the input buffer needs to have a specific name and cannot be left to the user to define (or not).
What I mean is that, though we need to define one name to use for already existing models, shouldn't the user be free to define this buffer with whatever name they prefer?
But now that I slept a little, I guess that you probably also agree on that and just wanted to define a name for the buffer in the already "standard" NEST models that have their .nestml files already in the repo, so you can probably ignore my ramble ^^

clinssen:
Indeed, it's partly about defining the convention, but also in part because this allows us to write integration tests that automatically check the neuronal behaviour for injected current (e.g. by comparing the effects of injecting into `I_e` versus into `I_stim`).

models/aeif_cond_alpha.nestml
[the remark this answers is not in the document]
jougs: I think `I_stim` is a good one.

tests/invalid/CoCoCurrentBufferTypeSpecified.nestml
```diff
@@ -31,8 +31,8 @@ neuron CoCoCurrentBufferTypeSpecified:
     end
 
     input:
-        currents <- inhibitory current # current buffer types can not be specified
+        currents pA <- inhibitory current # current buffer types can not be specified
```
jougs: Shouldn't this comment also be adapted now they *have to* be specified?

tests/valid/CoCoCurrentBufferTypeSpecified.nestml
```diff
@@ -31,8 +31,8 @@ neuron CoCoCurrentBufferTypeSpecified:
     end
 
     input:
-        currents <- current # current buffer not specified with any type or keywords, thus correct
+        currents pA <- current # current buffer not specified with any keywords, thus correct
```
jougs: Adapt comment.

tests/valid/CoCoSumNotCorrectlyParametrized.nestml
```diff
@@ -28,21 +28,21 @@
...
     initial_values:
-        V_m mV = 10mV
+        G real = 1.
     end
 
     equations:
-        shape V_m' = 20mV
-        function testB pA = convolve(V_m, spikeExc) # convolve aka. sum is now correctly parametrized
+        shape G' = -G
+        function testB pA = convolve(G, spikeExc) # convolve aka. sum is now correctly parametrized
```
jougs: Remove "aka. sum", or do we need it?

clinssen:
To eliminate confusion, I have renamed "input [port] type" to "input [port] qualifier", because it does not refer to the data type of the port, but to the "excitatory" or "inhibitory" keywords. I have also renamed "input line", which refers to an entry in the inputs block, to "input port".

The latest commits should resolve all of your comments.

jougs:
@Silmathoron: happy new year! Can you please indicate if you are happy with @clinssen's changes? Thanks!

models/aeif_cond_alpha.nestml
```diff
@@ -43,7 +43,7 @@ neuron aeif_cond_alpha:
     shape g_ex = (e/tau_syn_ex) * t * exp(-t/tau_syn_ex)
 
     # Add functions to simplify the equation definition of V_m
-    function exp_arg real = (V_bounded-V_th)/Delta_T
+    function exp_arg real = (V_bounded-V_th)/Delta_T / (mV/ms)
```
Silmathoron: there's an issue with the units here: the previous version was correct
```suggestion
    function exp_arg real = (V_bounded-V_th)/Delta_T
```

models/aeif_cond_alpha.nestml
```diff
@@ -177,12 +174,12 @@ neuron aeif_cond_alpha_implicit:
     shape g_ex'' = (-2/tau_syn_ex) * g_ex'-(1/tau_syn_ex**2) * g_ex
 
     # Add aliases to simplify the equation definition of V_m
-    function exp_arg real = (V_bounded-V_th)/Delta_T
+    function exp_arg real = (V_bounded-V_th)/Delta_T / (mV/ms)
```
Silmathoron: see above
```suggestion
    function exp_arg real = (V_bounded-V_th)/Delta_T
```

models/hh_cond_exp_traub.nestml
```diff
@@ -64,7 +64,7 @@ neuron hh_cond_exp_traub:
     function I_syn_exc pA = convolve(g_ex, spikeExc) * ( V_m - E_ex )
     function I_syn_inh pA = convolve(g_in, spikeInh) * ( V_m - E_in )
 
-    V_m' =( -I_Na - I_K - I_L - I_syn_exc - I_syn_inh + currents + I_e ) / C_m
+    V_m' =( -I_Na - I_K - I_L - I_syn_exc - I_syn_inh + I_e + I_stim ) / C_m
```
Silmathoron: Just for esthetics
```suggestion
    V_m' = ( -I_Na - I_K - I_L - I_syn_exc - I_syn_inh + I_e + I_stim ) / C_m
```

models/hh_cond_exp_traub.nestml
```diff
@@ -195,7 +196,7 @@ neuron hh_cond_exp_traub_implicit:
...
```
Silmathoron: again
```suggestion
    V_m' = ( -I_Na - I_K - I_L - I_syn_exc - I_syn_inh + I_e + I_stim ) / C_m
```

models/hh_psc_alpha.nestml
```diff
@@ -217,7 +219,7 @@ neuron hh_psc_alpha_implicit:
     function I_Na  pA = g_Na * Act_m * Act_m * Act_m * Inact_h * ( V_m - E_Na )
     function I_K   pA  = g_K * Act_n * Act_n * Act_n * Act_n * ( V_m - E_K )
     function I_L   pA = g_L * ( V_m - E_L )
-    V_m' =( -( I_Na + I_K + I_L ) + currents + I_e + I_syn_inh + I_syn_exc ) / C_m
+    V_m' =( -( I_Na + I_K + I_L ) + I_e + I_stim + I_syn_inh + I_syn_exc ) / C_m
```
Silmathoron: and again
```suggestion
    V_m' = ( -( I_Na + I_K + I_L ) + I_e + I_stim + I_syn_inh + I_syn_exc ) / C_m
```

models/iaf_psc_exp.nestml
```diff
@@ -86,10 +86,10 @@ neuron iaf_psc_exp:
...
 
   equations:
-    shape I_shape_in = exp(-1/tau_syn_in*t)
-    shape I_shape_ex = exp(-1/tau_syn_ex*t)
-    function  I_syn mV = (convolve(I_shape_in, in_spikes) + convolve(I_shape_ex, ex_spikes) + I_e + currents)
-    V_abs' = -V_abs/tau_m + (I_syn/C_m) *nS
+    shape I_shape_in = exp(-t / tau_syn_in)
+    shape I_shape_ex = exp(-t / tau_syn_ex)
+    function I_syn pA = convolve(I_shape_in, in_spikes) + convolve(I_shape_ex, ex_spikes)
+    V_abs' = -V_abs / tau_m + (I_syn + I_e + I_syn) / C_m
```
Silmathoron: double `I_syn`
```suggestion
    V_abs' = -V_abs / tau_m + (I_syn + I_e + I_stim) / C_m
```

clinssen:
Thanks for the review! I have merged all your comments.

PR #531, opened 2020-01-14
Add support for drawing random numbers from uniform and normal distributions

clinssen:
Adds a `randomNorm(mean, std)` function to NESTML, along with a unit test and new neuron model that uses randomNorm to generate Ornstein-Uhlenbeck noise.

With props to @TeEsTeBe! Quoting: "I added the new neuron model with integrated Ornstein-Uhlenbeck noise and a new ion channel. I also added an example script into tests/nest_tests, which reproduces some figures from Destexhe et al. 2001 and prints out the mean and variances of the conductances and compares them to the target values."


clinssen:
@jougs: By request, I am also adding a randomUniform function. Would you prefer this as a function without parameters, mapping to rng->drand() which returns a value in the interval [0, 1), or shall we give it "offset" and "scale" parameters, to make the interval arbitrary? (Of course this simply boils down to multiplying the obtained [0, 1)-sample with the scale parameter and adding the offset parameter, which conceivably the end user could also do explicitly in the NESTML code.)

Silmathoron:
``random_normal`` would probably make more sense to users, wouldn't it?

jougs:
@Silmathoron: fine with me!

clinssen:
Sorry for the slight feature creep here. Test discovery was not properly set up for the NEST Simulator integration tests, i.e. those in `tests/nest_tests` that run an actual simulation in NEST. Initial reason for this was that they need to run in independent processes, otherwise `nest.Install(...)` will be called more than once in the same process, leading to "module already loaded" errors. This was solved by a pytest extension called pytest-xdist.

PR #541, opened 2020-03-17
Update antlr version to v4.8

### LINKED ISSUE #540: ANTLR warning
Executing NESTML, I get lots of warning from ``antlr4-python3-runtime`` telling me:

    ANTLR runtime and generated code versions disagree: 4.8!=4.7.1

My guess is that this comes from stored generated files in the NESTML repo (e.g. [here](https://github.com/nest/nestml/blob/b100560aa1b4cfd32fec06e32bcaf6bf7e0eb2ec/pynestml/generated/PyNestMLParserVisitor.py#L1)).
Would it be possible to remove those generated files (regenerate during first call to NESTML on the user's computer) or document a way to remove the warning?
This makes the NESTML output completely unreadable ^^"

clinssen:
Fixes #540

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,9 @@ before_install:
 
   # install antrl4
   - sudo apt install openjdk-8-jre
-  - wget http://www.antlr.org/download/antlr-4.7.1-complete.jar
+  - wget http://www.antlr.org/download/antlr-4.8-complete.jar
   - echo \#\!/bin/bash > antlr4
-  - echo java -cp \"`pwd`/antlr-4.7.1-complete.jar:$CLASSPATH\" org.antlr.v4.Tool \"\$@\" >> antlr4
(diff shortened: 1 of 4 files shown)


Silmathoron:
This is a very temportary way of fixing the issue... could we actually fix the underlying problem?
Couldn't the files be generated locally upon install?

clinssen:
They are indeed regenerated in the Travis CI suite in order to ensure correctness, but requiring users to regenerate them locally upon install would imply a dependency on antlr, which in turn means a dependency on the Java runtime, which seems a bit heavyweight.

Silmathoron:
Damn... otherwise could we [turn the warning off](https://stackoverflow.com/questions/22559158/how-to-turn-off-the-mismatch-warnings-in-antlr4)?

jougs:
@Silmathoron: In a VC with @clinssen just now we decided that we're keeping the warnings, but merge this PR as it is. Would you be fine with this?

Silmathoron:
I am unfortunately completely drowning under the amount of things I need to do right now, so if you think this is the best thing to do, I'll trust that you checked possible alternatives and decided that this was was the least bad ;)

PR #548, opened 2020-05-26
Temporarily downgrade to an older version of sympy to fix CI until #515 is merged

clinssen:


clinssen:
Merging this now so we can move forward with other PRs. A commit undoing this change has been added to #515.

PR #562, opened 2020-08-26
Make all blocks optional

clinssen:
Successor to #558.

Allows no update block in the model. Altogether, this means that all blocks are now optional. This allowed a cleaning up of a lot of NESTML files used in the unit tests.


clinssen:
Unfortunately, I've had to integrate this as part of #515.

PR #582, opened 2020-09-30
Co-generation of neuron and synapse models

clinssen:
This PR supersedes #459, and adds the infrastructure to NESTML to allow code to be generated for neuron and synapse models in tandem, allowing the generation of efficient code for the given pairs. This is a crucial enabling factor for modeling synapses in NESTML.

## Notes for reviewers

- The rendered documentation can be found here: https://nestml-synapse.readthedocs.io/en/latest/synapses_in_nestml.html

## Todo

- [ ] Make spike_updates a mapping from input port to expressions
- [ ] Composability: if a neuron is connected to by more than one type of synapse, it needs to support the requirements of both (e.g. Clopath as well as Urbancik-Senn archiving)
- [ ] Make generated synapses thread-safe: issue inserting by string in recordables map. Use nest::names instead. Add custom named variables to nest::names during model registration (in nestml generated models module).
- [ ] Refactor "preReceive" and "postReceive" to "onReceive(input_port_name)"
- [ ] Error handling for synapse models and synapses: ensure that no parameters are changed unless it is certain that all parameters values are legal. See https://github.com/nest/nest-simulator/issues/249

## Discussion points

- Rather than having separate onPre and onPost blocks, and making "postsynaptic" ports by the ``post`` keyword, consider making this more generic using port names, e.g. onReceive(post_name).
- If the synapse's ``deliver_spike()`` is renamed to be equal to the neuron's ``send_spike()``, in principle we could eliminate the ``neuron`` and ``synapse`` distinction as the top-level NESTML element as ``node``. Which code is generated (neuron or synapse code) would be specified in the JSON code generator options (which is currently already used to indicate co-generation dyads).
- 3rd factor plasticity: distinction between e.g. post partner membrane potential and volume_transmitter in terms of infrastructure required. Ideally, from the NESTML model perspective, these would both be supplied via an extra, named input port, e.g. 
  ```
  input:
    ...
    post_V_m mV <- continuous
  end
  ```
  Again, we would use the JSON code generator options to indicate the connection with the postsynaptic membrane potential variable (which may or may not be named "V_m" in the model).
- Physical unit of the spiking input port of a synapse: should be unitless, otherwise we would need a different synapse model for current vs. conductance-based models.

## Third-factor plasticity rules

Third-factor plasticity is not currently supported by this PR. However, a design proposal can be found (and discussed!) at #646.


clinssen:
Superseded by #655.

PR #585, opened 2020-10-02
Add Ornstein-Uhlenbeck noise tutorial

clinssen:


TeEsTeBe:
Hi,

I have some questions regarding the equations in the tutorial notebook. 
1. In contrast to equation 2.47 from the Gillespie paper, there is no tau/2 term under the square root in the discrete update equation in the notebook.
2. I assume the equation for the variance is based on equation 2.49b from the paper. However, there seems to be an additional h in the notebook equation (sigma^2 * tau / 2**h**).

Could you explain to me why this is the case?

clinssen:
1. This factor is absorbed into the constant σ. Compare eq. 2.24 and the first equation in the notebook:
 
   \sqrt{c} = σ\sqrt{2/τ}

2. You're absolutely right, this expression is wrong. I have replaced it with the much more obvious "Var(X) = sigma^2", which was validated for numerical correctness.

TeEsTeBe:
Thank you for the clarification. Now everything looks good to me. Nice tutorial!

TeEsTeBe (approved):

PR #592, opened 2020-10-27
Add explanation on symbol resolution priority

### LINKED ISSUE #481: Detect incorrect variable name

clinssen:
Fixes #481.


Silmathoron:
Sounds clear to me.
I would maybe just make the

> if that fails

more explicit, adding a word about what "failing" means.

> if that fails (because resolving the expression using the variable's value does not match the expected dimension/unit)

clinssen:
Thanks for pointing that out, I updated the text and added an example to hopefully make this behaviour more clear.

Silmathoron (approved):

clinssen:
Cheers!

PR #593, opened 2020-11-16
Generate code for state variables not used in any equations

clinssen:
This PR fixes an issue with state variables that were declared in the model, but not used in the equations block. This means that they are not processed by ODE-toolbox, which currently results in a missing declaration in the corresponding C++ header file. This PR introduces a separate data structure in the code generator, containing just these variables, so that their declaration is properly generated, next to those variables that do go through ODE-toolbox.


jougs (approved):
LGTM!

PR #597, opened 2020-12-06
Replace pytest-xdist with a shell loop

clinssen:
Integration tests have to run in separate processes, because each test individually imports nest and installs one or more extension modules with the same name.

Up until now, we used a third-party extension to pytest, namely pytest-xdist, but this prevents us from inspecting debug output (it eats stdout). This PR replaces pytest-xdist with a simple shell loop, reducing the dependencies and allowing inspection of the complete log.


clinssen:
Note: this one is expected to fail until #593 is merged.

jougs (approved):
LGTM given that Travis is thinking the same.

PR #600, opened 2021-01-05
Fix print, println functions

pnbabu:
Support print and println functions in NESTML


clinssen:
For println, please set the argument type correctly in line 151 of pynestml/symbols/predefined_functions.py.

I'm still getting some runtime errors when printing more than two variables/parameters, e.g.
```nestml
parameters:
  bar ms = 42 ms
end

update:
  print("bar is: {bar}, bar = {bar}, bar = {bar}")
end
```

Do you think we should also print the physical unit, so it would print "bar is 42 ms" rather than "bar is 42"? This is a little tricky because physical units are known at the NESTML (Python) level, but once we generate C++ code, the parameters and variables are normalised, so we would have to keep track of the multipliers (e.g. all currents are assumed to be in pA in NEST).

pnbabu:
Thanks for pointing the error case. I had to split the print statement for the first occurrence of the variable instead of all of them. Fixed it in the recent commit.

It would be good to print the values along with the physical unit. I will see how we can accommodate this in the code.

pnbabu:
The pending task is to add integration tests for nest with print statements having variables with nest-incompatible units and to check if the unit conversion happens and the right units are displayed in the cout statements.

clinssen (commented):
I think the SI standard is to have one space between the number and the unit, could you double-check this and maybe even add a reference to the API docstrings?

I ran into a bug with the units, I'm not sure it's related but if you could please apply this diff:

```
--- a/pynestml/codegeneration/nest_reference_converter.py
+++ b/pynestml/codegeneration/nest_reference_converter.py
@@ -245,7 +245,7 @@ class NESTReferenceConverter(IReferenceConverter):
         variable_name = NestNamesConverter.convert_to_cpp_name(variable.get_complete_name())
         symbol = variable.get_scope().resolve_to_symbol(variable_name, SymbolKind.VARIABLE)
         if isinstance(symbol.get_type_symbol(), UnitTypeSymbol):
-            return symbol.get_type_symbol().unit.unit.name
+            return symbol.get_type_symbol().unit.unit.to_string()
 
         return ''
```

jougs:
> I think the SI standard is to have one space between the number and the unit, could you double-check this and maybe even add a reference to the API docstrings?

I am not aware we ever formally discussed this previously for NEST. However, as @clinssen correctly points out, the SI standard is to have a space between the numerical value and unit symbol (see [#15 here](https://physics.nist.gov/cuu/Units/checklist.html)) and `rgrep "\" ms" *` produces hits, which means that we're already following that recommendation at least to a certain degree.

@sarakonradi, could you please write this policy down in the appropriate documentation guideline?

pnbabu:
> I think the SI standard is to have one space between the number and the unit, could you double-check this and maybe even add a reference to the API docstrings?
> 
> I ran into a bug with the units, I'm not sure it's related but if you could please apply this diff:
> 
> ```
> --- a/pynestml/codegeneration/nest_reference_converter.py
> +++ b/pynestml/codegeneration/nest_reference_converter.py
> @@ -245,7 +245,7 @@ class NESTReferenceConverter(IReferenceConverter):
>          variable_name = NestNamesConverter.convert_to_cpp_name(variable.get_complete_name())
>          symbol = variable.get_scope().resolve_to_symbol(variable_name, SymbolKind.VARIABLE)
>          if isinstance(symbol.get_type_symbol(), UnitTypeSymbol):
> -            return symbol.get_type_symbol().unit.unit.name
> +            return symbol.get_type_symbol().unit.unit.to_string()
>  
>          return ''
> ```

What is the bug that you ran into?

pnbabu:
> I think the SI standard is to have one space between the number and the unit, could you double-check this and maybe even add a reference to the API docstrings?
> 

Thanks for pointing this out. I added a space between the variable and the physical unit and also updated the API docstrings.

> I ran into a bug with the units, I'm not sure it's related but if you could please apply this diff:
> 
> ```
> --- a/pynestml/codegeneration/nest_reference_converter.py
> +++ b/pynestml/codegeneration/nest_reference_converter.py
> @@ -245,7 +245,7 @@ class NESTReferenceConverter(IReferenceConverter):
>          variable_name = NestNamesConverter.convert_to_cpp_name(variable.get_complete_name())
>          symbol = variable.get_scope().resolve_to_symbol(variable_name, SymbolKind.VARIABLE)
>          if isinstance(symbol.get_type_symbol(), UnitTypeSymbol):
> -            return symbol.get_type_symbol().unit.unit.name
> +            return symbol.get_type_symbol().unit.unit.to_string()
>  
>          return ''
> ```

I checked that the variable `name` in `unit` is of type `str`. Could you tell me how this change fixes the bug you are seeing?

clinssen:
> What is the bug that you ran into?

Try replacing "V" in "V_m V = -50 mV" into "MA*Ohm" to trigger it.

pnbabu:
> > What is the bug that you ran into?
> 
> Try replacing "V" in "V_m V = -50 mV" into "MA*Ohm" to trigger it.

Yes, you are right. Updated the code and test. Thank you.

clinssen (approved):
Excellent, thanks a lot!

jougs (approved):
Very nice! I'm especially impressed by the tests :-)

clinssen:
One more thing: could you add support for printing the predefined variable t (for time)? Because it is not defined explicitly in any NESTML model, we can assume that its corresponding unit is the "neuroscience unit" of the target platform.

clinssen:
Perfect, cheers!

Just as a note for future reference: the time variable issue has been spun off to #609.

pynestml/codegeneration/nest_reference_converter.py
```diff
@@ -228,6 +236,65 @@ def convert_name_reference(self, variable, prefix=''):
...
+        stmt = stmt[stmt.index('"') + 1: stmt.rindex('"')]  # Remove the double quotes from the string
+        scope = function_call.get_scope()
+        return self.__convert_print_statement_str(stmt, scope)
+
+    def __convert_print_statement_str(self, stmt, scope):
+        """
+        Converts the string argument of the print or println function to NEST processable format
+        Variables are resolved to NEST processable format and printed with physical units as mentioned in model, separated by a space
+
+.. code-block:: nestml
```
clinssen, line 270:
```suggestion
        .. code-block:: nestml
```

clinssen: To make sure that this renders nicely as HTML, we have to be really pedantic about the whitespace: the code itself is indented three spaces beyond the start of "..", so e.g.:
```
.. code-block:: nestml
   print("Hello world")
```

Converted NEST:
```
.. code-block:: C++
   std::cout << "Hello world";
```

tests/nest_tests/print_statement_test.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+import os
+import subprocess
+import unittest
+
+
+class PrintStatementTest(unittest.TestCase):
+
```
clinssen, line 27: Could you add one-line docstring here describing what is tested for?

clinssen (changes requested):

clinssen (approved):

Issue #599, opened 2020-12-30
Error with astropy while installing NESTML


pnbabu:
The version of `astropy` in the nestml requirements.txt file is set to 2.0.3. This version is resulting in some errors in the Mac OS while installing NESTML.

```
5 warnings generated.
warning: unknown warning option '-Wno-format-overflow'; did you mean '-Wno-shift-overflow'? [-Wunknown-warning-option]
cextern/cfitsio/group.c:5476:3: error: implicit declaration of function 'getcwd' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  getcwd(buff,FLEN_FILENAME);
  ^
1 warning and 1 error generated.
error: Setup script exited with error: command 'gcc' failed with exit status 1
```

However, removing the version from requirements.txt, astropy is updated to the latest version which is 4.2, and the errors don't seem to occur. The unit tests pass with this version and the NESTML works fine. Thus, the requirements file should be updated with the appropriate new version of astropy.

jougs:
@pnbabu: thanks for your bug report. Could you please file a pull request fixing the `requirements.txt`? Thanks!

PR #601, opened 2021-01-12
Error with astropy while installing NESTML #599

pnbabu:
Remove the version of astropy to install the latest version


jougs (approved):
Fine by me, but I leave the final decision to @clinssen. Thanks for finding and fixing!

This fixes #599.

clinssen (approved):
Cheers!

PR #607, opened 2021-01-21
Use conda to generate documentation on ReadTheDocs

clinssen:
Porting changes in https://github.com/nest/nest-simulator/pull/1843 to NESTML ReadTheDocs environment.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,11 @@
-type: sphinx
-configuration: doc/sphinx-apidoc/conf.py
-python:
-    version: 3
-    requirements_file: requirements.txt
-formats:
-    - none
+version: 2
+
+sphinx:
+  builder: html
+  configuration: doc/sphinx-apidoc/conf.py
+
+conda:
+  environment: doc/environment.yml
+
+build:
+  image: latest
--- a/doc/conda_environment.yml
+++ b/doc/conda_environment.yml
@@ -1,0 +1,7 @@
+name: nestml-doc
+dependencies:
+  - python>=3.6
+  - pip
+  - pip:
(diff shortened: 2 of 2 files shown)


.readthedocs.yml
```diff
@@ -1,7 +1,11 @@
...
-formats:
-    - none
+version: 2
+
+sphinx:
+  builder: html
+  configuration: doc/sphinx-apidoc/conf.py
+
+conda:
+  environment: doc/environment.yml
```
pnbabu, line 8: Shouldn't this be renamed to conda_environment.yml?

jougs: I think it shouldn't. Using the default name (which to the best of my knowledge is just `environment.yml`) also makes sure that you don't have to specify anything special when creating the environment:
```bash
cd doc
conda env create
```

jougs (changes requested):
Sorry for making you do this :-/

I hope my suggestions help.

doc/conda_environment.yml
```diff
@@ -0,0 +1,7 @@
+name: nestml-doc
+dependencies:
+  - python>=3.6
+  - pip
+  - pip:
+    - -r file:../requirements.txt
```
jougs, line 6: I think this should be
```suggestion
    - -r file:requirements.txt
```
and that file (i.e. probably a new one under `doc`) should contain the sphinx requirements. Right now, this will install the dependencies for _running_ NESTML, wouldn't it?

clinssen:
The current best practice for the document building environment is to move away from conda and towards requirements.txt (see https://github.com/nest/nest-simulator/pull/1965), which is what is already implemented currently. Therefore, the PR can be closed without merging.

PR #610, opened 2021-01-27
Add 'current' type input ports as zero-value parameters for ODE-toolbox analysis

clinssen:
The dynamics of the iaf_psc_exp neuron is defined as:
```
inline I_syn pA = convolve(I_kernel_in, in_spikes) + convolve(I_kernel_ex, ex_spikes) + I_e + I_stim
V_abs' = -V_abs / tau_m + I_syn / C_m
```
with
```
parameters:
    [...]
    I_e pA = 0 pA      # constant external input current
end
```
and 
```
input:
    [...]
    I_stim pA <- current
end
```
The issue is that ODE-toolbox suggests a numeric solver for V_m, rather than an analytic solver. This is because the input port ``I_stim`` is not passed to ODE-toolbox, so it doesn't know about and assumes the most general case: that it is not constant within each timestep. Hence, it recommends a numeric solver.

This PR adds all continuous-valued ("current" type) input ports to the ODE-toolbox input as a parameter with value 0, solving the issue.

@jougs: could you verify that "current" type input ports are indeed expected to have a constant value within a single timestep? (I am 99% sure of this but want to double-check that I'm not missing certain design considerations.)

### DIFF
--- a/pynestml/codegeneration/nest_codegenerator.py
+++ b/pynestml/codegeneration/nest_codegenerator.py
@@ -807,6 +807,12 @@ class NESTCodeGenerator(CodeGenerator):
                     odetoolbox_indict["parameters"][var.get_complete_name(
                     )] = gsl_printer.print_expression(decl.get_expression())
 
+        # add "current" type input ports to parameters with zero value, to prevent unknown terms from preventing an analytic solution
+        if neuron.get_input_blocks() is not None:
+            for input_port in neuron.get_input_blocks().get_input_ports():
+                if input_port.is_current():
+                    odetoolbox_indict["parameters"][input_port.get_name()] = "0"
+
         return odetoolbox_indict
 
     def make_inline_expressions_self_contained(self, inline_expressions: List[ASTInlineExpression]) -> List[ASTInlineExpression]:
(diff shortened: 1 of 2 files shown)


jougs:
@heplesser, @diesmann: I agree with @clinssen, but maybe you also have something to say here? Many thanks!

pynestml/codegeneration/nest_codegenerator.py
```diff
@@ -807,6 +807,12 @@ def transform_ode_and_kernels_to_json(self, neuron: ASTNeuron, parameters_block,
                     odetoolbox_indict["parameters"][var.get_complete_name(
                     )] = gsl_printer.print_expression(decl.get_expression())
 
+        # add "current" type input ports to parameters with zero value, to prevent unknown terms from preventing an analytic solution
```
heplesser, line 810: I don't understand the comment. What to you mean by "adding input ports to parameters"?

heplesser (commented):
@clinssen Your approach seems sensible to me, but just so I understand fully: Nothing will change on the NESTML level (still `I_stim <- pA current`, and there will still be a port), you just set the current value to zero before you pass the equations to the ODE solver? I assume there is no way to explicitly tell the solver that a certain quantity represents a stepwise constant quantity?

heplesser:
>  I assume there is no way to explicitly tell the solver that a certain quantity represents a stepwise constant quantity?

Pasting an answer from @diesmann:

Yes, there is.
You can add the equation \dot{I}=0 to your system of equations. This is
what we doin the EI models and was one of the main insights of the 99 paper.
This promises the solver that over the duration of the integration step
the value of I will not change (its derivative is zero).

clinssen:
Closing due to upstream improvements in ODE-toolbox, making this no longer necessary.

PR #612, opened 2021-01-31
Add JSON code generator options feature

clinssen:
This PR adds the feature to pass in a JSON file containing options for the code generator. This will be necessary for #582, but can also be used independently. For example, the user could inherit their NESTML neuron model from ArchivingNode (the default), but could also choose to inherit from StructuralPlasticityNode when the ArchivingNode functionality is not needed.

This PR also refactors the CodeGenerator classes a little bit: the inheritance is now more clean, and CodeGenerator instances are produced by a static factory method (*from_target_name()*) of the CodeGenerator base class.


clinssen:
To support the correct type annotations, Python on Travis CI had to be upgraded from 3.6 to 3.8.

pynestml/codegeneration/codegenerator.py
```diff
@@ -19,38 +19,56 @@
...
+
+    def generate_neuron_code(self, neuron) -> None:
+        """the base CodeGenerator class does not generate any code"""
+        pass
+
+    def set_options(self, options: Mapping[str, Any]):
+        if not "_default_options" in dir(self.__class__):
+            assert "Code generator class \"" + str(self.__class__) + "\" does not support setting options."
+        for k in options.keys():
+            assert k in self.__class__._default_options, "Option \"" + str(k) + "\" does not exist in code generator"
```
jougs, line 60:
```suggestion
            assert k in self.__class__._default_options, f'Option "{k}" does not exist in code generator.'
```

pynestml/codegeneration/codegenerator.py
```diff
@@ -60,24 +78,26 @@ def generate_neurons(self, neurons: List[ASTNode]):
...
+        return targets
 
-        :param neurons: a list of neurons.
-        :type neurons: List[ASTNode]
-        """
-        if self._target.upper() == "NEST":
+    @staticmethod
+    def from_target_name(target_name: str, options: Optional[Mapping[str, Any]]=None) -> CodeGenerator:
+        """Static factory method that returns a new instance of a child class of CodeGenerator"""
+        assert target_name.upper() in CodeGenerator.get_known_targets(), "Unknown target platform requested: \"" + str(target_name) + "\""
```
jougs, line 90:
```suggestion
        assert target_name.upper() in CodeGenerator.get_known_targets(), f'Unknown target platform requested: {target_name}.'
```

pynestml/codegeneration/codegenerator.py
```diff
@@ -60,24 +78,26 @@ def generate_neurons(self, neurons: List[ASTNode]):
...
-            _codeGenerator = NESTCodeGenerator()
-            _codeGenerator.generate_code(neurons)
-        elif self._target.upper() == "AUTODOC":
+            return NESTCodeGenerator(options)
+        elif target_name.upper() == "AUTODOC":
             from pynestml.codegeneration.autodoc_codegenerator import AutoDocCodeGenerator
-            _codeGenerator = AutoDocCodeGenerator()
-            _codeGenerator.generate_code(neurons)
-        else:
+            assert options is None or options == {}, "\"autodoc\" code generator does not support options"
```
jougs, line 96:
```suggestion
            assert options is None or options == {}, '"autodoc" code generator does not support options'
```

pynestml/frontend/pynestml_frontend.py
```diff
@@ -117,7 +126,13 @@ def install_nest(models_path, nest_path):
 
 
 def main():
-    """Returns the process exit code: 0 for success, > 0 for failure"""
+    """
+    Entry point for the command-line application.
+
+    Returns
+    -------
+    The process exit code: 0 for success, > 0 for failure
```
jougs, line 134: Do we have/need a list of the exit codes > 0 and what they mean? Would it be good to have them here?

requirements.txt
```diff
@@ -3,6 +3,6 @@ sympy >= 1.1.1
 antlr4-python3-runtime >= 4.7
 setuptools
 Jinja2 >= 2.10
-typing
+typing;python_version<"3.5"
```
jougs, line 6: Can there be a space between the semicolon and the next word? :crying_cat_face:

pynestml/frontend/frontend_configuration.py
```diff
@@ -185,14 +170,70 @@ def get_module_name(cls):
...
+    @classmethod
+    def set_codegen_opts(cls, codegen_opts):
+        """Set the code generator options dictionary"""
+        cls.codegen_opts = codegen_opts
+
+    @classmethod
+    def handle_codegen_opts_fn(cls, codegen_opts_fn):
+        """If a filename of a JSON file containing code generator options is passed on the command line, read it into a Python dictionary"""
+        if codegen_opts_fn and not os.path.isfile(codegen_opts_fn):
+            raise Exception('The specified code generator options file ("' + codegen_opts_fn + '") cannot be found')
```
jougs, line 195: yadda yadda yadda

jougs (changes requested):
Thanks! I'm mostly fine, but added some suggestions anyway ;-)

One thing I noticed throughout is that strings are delimited pretty inconsistently by single and double quotes which makes reading them a bit hard if they then also contain double or single quotes (some of which are even escaped using backslashes). In general, I think we should use single quotes to delimit all but multiline strings and use double quotes to delimit arguments within strings.

pynestml/frontend/pynestml_frontend.py
clinssen: Anything nonzero is typically interpreted as an error in POSIX. We only have 0 and 1, and I could just replace "> 0" with "1".

requirements.txt
clinssen, line 6:
```suggestion
typing; python_version<"3.5"
```

clinssen:
I have created  #617 for further discussion and addressing of how to format strings.

jougs (approved):
I'm fine with the newest changes and postponing the string-changes via #617. Thanks!

clinssen:
Thank you!

PR #615, opened 2021-02-05
Allow passing lists of input model files and directories

clinssen:
Currently, it is only possible to pass a single file or a single directory path as input to NESTML. This PR allows passing an enumeration of paths, so that you can single out particular models, for example:

```sh
nestml --input_path "models/iaf_psc_exp.nestml" "models/stdp.nestml"
```

This also works in the Python API:
```python
to_nest(input_path=["models/iaf_psc_exp.nestml", "models/stdp.nestml"], ...)
```

### DIFF
--- a/doc/running.rst
+++ b/doc/running.rst
@@ -18,7 +18,7 @@ where arguments are:
    * - ``-h`` or ``--help``
      - Print help message.
    * - ``--input_path``
-     - Path to the source file or directory containing the model.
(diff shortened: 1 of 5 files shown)


pnbabu:
@clinssen Looks good to me. Could you add unit tests for the change?

pnbabu (approved):
Thanks. Looks good to me.

clinssen:
Thank you!

PR #616, opened 2021-02-05
Fix static checker issues

clinssen:
Fix errors raised by mypy (a static code checker), especially those related to type annotations. Not meant to be exhaustive (but a step in the right direction).

### DIFF
--- a/pynestml/codegeneration/nest_codegenerator.py
+++ b/pynestml/codegeneration/nest_codegenerator.py
@@ -19,26 +19,22 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-from typing import Dict, List, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence
 
-import copy
-import json
 import datetime
 import os
 import re
 import sympy
-from typing import Any, Dict, List, Mapping, Optional, Union
 from jinja2 import Environment, FileSystemLoader, TemplateRuntimeError
 from odetoolbox import analysis
 
 import pynestml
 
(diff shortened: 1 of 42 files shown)


jougs (approved):
Fine by me. Although I only understand half of the changes ;-)

pynestml/codegeneration/nest_codegenerator.py
```diff
@@ -19,6 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
+from typing import Dict, List, Mapping, Optional, Sequence, Union
```
pnbabu, line 22: `typing` is imported twice. It could be merged into one.

pynestml/codegeneration/nest_codegenerator.py
```diff
@@ -54,6 +56,7 @@
 from pynestml.meta_model.ast_neuron import ASTNeuron
 from pynestml.meta_model.ast_node_factory import ASTNodeFactory
```
pnbabu, line 57: I see in my IDE that some of the imports are unused. Could you remove them as well?

pynestml/codegeneration/nest_codegenerator.py
```diff
@@ -850,8 +853,9 @@ def log_set_source_position(node):
 
         return inline_expressions
 
-    def replace_inline_expressions_through_defining_expressions(self, definitions, inline_expressions):
-        # type: (list(ASTOdeEquation), list(ASTInlineExpression)) -> list(ASTInlineExpression)
+    def replace_inline_expressions_through_defining_expressions(self,
+                                                                definitions: Sequence[ASTOdeEquation],
+                                                                inline_expressions: Sequence[ASTInlineExpression]) -> Sequence[ASTOdeEquation]:
```
pnbabu, line 858: Should the return type here be `Sequence[ASTInlineExpression]`?

clinssen: The returned object is the same as the first function parameter, so is correct. However, I updated the outdated docstring.

clinssen:
@pnbabu: thanks for the comments! Everything should be addressed in the latest commit. Could you take another look?

pnbabu (approved):
Looks fine to me.

clinssen:
Cheers!

PR #618, opened 2021-02-09
Fix parsing of newlines in grammar

clinssen:
Newlines shouldn't be on a separate channel if we want to use them in the grammar.

This fixes parsing
```
update:
   a = 42
   temp nS = 42 nS
end
```
as one assignment and one declaration.

Old parse result was (two assignments):
```
a = 42 temp
nS = 42 nS
```


clinssen:
Sorry, this needs more work to fix processing of comments.

clinssen:
Notes for review:

The main change is that the NEWLINE grammar rule does not cause newlines to go to channel 3 any longer. Consequently, newlines now have to be explicitly permitted in a few places in the grammar. Tests in Python code for whether tokens are on channel 3 have had to be replaced.

The test has been improved to test all of these already at the unit test stage (as opposed to the integration test stage). (The test was actually not functional as-is, because it did not cause pytest to fail on parse errors.)

pynestml/visitors/ast_builder_visitor.py
```diff
@@ -50,7 +50,10 @@ def visitNestMLCompilationUnit(self, ctx):
         for child in ctx.neuron():
             neurons.append(self.visit(child))
         # extract the name of the artifact from the context
-        artifact_name = ntpath.basename(ctx.start.source[1].fileName)
+        if hasattr(ctx.start.source[1], 'fileName'):
+            artifact_name = ntpath.basename(ctx.start.source[1].fileName)
+        else:
+            artifact_name = 'parsed from string'
```
pnbabu, line 56: Is it better to have this name with underscores instead of spaces?

tests/nestml_printer_test.py
```diff
@@ -168,11 +168,11 @@ def test_equations_block_without_comments(self):
...
-        stmt = '\n/*pre*/\n' \
-               'for i in 10 - 3.14...10 + 3.14 step -1: # in\n' \
-               'end\n' \
-               '/*post*/\n\n'
-        model = ModelParser.parse_for_stmt(stmt)
+        stmt = '\n  /*pre*/\n' \
+               '  for i in 10 - 3.14...10 + 3.14 step -1: # in\n' \
+               '  end\n' \
+               '  /*post*/\n\n'
+        model = ModelParser.parse_block(stmt)
```
pnbabu, line 175: Any reason for changing these to `parse_` functions?

clinssen: With newlines no longer going to channel 3, they need to be mentioned explicitly in grammar rules where they may be used. To allow for parsing the newlines in the test cases as they were defined in ``nestml_printer_test.py``, the functions have mostly been replaced by calls to parsing the parent or enclosing grammar rule (e.g. from *assignment* to *block*).

pnbabu (approved):

clinssen:
Thank you!

PR #621, opened 2021-02-16
GitHub actions for NESTML

pnbabu:
- Migrate content from `.travis.yml` to `.github/workflows/nestml-build.yml`
- Remove `.travis.yml` file
- Successful build link: https://github.com/pnbabu/nestml/actions/runs/571667802


jougs (approved):
Amazing work! Many thanks! I hereby approve, but leave the final decision to @clinssen.

.github/workflows/nestml-build.yml
```diff
@@ -0,0 +1,121 @@
...
+        run: |
+          wget http://www.antlr.org/download/antlr-4.8-complete.jar
+          echo \#\!/bin/bash > antlr4
+          echo java -cp \"`pwd`/antlr-4.8-complete.jar:$CLASSPATH\" org.antlr.v4.Tool \"\$@\" >> antlr4
+          echo >> antlr4
+          chmod +x antlr4
+          echo PATH=$PATH:`pwd` >> $GITHUB_ENV
+
+      # Install Python dependencies
+      - name: Python dependecies
```
clinssen, line 45:
```suggestion
      - name: Python dependencies
```

.github/workflows/nestml-build.yml
```diff
@@ -0,0 +1,121 @@
...
+
+      # Tests
+      - name: Run unit tests
+        run: |
+          pytest -s -o norecursedirs='*'  -o log_cli=true -o log_cli_level="DEBUG" tests
+
+      # Install NEST simulator
+      - name: NEST simulator
+        run: |
+          python -m pip install cython pytest-xdist
```
clinssen, line 81:
```suggestion
          python -m pip install cython
```
I believe this is no longer necessary.

clinssen (changes requested):
Many thanks for this! Just a few comments.

The "Install NESTML" stage appears twice in slightly different form, is this necessary?

clinssen:
One more suggestion. Could we use the same versions for Antlr (in Java) as for the Python Antlr4 runtime?  Right now the former is at 4.8 and the latter is at 4.9.1. This results in a lot of warning messages in the log. Either we can ==4.8 in requirements.txt, or you'll have to update the links to the Antlr .jar download. Cheers!

pnbabu:
> One more suggestion. Could we use the same versions for Antlr (in Java) as for the Python Antlr4 runtime? Right now the former is at 4.8 and the latter is at 4.9.1. This results in a lot of warning messages in the log. Either we can ==4.8 in requirements.txt, or you'll have to update the links to the Antlr .jar download. Cheers!

Yes, changed the version of the jar to 4.9.1

clinssen (approved):
Don't know why the removed .travis.yml is giving a merge conflict—if you could just fix that, I'm A-OK with the changes. Really great to have this, many thanks!

clinssen (approved):

PR #629, opened 2021-03-01
add split simulation NEST integration test

### LINKED ISSUE #611: Add integration test for splitting a simulation into multiple intervals
Check that ``nest.Simulate(100)`` yields the same behaviour as calling ``nest.Simulate(50)`` twice in a row.

Possibly, change the simulation resolution between the two calls.

This will check that calibrate() initialises internal variables correctly.

clinssen:
Fixes #611.

Changing the simulation resolution between two nest.Simulate() calls is not possible. Hence, this test is not as powerful as I had originally hoped. @jougs: do you know if there is any other way for a node's calibrate() method to be called more than once during the lifetime of the instance?

### DIFF
--- a/tests/nest_tests/nest_split_simulation_test.py
+++ b/tests/nest_tests/nest_split_simulation_test.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+#
+# nest_split_simulation_test.py
+#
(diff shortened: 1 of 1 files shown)


pnbabu (approved):
Looks good to me.

Issue #627, opened 2021-03-01
Remove code from the jinja template that inserts parameters and internals to the recordables map


pnbabu:
The variables declared in the `state` block change during the course of simulation and can be recorded by the multimeter. These are currently added to a recordable map based on the flag `is_recordable = true` in each variable.

On the other hand, the variables declared in the `internals` and `parameters` blocks remain unchanged during a simulation with the `is_recordable` flag set to `false`, hence, do not need to be recorded using a multimeter

In view of this, the following code which inserts variables from the `parameters` and `internals` block into a recordable map could be potentially removed from the `NeuronClass.jinja2` template
```
// internals
{%- filter indent(2,True) %}
{%- for variable in neuron.get_internal_symbols() %}
{%-   include "directives/RecordCallback.jinja2" %}
{%- endfor %}
{%- endfilter %}
  
  // parameters
{%- filter indent(2,True) %}
{%- for variable in neuron.get_parameter_symbols() %}
{%-   include "directives/RecordCallback.jinja2" %}
{%- endfor %}
{%- endfilter %}
```

PR #630, opened 2021-03-01
Templates cleanup

### LINKED ISSUE #574: Add NEST integration test for recording of state variables

pnbabu:
Fixes #574 and #627 

- Remove the code from the jinja template that inserts variables from the `internals` and `parameters` block into the `RecordablesMap`
- Add an integration test to test all types of recordable values: variables defined in `state` and `initial_values` blocks, `inline` expressions in the `equations` block with the `recordable` keyword
- Update documentation


tests/nest_tests/resources/RecordableVariables.nestml
```diff
@@ -0,0 +1,65 @@
...
+      V_ex = -10 mV
+
+      if V_abs >= V_thr:
+          V_abs = V_reset
+          emit_spike()
+      end
+    end
+
+    output: spike
+end
```
clinssen, line 65: no newline at end of file

tests/nest_tests/recordable_variables_test.py
```diff
@@ -0,0 +1,77 @@
...
+        # Get the recordable variables
+        events = nest.GetStatus(mm)[0]["events"]
+        V_m = events["V_m"]
+        self.assertIsNotNone(V_m)
+
+        V_abs = events["V_abs"]
+        self.assertIsNotNone(V_abs)
+
+        V_ex = events["V_ex"]
+        self.assertEqual(V_ex[-1], -10.0)
```
clinssen, line 77: I think it's OK here to use a strict equality check, since the value 10 is precisely representable using floating point numbers. But it might still be better to use a floating point comparison from np.testing.

Could we add the check that np.testing.assert_allclose(V_m, V_abs + V_reset)?

clinssen (changes requested):

tests/nest_tests/recordable_variables_test.py
pnbabu: Added this. Thanks.

doc/nestml_language.rst
```diff
@@ -1053,13 +1053,14 @@ Setting and retrieving model properties
 Recording values with devices
 -----------------------------
 
--  All values in the ``state`` block are recordable by a ``multimeter`` in NEST.
--  The ``recordable`` keyword can be used to also make variables in other blocks (``parameters, internals``) available to recording devices.
+-  All values in the ``state`` and ``initial_values`` blocks are recordable by a ``multimeter`` in NEST.
+-  The ``recordable`` keyword can be used to also make ``inline`` expressions in ``equations`` block available to recording devices.
```
jougs, line 1057:
```suggestion
-  The ``recordable`` keyword can be used to also make ``inline`` expressions in the ``equations`` block available to recording devices.
```

jougs (approved):
Just a very small suggestion. Otherwise I'm fine. Thanks for this contribution!

clinssen (approved):
Thank you!

PR #645, opened 2021-03-29
Fix code generation for modulo operator

pnbabu:
Add unit test to check the generated code for expressions in NEST.


clinssen (approved):
Cheers!

PR #647, opened 2021-03-30
Fix code generation for while and for loops

pnbabu:
- Remove an extra `}` in the templates that generated code for `while` and `for` loops.
- Add NEST integration tests for `while` and `for` loops


clinssen (approved):
Great, thanks!

PR #648, opened 2021-04-05
Refactor/fix references to 'function' where 'inline expression' is intended

clinssen:
No behavioural change; a purely internal refactoring.

Remaining incorrect references to 'function' were refactored to 'inline expressions'; the latter was a new syntax introduced in #515 to clear up the difference between that and callable functions with a return value.


pynestml/meta_model/ast_neuron.py
```diff
@@ -325,11 +324,10 @@ def get_parameter_symbols(self):
                 ret.append(symbol)
         return ret
 
-    def get_state_symbols(self):
+    def get_state_symbols(self) -> Iterable[VariableSymbol]:
```
pnbabu, line 327: Wouldn't it better to have `List` here instead of `Iterable`?

pynestml/utils/ast_nestml_printer.py
```diff
@@ -308,8 +308,8 @@ def print_declaration(self, node):
         ret += print_n_spaces(self.indent)
         if node.is_recordable:
             ret += 'recordable '
-        if node.is_function:
-            ret += 'function '
+        if node.is_inline_expression:
```
pnbabu, line 311: This attribute should also be changed in the `ASTDeclaration` class. Some tests are failing due to this.

pynestml/meta_model/ast_neuron.py
clinssen: You're right, this is probably better as it's more specific. Having an Iterable instead of a List might be desirable as an input parameter type, but is too nonspecific as a return type.

pynestml/utils/ast_nestml_printer.py
clinssen: Done!

pynestml/utils/ast_utils.py
```diff
@@ -195,15 +197,12 @@ def deconstruct_assignment(cls, lhs=None, is_plus=False, is_minus=False, is_time
         return expr
 
     @classmethod
-    def get_alias_symbols(cls, ast):
+    def get_inline_expression_symbols(cls, ast: ASTNode) -> Iterable[VariableSymbol]:
```
pnbabu, line 200: It could be `List` instead of `Iterable` here as well.

pynestml/meta_model/ast_neuron.py
```diff
@@ -19,16 +19,15 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-from typing import Optional, Union, List, Dict
+from typing import Iterable, Optional, Union, List, Dict
```
pnbabu, line 22:
```suggestion
from typing import Optional, Union, List, Dict
```

pnbabu (approved):
Looks good to me.

PR #651, opened 2021-04-26
Compartmental Model

pythonjam:
Submitting compartmental model implementation for review

Should you merge, you may want to exclude linkingModel.py (or not, up to you)
You may also want to exclude the directory
pynestml/codegeneration/resources_nest/cm_templates_initial 
which contains the original unparameterized files


clinssen:
Awesome, thanks! We held a preliminary review and have some initial thoughts/suggestions:

- please merge master
- remove unrelated files (IDE project files, linkingModel.py)
- some minor NESTML model file cleanup related to: initial_values block was removed from NESTML (merged with state block); empty blocks not required any longer
- request to add unit tests (for example, test that coco fails on incorrect model file)
- neuron_to_cm_info is generated in coco, but used later on during code generation. Suggest to not exchange data between coco and code generator. Perhaps move functions that generate this cm_info data structure to an external file, that is then invoked by both the coco and the code generator. (Data structure could be generated twice—probably not much computational overhead.) Perhaps move common code to ``pynestml/utils``.
- with_origins: adapt template/C++ to use State_. Make more similar to existing NESTML C++ code organisation. Then with_origins no longer needed.
- what is minimal difference between existing NESTML (non-compartmental model) neuron templates, and new templates?

clinssen:
Superseded by #668.

PR #663, opened 2021-06-20
Fix/refactor types pretty printer in code generator

clinssen:
Real-valued NESTML quantities were being printed as integers, which caused confusion for the C++ compiler.

Additionally, refactored the types printers into their own files and inheritance hierarchy.

### DIFF
--- a/pynestml/codegeneration/cpp_types_printer.py
+++ b/pynestml/codegeneration/cpp_types_printer.py
@@ -33,16 +33,16 @@ class CppTypesPrinter(TypesPrinter):
     def pretty_print(cls, element):
         if isinstance(element, bool) and element:
             return 'true'
-        
+
         if isinstance(element, bool) and not element:
             return 'false'
-        
+
         if isinstance(element, int):
             return str(element)
-        
+
         if isinstance(element, float):
             return "{0:E}".format(element)
-        
+
         if isinstance(element, str):
             return element
 
--- a/pynestml/codegeneration/cpp_types_printer.py
+++ b/pynestml/codegeneration/cpp_types_printer.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# cpp_types_printer.py
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Tuple
+
+from pynestml.codegeneration.types_printer import TypesPrinter
+
+
+class CppTypesPrinter(TypesPrinter):
+    """
+    Returns a C++ syntax version of the handed over element.
+    """
+
+    @classmethod
+    def pretty_print(cls, element):
+        if isinstance(element, bool) and element:
+            return 'true'
+        
+        if isinstance(element, bool) and not element:
+            return 'false'
+        
+        if isinstance(element, int):
+            return str(element)
+        
+        if isinstance(element, float):
+            return "{0:E}".format(element)
+        
+        if isinstance(element, str):
+            return element
+
+        raise Exception("Tried to print unknown type: " + str(type(element)) + " (string representation: " + str(element) + ")")
--- a/pynestml/codegeneration/expressions_pretty_printer.py
+++ b/pynestml/codegeneration/expressions_pretty_printer.py
@@ -21,7 +21,7 @@
 
 from typing import Tuple
 
-from pynestml.codegeneration.i_reference_converter import IReferenceConverter
+from pynestml.codegeneration.cpp_types_printer import CppTypesPrinter
 from pynestml.codegeneration.nestml_reference_converter import NestMLReferenceConverter
 from pynestml.meta_model.ast_expression import ASTExpression
 from pynestml.meta_model.ast_expression_node import ASTExpressionNode
@@ -40,8 +40,6 @@ class ExpressionsPrettyPrinter:
     """
 
     def __init__(self, reference_converter=None, types_printer=None):
-        # type: (IReferenceConverter,TypesPrinter) -> None
-        # todo by kp: this should expect a ITypesPrinter as the second arg
         if reference_converter is not None:
             self.reference_converter = reference_converter
         else:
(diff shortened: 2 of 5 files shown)


tfardet:
Unfortunately this does not seem to fix the issue: using max still converts ``0.`` in nestml file to ``0`` in the C++ file

clinssen:
That's odd, could you post a minimal reproducing case? The TypesPrinter now prints as ``"{0:E}".format()``, so for my local example:
```
neuron foo:
        state:
                x real = 42.
        end

        update:
                y real = max(0., x)
        end
end
```
It produces:

```C++
    // NESTML generated code for the update block:
    double y = std::max(0.000000E+00, S_.x);
```

tfardet:
Apparently the problem occurs if you use it in the ``equations`` block:

```
neuron foo:
        state:
                x real = 42.
        end

        equations:
                inline test real = max(x, 0)
        end

        update:
                y real = max(0., x)
                z real = 2*test
        end
end
```

tfardet:
OK, issue is whithin ``integrate_odes``

```
neuron foo:
        state:
                x real = 42.
        end

        equations:
            inline test real = max(x, 0.)

            x' = -0.5*test / (1*ms)
        end

        update:
                y real = max(0., x)
                z real = 2*test

                integrate_odes()
        end
end
```

clinssen:
I cannot reproduce any failure with the model from your last post. All mentions of max() get printed in the exponential format:
```C++
  inline double get_test() const  {
    return std::max(S_.ode_state[State_::x], 0.000000E+00);
```
```C++
extern "C" inline int foo_dynamics(double, const double ode_state[], double f[], void* pnode)
{
  // [...]
  f[State_::x] = -(5.000000E-01) * std::max(0.000000E+00, ode_state[State_::x]);
```
```C++
void foo::update(nest::Time const & origin,const long from, const long to)
{
    // [...]
    double y = std::max(0.000000E+00, S_.ode_state[State_::x]);
```
The generated C++ code compiles OK (I didn't try to instantiate the model).

tfardet:
Very strange...
here is what I get:

```cpp
extern "C" inline int foo_dynamics(double, const double ode_state[], double f[], void* pnode)
{
  // [...]
  f[State_::x] = -(5.000000E-01) * std::max(0, ode_state[State_::x]);
}
```
Others are fine though

```cpp
  inline double get_test() const
  {
    return std::max(S_.ode_state[State_::x], 0.000000E+00);
  }
```

```cpp
void foo::update(nest::Time const & origin,const long from, const long to)
{
    // [...]
    double y = std::max(0.000000E+00, S_.ode_state[State_::x]);/
```

The generated code is using GSL, but I don't cannot get PyGSL installed, could that be related?

tfardet:
Are there some cached files other than in the ``build`` and ``report`` folders?
One thing I noticed is that the order of the variables does not change when I switch them in the NESTML ``max`` function so I wonder if there's not a cache issue...

tfardet:
I'll try to track what's happening in NEST, could tell me whether [these NEST](https://github.com/Silmathoron/elif-madexp) models compile on your machine?

tfardet:
I tracked the problem up to [``ExpressionsPrettyPrinter.__do_print``](https://github.com/clinssen/nestml/blob/types-printer/pynestml/codegeneration/expressions_pretty_printer.py#L93) that showed my that the ``ASTSimpleExpression`` node has been generated as ``"max(0,x)"`` so the element that is passed to the type printer is already an int.
I guess the issue is therefore independent of the type printing which is correct.
Therefore I think this problem is probably independent of the PR.

tfardet:
Found the issue: it's in the [``odetoolbox.analysis`` call](https://github.com/clinssen/nestml/blob/types-printer/pynestml/codegeneration/nest_codegenerator.py#L546).
So the issue is coming from ``odetoolbox`` and not NESTML (pip tells me it's 2.3 but there's no visible ``__version__`` attribute)

pnbabu:
> Found the issue: it's in the [`odetoolbox.analysis` call](https://github.com/clinssen/nestml/blob/types-printer/pynestml/codegeneration/nest_codegenerator.py#L546): the expression enters with a floating zero and comes out with an integer one.
> So the issue is coming from `odetoolbox` and not NESTML (pip tells me it's 2.3 but there's no visible `__version__` attribute)

@tfardet with the latest NESTML 4.0 release, we have also released ode-toolbox 2.3. You could install it via `pip install ode-toolbox==2.3`. I have 2.3 on my system and I could reproduce the issue that you are facing. The result from the ode-toolbox is returning `max(0,x)` and the output C++ is without the floating point.
```
// ode_state[] here is---and must be---the state vector supplied by the integrator,
// not the state vector in the node, node.S_.ode_state[].
f[State_::x] = -(5.000000E-01) * std::max(0, ode_state[State_::x]);
```

tfardet:
Thanks @pnbabu 
Given that you managed to replicate it, I think there are two things to check:
- why this does not happen on @clinssen 's machine
- whether this was actually the only issue all along.

Therefore I propose to put this PR on hold for now and decide whether it should be closed or merged after odetoolbox is updated

clinssen:
@pnbabu: could you test if this also happens for you with the latest ODE-toolbox master?

It seems to work OK for me again (at least this is consistent):
```
In [11]: odetoolbox.analysis(
    ...:     indict={"dynamics": [{"expression": "x'=max(0, x)", "initial_value": "-1000"}]},
    ...:     disable_stiffness_check=True,
    ...:     log_level= 30
    ...: )
Out[11]: 
[{'update_expressions': {'x': 'max(0, x)'},
  'state_variables': ['x'],
  'initial_values': {'x': '-1000'},
  'solver': 'numeric'}]

In [13]: odetoolbox.analysis(
    ...:     indict={"dynamics": [{"expression": "x'=max(0., x)", "initial_value": "-1000."}]},
    ...:     disable_stiffness_check=True,
    ...:     log_level= 30
    ...: )
Out[13]: 
[{'update_expressions': {'x': 'max(0.0, x)'},
  'state_variables': ['x'],
  'initial_values': {'x': '-1000.00000000000'},
  'solver': 'numeric'}]
```

pnbabu:
@clinssen I still see the same results with the latest ode-toolbox master. 

```
odetoolbox.analysis(indict={"dynamics": [{"expression": "x'=max(0, x)", "initial_value": "-1000"}]},
                                  disable_stiffness_check=True, log_level= 30)

[{'update_expressions': {'x': 'max(0, x)'}, 'state_variables': ['x'], 'initial_values': {'x': '-1000'}, 
'solver': 'numeric'}]

odetoolbox.analysis(indict={"dynamics": [{"expression": "x'=max(0., x)", "initial_value": "-1000."}]}, 
                                 disable_stiffness_check=True, log_level= 30)

[{'update_expressions': {'x': 'max(0, x)'}, 'state_variables': ['x'], 'initial_values': {'x': '-1000.00000000000'}, 
'solver': 'numeric'}]
```

clinssen:
OK, this helps! It looks like (another) issue with the latest sympy. I was using version 1.4 before. Upgrading to 1.8, I can reproduce the issue. For now the recommendation would be to install sympy==1.4. I will take this up further on the sympy repository.

clinssen:
See https://github.com/sympy/sympy/issues/21740 for further upstream discussion.

pnbabu (approved):

PR #665, opened 2021-06-27
Add banner in documentation landing page about NESTML synapse feature

clinssen:
Alpha release of the synapse would probably amount to a merge into the master branch. While that is not yet ready, this is a proposal to direct the attention of visitors to the latest state of things.


jougs (approved):
Good idea!

clinssen:
Thanks for the quick review!

PR #669, opened 2021-07-21
Code generation with custom templates

pnbabu:
Fixes #643 

This PR adds an option to `codegen_opts`, to provide a path to the custom templates used for code generation. The options are as follows:

```
{" templates": 
  {
     "path": "/path/to/templates",
     "model_templates": ["NeuronClass.cpp.jinja2", ....],
     "module_templates": ["setup/ModuleHeader.h.jinja2", ...]
  }
}
```
- `path`: root directory of custom templates
- `model_templates`: a list of model template files
- `module_templates`: a list of template files for building a NEST extension module

PR #672, opened 2021-07-23
Add NEST2 compatibility

clinssen:
Add support for generating NEST2 (in particular, 2.20.1) compatible model+module code.

Based on #669, which should be merged first.


clinssen:
Should we add a CI stage to check NEST 2.20.1 compatibility?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -229,6 +229,14 @@ namespace nest
 //   Node initialization functions
 // ---------------------------------------------------------------------------
 
+#if NEST2_COMPAT
```
pnbabu, line 232: Would it be better to keep the template files in separate directories based on the version instead of having `#if` directives? Otherwise, we need to add these directives for each NEST released version, correct?

jougs: I would much prefer separate directories. That would also allow `diff`ing templates much more easily.

pnbabu (approved):

PR #673, opened 2021-07-23
Use nest::names::Name instead of string for recordables names

### LINKED ISSUE #625: RecordablesMap::insert_() should use nest::names instead of string
Upon instantiation of a generated C++ neuron class, the RecordablesMap is filled with entries, along the lines of:

```C++
RecordablesMap::insert_("V_m", &iaf_custom_nestml::get_V_m);
```

The fact that the recordable V_m is indexed by a string identifier ("V_m") causes issues with multithreading, because the RecordablesMap can only be copied thread-safely when elements are indexed by nest::names members, e.g.:

```C++
RecordablesMap::insert_(names::V_m, &iaf_custom_nestml::get_V_m);
```

To replace the strings with nest::names, a method is needed to dynamically (from the generated NESTML user module) insert elements into the nest::names namespace.

clinssen:
Fixes #625.

Before merging this, we should add a multiprocessing test to CI, that is, one where
```Python
nest.SetKernelStatus({"local_num_threads": T})
```
for T > 1. This is because nest::names::Name, but not strings, should be compatible with multithreading.

### DIFF
--- a/pynestml/codegeneration/resources_nest/NeuronClass.jinja2
+++ b/pynestml/codegeneration/resources_nest/NeuronClass.jinja2
@@ -71,6 +71,21 @@ nest::RecordablesMap<{{neuronName}}> {{neuronName}}::recordablesMap_;
 
 namespace nest
 {
+
+namespace names
+{
+{%- if recordable_state_variables|length > 0 %}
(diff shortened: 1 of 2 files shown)


clinssen:
Already fixed by #716.

PR #684, opened 2021-09-29
Fix NESTML logo in documentation page

### LINKED ISSUE #681: NESTML logo missing from documentation page
The NESTML logo has gone missing from https://nestml.readthedocs.org/, top left.

@pnbabu: assigning to you because you worked on #679; probably this issue is due to the same batch of upstream changes on RtD.

pnbabu:
Fixes #681 

Sphinx version `3.5.1` requires the logo file to be provided with the relative path of the logo file in `html_logo` variable of `conf.py`.

Documentation with the changes: https://nestml-doc.readthedocs.io/en/latest/index.html

### DIFF
--- a/doc/sphinx-apidoc/conf.py
+++ b/doc/sphinx-apidoc/conf.py
@@ -106,12 +106,12 @@ fns = [ fn for fn in fns if fn.endswith(".rst") and not "sphinx-apidoc" in fn ]
 print(fns)
 """
 for fn in matches:
-	if "sphinx-apidoc" in fn:
-		continue
-	fn_from = fn
-	fn_to = os.path.join(static_docs_dir, "sphinx-apidoc", fn[len(static_docs_dir)+1:])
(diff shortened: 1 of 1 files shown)


doc/sphinx-apidoc/conf.py
```diff
@@ -235,9 +235,9 @@
 
 
 def setup(app):
```
clinssen, line 237: ``custom.js`` does not exist and can be removed from here. Should we use the approach described in https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html and remove the setup() function entirely?

clinssen (changes requested):

doc/sphinx-apidoc/conf.py
pnbabu: Thanks, done.

clinssen (approved):
Great, thanks!

PR #691, opened 2021-10-25
Install Module from outside NEST lib directory

med-ayssar:
### Frontend:

1. Specify an arbitrary path to build and load the modules
2. Allow to use both absolute and relative paths for the `install_path`
3. Explicit addition of external existing libraries to `SLI`

### Setup:

1. Set `CMAKE_INSTALL_PREFIX` to `NEST_PREFIX` if `install_path` was not given, and append `lib/nest` to the path
2. Else, set `CMAKE_INSTALL_PREFIX` to `install_path` and don't add `lib/nest`  as suffix to the lib path
3. Delete the `Install doc` phase in `cmake` and all related `variables`


pynestml/frontend/pynestml_frontend.py
```diff
@@ -40,6 +42,8 @@
 from pynestml.utils.model_installer import install_nest as nest_installer
 
 
+
+
```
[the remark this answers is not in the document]
clinssen: Fixed in https://github.com/nest/nestml/pull/697.

pnbabu (commented):
Could you add an integration test for this? You could check the tests in `tests/nest_tests/` directory for reference.

pynestml/frontend/pynestml_frontend.py
```diff
@@ -117,12 +119,32 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+        # add the install_path to the python library
+        system = platform.system()
+        lib_key = ""
+
+        if system == "Linux":
+            lib_key = "LD_LIBRARY_PATH"
+        else:
+            lib_key = "DYLD_LIBRARY_PATH"
+
+        lib_path = os.path.join(install_path, "lib", "nest")
```
tfardet, line 141: I think the expected behavior would be to do as the docstring says: use the path given by the user without any change.
Now I can somewhat understand that we might want to append ``/nest`` to it to prevent collisions but ``/lib``? I don't think anyone would expect that...

The user should be expected to pass ``./local/lib`` and we use it directly.
It might be OK to add ``/nest`` to it in addition but in that case this should really be stated very clearly in the docstring.

pynestml/frontend/pynestml_frontend.py
```diff
@@ -117,12 +119,32 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+        if lib_key in os.environ:
+            os.environ[lib_key] += os.pathsep + lib_path
```
tfardet, line 143: might be good to check if the path is already in the environment variable and warn the user that it should be added there if not, especially since users that are going to encounter this issue are likely to be ones that did not compile NEST themselves and may not be familiar with the notion of paths

tfardet (changes requested):
Looks mostly good overall (though I did not test it yet).
I think we need to make a decision on precisely what path the user is expected to provide and whether this path should be modified at all or not.
I would be in favor on using the path provided by the user "as is", but I think appending ``/nest`` (and that only) could also be fine provided it is clearly explained is the docstring.

pynestml/frontend/pynestml_frontend.py
```diff
@@ -117,12 +119,32 @@ def install_nest(target_path: str, nest_path: str) -> None:
     nest_path : str
         Path to the NEST installation, which should point to the main directory where NEST is installed. This folder contains the ``bin``, ``lib(64)``, ``include``, and ``share`` folders of the NEST install. The ``bin`` folder should contain the ``nest-config`` script, which is accessed by NESTML to perform the installation. This path is the same as that passed through the ``-Dwith-nest`` argument of the CMake command before building the generated NEST module. The suffix ``bin/nest-config`` will be automatically appended to ``nest_path``.
 
+    install_path: str
+        Path to the install directory, where the generated module library will be created.
```
clinssen, line 123:
```suggestion
        Path to the directory where the generated NEST extension module will be installed into. If the parameter is not specified, the module will be installed into the NEST Simulator installation directory, as reported by nest-config.
```

pynestml/frontend/pynestml_frontend.py
```diff
@@ -117,12 +119,32 @@ def install_nest(target_path: str, nest_path: str) -> None:
     nest_path : str
```
clinssen, line 119: Could you remove the type annotations from the docstring here? These are redundant with Python type annotations now.

clinssen (changes requested):
There seem to be several points where the line has been wrapped at 80 characters; could you check if these are really necessary? We use a line length of 120 on our CI system. Please remove the ipynb files from the PR. Thanks!

pynestml/frontend/pynestml_frontend.py
med-ayssar: This will be done in another PR, because all the functions and methods have the redundant annotations.

pynestml/codegeneration/resources_nest/point_neuron/setup/CMakeLists.txt.jinja2
```diff
@@ -175,37 +177,30 @@ execute_process(
...
 
-# Install all binaries to NEST's install directories.
-set( CMAKE_INSTALL_LIBDIR ${NEST_LIBDIR}/nest CACHE STRING "object code libraries (lib/nest or lib64/nest or lib/<multiarch-tuple>/nest on Debian)" FORCE )
-set( CMAKE_INSTALL_DOCDIR ${NEST_DOCDIR} CACHE STRING "documentation root (DATAROOTDIR/doc/nest)" FORCE )
-set( CMAKE_INSTALL_DATADIR ${NEST_DATADIR} CACHE STRING "read-only architecture-independent data (DATAROOTDIR/nest)" FORCE )
+
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  # Use the `NEST_PREFIX` as `CMAKE_INSTALL_PREFIX`.
+	set( CMAKE_INSTALL_PREFIX ${NEST_PREFIX} CACHE STRING "Install path prefix, prepended onto install directories." FORCE )
+  # Retrieve libs folder in nest
```
jougs, line 187: This seems to repeat the code on lines 123-125 above. Is this by accident or by design?

pynestml/frontend/pynestml_frontend.py
```diff
@@ -228,7 +252,8 @@ def init_predefined():
 
 def create_report_dir():
     if not os.path.isdir(os.path.join(FrontendConfiguration.get_target_path(), '..', 'report')):
-        os.makedirs(os.path.join(FrontendConfiguration.get_target_path(), '..', 'report'))
+        os.makedirs(os.path.join(
+            FrontendConfiguration.get_target_path(), '..', 'report'))
```
jougs, line 256:
```suggestion
        os.makedirs(os.path.join(FrontendConfiguration.get_target_path(), '..', 'report'))
```

pynestml/utils/model_installer.py
```diff
@@ -45,13 +47,23 @@ def install_nest(target_path: str, nest_path: str) -> None:
     InvalidPathException
         If a failure occurs while trying to access the target path or the NEST installation path.
     """
+    cmake_cmd = ["cmake"]
+    if not os.path.isdir(nest_path):
+        raise InvalidPathException(
+            'NEST path (' + nest_path + ') is not a directory!')
```
jougs, line 53:
```suggestion
        raise InvalidPathException(f"NEST path ({nest_path}) is not a directory!")
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,31 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
     # first call cmake with all the arguments
     try:
-        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT, shell=shell,
-                                       cwd=str(os.path.join(target_path)))
+        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT,
+                                       shell=shell, cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
+        raise GeneratedCodeBuildException(
+            'Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
```
jougs, line 87:
```suggestion
        msg = "Error during 'cmake'. More detailed error messages can be found in stdout."
        raise GeneratedCodeBuildException(msg)
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,31 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
 
     # now execute make all
     try:
         subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
                               cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
+        raise GeneratedCodeBuildException(
+            'Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
```
jougs, line 95:
```suggestion
        msg = "Error during 'make all'. More detailed error messages can be found in stdout."
        raise GeneratedCodeBuildException(msg)
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,31 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
 
     # finally execute make install
     try:
         subprocess.check_call(make_install_cmd, stderr=subprocess.STDOUT, shell=shell,
                               cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
+        raise GeneratedCodeBuildException(
+            'Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
```
jougs, line 103:
```suggestion
        msg = "Error during 'make install'. More detailed error messages can be found in stdout."
        raise GeneratedCodeBuildException(msg)
```

tests/nest_tests/nest_custom_templates_test.py
```diff
@@ -53,7 +52,8 @@ def test_custom_templates(self):
                                  'setup/ModuleHeader.h.jinja2', 'setup/ModuleClass.cpp.jinja2']
         }}
 
-        to_nest(input_path, target_path, logging_level, module_name, store_log, suffix, dev, codegen_opts)
+        to_nest(input_path, target_path, logging_level,
+                module_name, store_log, suffix, dev, codegen_opts)
```
jougs, line 56:
```suggestion
        to_nest(input_path, target_path, logging_level, module_name, store_log, suffix, dev, codegen_opts)
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,31 @@ def install_nest(target_path: str, nest_path: str) -> None:
     else:
         shell = False
 
+    # remove CMakeCache.txt if exists
+    cmakeCache = os.path.join(target_path, "CMakeCache.txt")
+    if os.path.exists(cmakeCache):
+        os.remove(cmakeCache)
```
jougs, line 79:
```suggestion
    cmake_cache = os.path.join(target_path, "CMakeCache.txt")
    if os.path.exists(cmake_cache):
        os.remove(cmake_cache)
```

jougs (changes requested):
Thanks for the updates. I have some suggestions regarding mostly style and one question about the logic.

pynestml/codegeneration/resources_nest/point_neuron/setup/CMakeLists.txt.jinja2
med-ayssar: Ops, Yeah, it seems that I forgot to delete that block. Here I mean the block at 123-125.

pynestml/utils/model_installer.py
med-ayssar: Apart from the new line after the 79 characters, you still want to escape the single quote in the `cmake`

pynestml/utils/model_installer.py
med-ayssar: Again, those lines exceed the rule of 120 max-length.

pynestml/utils/model_installer.py
jougs: By breaking the line into two statements (`msg = ...` and `raise ...`), none of the lines exceeds the limits and there's no need to escape the single quotes inside the string if you enclose the whole string in double quotes instead.

pynestml/utils/model_installer.py
jougs: My line 94 has 90 characters, my line 95 has 47.

pynestml/utils/model_installer.py
jougs: Please change the variable name in order to be consistent with the rest of the code.

tests/nest_tests/nest_custom_templates_test.py
jougs: 107 characters.

jougs:
@med-ayssar: is there a specific reason for adding my changes manually instead of (batch) applying my suggestions? While the former of course also works, I think the latter would be much more transparent and easier to track by reviewers. It also minimizes the risk of typos and similar issues during the transfer.

pynestml/utils/model_installer.py
```diff
@@ -45,13 +47,23 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+    cmake_cmd.append(nest_config_path)
+
+    if install_path:
+        if not os.path.isabs(install_path):
+            install_path = os.path.abspath(install_path)
+        install_prefix = f"-DCMAKE_INSTALL_PREFIX={install_path}"
+        cmake_cmd.append(install_prefix)
+
     if not os.path.isdir(target_path):
         raise InvalidPathException('Target path (' + target_path + ') is not a directory!')
```
jougs, line 64:
```suggestion
        raise InvalidPathException(f"Target path ({target_path}) is not a directory!")
```

tests/nest_tests/nest_custom_templates_test.py
jougs: Why is this not applied and resolved without a comment?

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+
     # first call cmake with all the arguments
     try:
-        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT, shell=shell,
-                                       cwd=str(os.path.join(target_path)))
+        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT,
+                                       shell=shell, cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
+        msg = "Error occurred during 'cmake'! More detailed error messages can be found in stdout."
```
jougs, line 86:
```suggestion
        msg = "Error during 'cmake'. More detailed error messages can be found in stdout."
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+        raise GeneratedCodeBuildException(msg)
 
     # now execute make all
     try:
-        subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
-                              cwd=str(os.path.join(target_path)))
+        subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell, cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
+        msg = "Error occurred during 'make all'! More detailed error messages can be found in stdout."
```
jougs, line 93:
```suggestion
        msg = "Error during 'make all'. More detailed error messages can be found in stdout."
```

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+        raise GeneratedCodeBuildException(msg)
 
     # finally execute make install
     try:
         subprocess.check_call(make_install_cmd, stderr=subprocess.STDOUT, shell=shell,
                               cwd=str(os.path.join(target_path)))
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
+        msg = "Error occurred during 'make install'! More detailed error messages can be found in stdout."
```
jougs, line 101:
```suggestion
        msg = "Error during 'make install'. More detailed error messages can be found in stdout."
```

jougs (changes requested):

med-ayssar:
> @med-ayssar: is there a specific reason for adding my changes manually instead of (batch) applying my suggestions? While the former of course also works, I think the latter would be much more transparent and easier to track by reviewers. It also minimizes the risk of typos and similar issues during the transfer.

I was scrolling them down /up one by one and checking respectively the solved sub-task.

pynestml/utils/model_installer.py
```diff
@@ -45,13 +47,23 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
     """
+    cmake_cmd = ["cmake"]
+    if not os.path.isdir(nest_path):
+        raise InvalidPathException(f"NEST path ({nest_path}) is not a directory!")
+
+    nest_config_path = '-Dwith-nest=' + os.path.join(nest_path, 'bin', 'nest-config')
+    cmake_cmd.append(nest_config_path)
+
+    if install_path:
+        if not os.path.isabs(install_path):
```
pnbabu, line 58: Could you also check if `install_path` is a directory?

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+    # remove CMakeCache.txt if exists
+    cmake_cache = os.path.join(target_path, "CMakeCache.txt")
+    if os.path.exists(cmake_cache):
+        os.remove(cmake_cache)
+
     # first call cmake with all the arguments
     try:
-        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT, shell=shell,
-                                       cwd=str(os.path.join(target_path)))
+        result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT,
```
pnbabu, line 83: Is this a formatting issue due to line length?

pnbabu (commented):
As a general remark, you could set the line length to 120 in the editor you are using and later format these files.

pynestml/utils/model_installer.py
med-ayssar: `install_path` may not exist, and it will be created by `cmake` if necessary.

pynestml/utils/model_installer.py
med-ayssar: Yeah, the last `)` is at position 125.

tests/nest_tests/nest_custom_templates_test.py
jougs: Why is this line broken, when it is clearly below 120 characters?

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
+                                       shell=shell, cwd=str(os.path.join(target_path)))
```
jougs, line 84: What does the `join` on the `target_path` do here anyway? And shouldn't that be a string already?

pynestml/utils/model_installer.py
```diff
@@ -61,23 +73,30 @@ def install_nest(target_path: str, nest_path: str) -> None:
...
     except subprocess.CalledProcessError as e:
-        raise GeneratedCodeBuildException('Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
+        msg = "Error during 'cmake'. More detailed error messages can be found in stdout."
+        raise GeneratedCodeBuildException(msg)
 
     # now execute make all
     try:
-        subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
-                              cwd=str(os.path.join(target_path)))
+        subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell, cwd=str(os.path.join(target_path)))
```
jougs, line 91: See above.

jougs (changes requested):

tfardet:
From a logic standpoint, I think this now looks fine, I'll try to test it later this week with a repo-install of NEST (system-wide) and check if everything works as expected.

jougs (approved):
I'm formally fine with this now. I trust the other reviewers with testing this.

pnbabu:
> Could you add an integration test for this? You could check the tests in `tests/nest_tests/` directory for reference.

@med-ayssar could you add a test?

med-ayssar:
@jougs is there a safe way to unload `nestkernel` and `SLI` from memory?
Does `del nest` also remove all loaded shared objects used by `nestkernel` and `SLI` from memory, if not the case, how can I explicitly unload all related `nest` functionalities.
#### Why do I need that:
I want to test installing an already existing built module, but running the tests in the same script returns a `Module 'someNamemodule' is loaded already.` error.

#### How I want to do the tests.
##### Test 1
1.  Compile and build the module and copy it to tmp location ( here just using `to_nest` and `nest_install`)
2. Use `nest.Install` should not cause any errors, since `LD_LIBRARY_PATH` has the location of the module
##### Test 2

1.  Run `del nest`
2. Add new module path to `LD_LIBRARY_PATH` 
3. Use `nest.Install`
     * here I receive the above-mentioned error (`is loaded already`)

jougs:
There's no way of unloading a module from Python. For NEST, I'm working on module unloading in the context of https://github.com/nest/nest-simulator/issues/1717, https://github.com/nest/nest-simulator/issues/1642, https://github.com/nest/nest-simulator/pull/2208, but it will be a bit longer until something is ready.

Right now, you have to close Python and restart to get a clean slate.

tfardet (approved):
Confirmed working with debian system-wide nest install :+1:

med-ayssar:
> > Could you add an integration test for this? You could check the tests in `tests/nest_tests/` directory for reference.
> 
> @med-ayssar could you add a test?

Can you recheck it again?

tests/nest_tests/nest_install_module_in_different_location_test.py
```diff
@@ -0,0 +1,83 @@
...
+            os.pardir, os.pardir, "models", "neurons", f"{model_name}.nestml"))))
+        nest_path = nest.ll_api.sli_func("statusdict/prefix ::")
+        install_dir = tempfile.mkdtemp(prefix="nest_install", suffix="")
+        target_path = 'target'
+
+        logging_level = 'INFO'
+        store_log = False
+        suffix = '_location_test'
+        dev = True
+        codegen_opts = {"templates": {
```
pnbabu, line 51: `codegen_opts` is not necessary here as its default value is already set. You can remove it.

tests/nest_tests/nest_install_module_in_different_location_test.py
```diff
@@ -0,0 +1,83 @@
...
+            },
+            "module_templates": ['setup/CMakeLists.txt.jinja2',
+                                 'setup/ModuleHeader.h.jinja2', 'setup/ModuleClass.cpp.jinja2']
+        }}
+
+        to_nest(input_path, target_path, logging_level, module_name, store_log, suffix, dev, codegen_opts)
+        install_nest(target_path, nest_path, install_dir)
+
+        expected_found_module = f"{install_dir}/{module_name}.so"
+        actual_found_module = glob.glob(f"{install_dir}/*so")
```
pnbabu, line 65: Could you rename `expected_found_module` to `expected_module_path` and `actual_found_module` to `actual_module_path`?

pnbabu (commented):
Thanks for adding the test! I have some minor comments.

clinssen (approved):
I am in general happy with this, thanks!

In the light of #720, which removes the ``install_nest()`` function entirely, I propose to pass ``install_path`` as a keyword/value parameter in the code generator options dictionary instead, as it is a NEST-specific option.

med-ayssar:
> I am in general happy with this, thanks!
> 
> In the light of #720, which removes the `install_nest()` function entirely, I propose to pass `install_path` as a keyword/value parameter in the code generator options dictionary instead, as it is a NEST-specific option.

Wouldn't it be better to have a sort of static object representing a configuration for `pynestml`, and then the user can set the installation path to all `install_nest` (or `generate_target`) function calls.

PR #700, opened 2021-11-30
Remove help generator

clinssen:
``help_generator`` was renamed ``slihelp_generator`` in NEST Simulator. Instead of renaming, this PR removes all ties to this help generator entirely, as (correct me if I'm wrong) it has been superseded by ReadTheDocs/Sphinx.

### DIFF
--- a/doc/sphinx-apidoc/conf.py
+++ b/doc/sphinx-apidoc/conf.py
@@ -19,13 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-"""
+r"""
 Readthedocs configuration file
-------------------------------
-
-Use:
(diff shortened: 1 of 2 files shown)


jougs:
I'm fine with removing this from the NESTML side (I did not review yet), but still want to say that you're wrong ;-)

The RtD/Sphinx pipeline is solely for building user level documentation and model documentation from RST files/snippets, while the slihelp_generator is for estracting the SLI API documentation from the files containing SLI command definitions (.cpp, .h, .sli). As SLI modules can in principle contain such definitions, it was once considered a good idea to also run the slihelp_generator on them.

clinssen:
I initially thought that the help_generator was also still involved in the generation of local HTML files for models documentation and such, but if it really only pertains to SLI, it can probably be removed without much discussion.

jougs (approved):

PR #710, opened 2021-12-10
Add documentation link checker

clinssen:
From https://github.com/lycheeverse/lychee-action:

> This will check all repository links during any git push event and for all pull requests. If there's an error, it will fail the action. This has the benefit of ensuring that during a Pull Request, no link is added that is broken and any existing link will be caught if they become broken.


clinssen:
@terhorstd: could you comment on any security implications of running these random people's code on our CI? Thanks!

PR #716, opened 2021-12-23
Create nest Name objects for state and parameter variables

### LINKED ISSUE #625: RecordablesMap::insert_() should use nest::names instead of string

pnbabu:
Fixes #625 

The multithreading error is now solved by declaring all the `state` and `parameter` variables as `Name` objects inside the neuron and synapse header files. These `Name` objects are referenced wherever string equivalent of the variable names were used before.


clinssen:
Great solution with the namespaces. Do the ``@nest::delay`` and ``@nest::weight`` annotations still function the same way as before?

pnbabu:
@clinssen the `@nest::delay` and `@nest::weight` are untouched with the new changes as they are handled inside an if condition. However, the `names::` decorator was missing while handling `@nest::weight` variable which I have added now in this [commit](https://github.com/nest/nestml/pull/716/commits/24648644b3ce93018f7e44b65107ff878d822dcb)

jougs:
I think this is the best solution for now, although I would prefer if we could get rid of `Name`s altogether as soon as possible...

One question: Shouldn't similar functionality also be added to the templates in `pynestml/codegeneration/resources_nest/point_neuron_nest2` and to the ones for the compartmental neurons added in #709?

pnbabu:
@jougs the changes are added to templates in `pynestml/codegeneration/resources_nest/point_neuron/common` which are shared between the NEST 2.* and latest templates.

clinssen (approved):

jougs (approved):

PR #718, opened 2022-01-13
Print single nestml blocks as C++ code

med-ayssar:
- For the JIT project, extracting a lightweight version of the model is very important, and it must be independent of the NEST context.
- The idea is to extract the `States` and `Parameters` blocks from the NESTML file and create a smaller-partial version of the model that allows for querying and updating in the internal state of the model.

### What is being extracted:

-  Struct for the `States` and `Parameters` blocks 
- User defined functions in the NESTML file
- Constructor
- Getter/Setter for attributes in the `States` and `Parameters` blocks


jougs:
Can you please add a description of the problem to the description of the solution at the top (or add an issue that explains the problem you are trying to solve by this PR and link to that)? As it stands, I have no clue *why* this PR is needed.

clinssen:
Closing after merge of #718 which should already provide the necessary functionality; please re-open in case anything additional is needed.

PR #720, opened 2022-01-17
Add NEST 2 stage to CI

### LINKED ISSUE #503: Merge of to_nest() and nest_install()
`to_nest()` requires the user to specify the directory where generated code will be written into. The same directory needs to be specified as a parameter to `install_nest()`. If these functions are merged, and the user does not particularly care where the model is generated, there should be no need to specify a target path.

The target location should default to a place where NEST can find the generated `.so` file without additional specification of `LD_LIBRARY_PATH`. Possibly, `nest-config` should be interrogated for the requisite path.

clinssen:
Combines to_nest() and install_nest() NESTML API functions into a single generate_target(), thereby fixing #503.

Introduces the concept of an abstract base class, Builder, analogous to CodeGenerator.

Adds NEST 2.20.2 integration test to CI.


pynestml/codegeneration/builder.py
```diff
@@ -0,0 +1,76 @@
...
+    @staticmethod
+    def from_target_name(target_name: str, options: Optional[Mapping[str, Any]]=None) -> Builder:
+        """Static factory method that returns a new instance of a child class of Builder"""
+        assert target_name.upper() in Builder.get_known_targets(), "Unknown target platform requested: \"" + str(target_name) + "\""
+
+        if target_name.upper() in ["NEST", "NEST2"]:
+            from pynestml.codegeneration.nest_builder import NESTBuilder
+            return NESTBuilder(target_name, options)
+
+        return None   # no builder requested or available
```
med-ayssar, line 72: Won't be better to throw an error, so that we make sure if we have builder or not.

pynestml/codegeneration/code_generator.py
```diff
@@ -58,10 +58,12 @@ def generate_neuron_code(self, neuron: ASTNeuron) -> None:
...
         for k in options.keys():
-            assert k in self.__class__._default_options, "Option \"" + str(k) + "\" does not exist in code generator"
-        self._options.update(options)
+            if k in self.__class__._default_options:
+                self._options[k] = options[k]
+            else:
+                print("Warning: Option \"" + str(k) + "\" does not exist in code generator")
 
     def get_option(self, k):
         return self._options[k]
```
med-ayssar, line 69: I think the following would be better: `self._options.get(k, None)`, in case if the dictionary doesn't contain the key.

med-ayssar (commented):
Since the installation path of `nest` is constant, wouldn't better to set the default path without having the user explicitly calling those `nest` functions to extract the path?

tfardet:
I'm fairly suprised with this change of ``to_nest`` to ``generate_target`` that's going to break all existing code.
Is there a plan for *NEST*ML to support other platforms than NEST?
If not I'd really advise to reconsider this: repeatedly breaking backward compatibility tends to discourage users.
Given that it does not say anywhere that NESTML API is unstable and that we've been advertising NESTML for years now, people would (understandably) expect it to be stable.

clinssen:
> Since the installation path of `nest` is constant, wouldn't better to set the default path without having the user explicitly calling those `nest` functions to extract the path?

I like this idea, because it means that we don't have to explicitly pass ``nest_path`` in the code generation opts of every single call to ``generate_target()``. Because we should avoid for NESTML to ``import nest`` anywhere outside of the tests, I'll implement this as a single external Python script that just does ``import nest; print(nest.ll_api.sli_func("statusdict/prefix ::"))``, then call this script in ``nest_codegenerator.py`` while capturing stdout.

> I'm fairly suprised with this change of `to_nest` to `generate_target` that's going to break all existing code. Is there a plan for _NEST_ML to support other platforms than NEST? If not I'd really advise to reconsider this: repeatedly breaking backward compatibility tends to discourage users. Given that it does not say anywhere that NESTML API is unstable and that we've been advertising NESTML for years now, people would (understandably) expect it to be stable.

I agree completely that API changes should be kept to a minimum, but we have a growing number of code generation targets: due to evolution of the NEST API, we now even have two NEST code generators (for major versions 2 and 3); there is also "audodoc" for generating model documentation, "spinnaker" is hopefully coming soon, and I am secretly planning a minimal self-contained Python code generation option for debugging purposes.

These changes are still very flexible and up for discussion, for example, the following may or may not be true:
- Generating code without building it is not useful, so we don't need separate calls for that, just one "generate_and_build": ``generate_target()``
- All code generator-specific options should go into the ``codegen_opts`` dictionary. For instance, SpiNNaker (or other non-NEST code generators) might not have any notion of a "module", so perhaps the parameter ``module_name`` should also go into ``codegen_opts``.

I will create an issue to create a transition guide for NESTML 4 → 5, which will document the changes and how to adapt scripts.

tfardet:
> I agree completely that API changes should be kept to a minimum, but we have a growing number of code generation targets: due to evolution of the NEST API, we now even have two NEST code generators (for major versions 2 and 3); there is also "audodoc" for generating model documentation, "spinnaker" is hopefully coming soon, and I am secretly planning a minimal self-contained Python code generation option for debugging purposes.

This sound like very different targets and use cases, I'd rather have separate functions rather than an argument. It should also make it easier for new people to find them and understand their logic.

> These changes are still very flexible and up for discussion, for example, the following may or may not be true:
> 
>     * Generating code without building it is not useful, so we don't need separate calls for that, just one "generate_and_build": `generate_target()`

It can probably be useful when making new models, for debugging, and generally for user to understand what the NESTML code does, but probably minor, yes.

>     * All code generator-specific options should go into the `codegen_opts` dictionary. For instance, SpiNNaker (or other non-NEST code generators) might not have any notion of a "module", so perhaps the parameter `module_name` should also go into `codegen_opts`.

Again, I really think this screams for separate functions;

pynestml/codegeneration/builder.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
+        if not target.upper() in self.get_known_targets():
+            code, msg = Messages.get_unknown_target(target)
+            Logger.log_message(message=msg, code=code, log_level=LoggingLevel.ERROR)
+            self._target = ""
+            raise InvalidTargetException()
+
+        self._target = target
+        if "_default_options" in dir(self.__class__):
```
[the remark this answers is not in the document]
clinssen: Fixed by 9b51cad7.

pynestml/codegeneration/builder.py
clinssen: It's alright if there is a code generator (like for .rst documents) that doesn't have a build stage.

pynestml/codegeneration/code_generator.py
clinssen: I'd rather it throw an exception. What if there exists a known key but with value None?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -481,11 +481,7 @@ void {{neuronName}}::handle(nest::DataLoggingRequest& e)
 {% if has_spike_input %}
 void {{neuronName}}::handle(nest::SpikeEvent &e)
 {
-#if NEST2_COMPAT
```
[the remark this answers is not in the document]
clinssen: This was changed some time around NEST 2.18. @jougs: which NEST2 version(s) do we want to support here, does it make sense to retain support for older versions of NEST 2 (that is, older than 2.20.2)?

pynestml/codegeneration/builder.py
```diff
@@ -0,0 +1,76 @@
...
+            if k in self.__class__._default_options:
+                self._options[k] = options[k]
+            else:
+                print("Option \"" + str(k) + "\" does not exist in builder")
+
+    def get_option(self, k):
+        return self._options[k]
+
+    @staticmethod
+    def get_known_targets():
```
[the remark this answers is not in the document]
clinssen: Moved into the frontend.

doc/running.rst
```diff
@@ -39,7 +39,7 @@ where arguments are:
    * - ``--codegen_opts``
      - (Optional) Path to a JSON file containing additional options for the target platform code generator.
 
-Generated artifacts are copied to the selected target directory (default is ``target``). In order to install the models into NEST, the following commands have to be executed from within the target directory:
+Generated code is copied to the selected target directory. In order to install the models into NEST, the following commands have to be executed from within the target directory:
 
 .. code-block:: bash
```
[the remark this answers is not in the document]
clinssen: Documentation was revised.

pynestml/codegeneration/resources_nest/point_neuron/SynapseHeader.h.jinja2
```diff
@@ -18,1332 +18,4 @@ GNU General Public License for more details.
...
-  t_lastspike_ = t_trig;
-}
-
-
-{%- endif %}
-
-} // namespace
-
-#endif /* #ifndef {{synapseName.upper()}}_H */
+{%- include "common/SynapseHeader.h.jinja2" %}
```
med-ayssar, line 21: If it doesn't include new changes, why is it being moved to a different place

tests/nest_tests/resources/nestml_allmodels_codegen_opts.json
```diff
@@ -0,0 +1,15 @@
+{
```
pnbabu, line 1: Could you rename this file to include `nest_2` somewhere since this is for the NEST2 target?

tests/vector_code_generator_test.py
```diff
@@ -34,7 +34,7 @@
 from pynestml.symbols.predefined_functions import PredefinedFunctions
 from pynestml.symbols.predefined_types import PredefinedTypes
 from pynestml.symbols.predefined_units import PredefinedUnits
-from pynestml.frontend.pynestml_frontend import to_nest, install_nest
+from pynestml.frontend.pynestml_frontend import generate_target
```
pnbabu, line 37:
```suggestion
```

.github/workflows/nestml-build.yml
```diff
@@ -136,11 +135,9 @@ jobs:
           # exclude third factor plasticity models; these will only compile successfully if code generation is as a neuron+synapse pair
           export ALL_MODEL_FILENAMES=`find models/neurons -name "*.nestml" | paste -sd " "`
           echo $ALL_MODEL_FILENAMES
-          nestml --input_path $ALL_MODEL_FILENAMES --target_path target --suffix _nestml --logging_level INFO --module_name nestml_allmodels_module
-          cd target
           echo "NEST_INSTALL = ${{ env.NEST_INSTALL }}"
-          cmake -Dwith-nest=${{ env.NEST_INSTALL }}/bin/nest-config .
-          make && make install
+          sed -i 's|%NEST_PATH%|${{ env.NEST_INSTALL }}|' tests/nest_tests/resources/nest_codegen_opts.json
```
pnbabu, line 139: Is this required since the path is initialized in the `NestBuilder` class by default?

pynestml/frontend/pynestml_frontend.py
```diff
@@ -108,31 +155,35 @@ def to_nest(input_path: Union[str, Sequence[str]], target_path=None, logging_lev
...
-    else:
-        nest_installer(target_path, nest_path)
+    dev : bool, optional (default: False)
+        Enable development mode: code generation is attempted even for models that contain errors, and extra information is rendered in the generated code.
+    codegen_opts : Optional[Mapping[str, Any]]
+        A dictionary containing additional options for the target code generator.
+    """
+    generate_target(input_path, target_path, target_platform="NEST", logging_level=logging_level,
+                    module_name=module_name, store_log=store_log, suffix=suffix, install_path=install_path,
+                    dev=dev, codegen_opts=codegen_opts)
```
med-ayssar, line 186: Isn't the call to `add_libraries_to_sli` missing here?

tests/nest_tests/nest_install_module_in_different_location_test.py
```diff
@@ -36,33 +36,35 @@ class NestInstallExistingModule(unittest.TestCase):
     def test_installing_module_outside_nest(self):
 
         model_name = "iaf_psc_exp"
-        module_name = f'{model_name}module'
+        module_name = f"{model_name}module"
 
         input_path = os.path.join(os.path.realpath(os.path.join(os.path.dirname(__file__), os.path.join(
             os.pardir, os.pardir, "models", "neurons", f"{model_name}.nestml"))))
         nest_path = nest.ll_api.sli_func("statusdict/prefix ::")
```
med-ayssar, line 43: `nest_path` should be now removed, right?

pynestml/codegeneration/nest_builder.py
```diff
@@ -0,0 +1,124 @@
...
+        GeneratedCodeBuildException
+            If any kind of failure occurs during cmake configuration, build, or install.
+        InvalidPathException
+            If a failure occurs while trying to access the target path or the NEST installation path.
+        """
+        cmake_cmd = ["cmake"]
+        target_path = FrontendConfiguration.get_target_path()
+        install_path = FrontendConfiguration.get_install_path()
+        if install_path is not None:
+            add_libraries_to_sli(install_path)
```
med-ayssar, line 74: Doesn't this cause *Circular dependency*

pynestml/codegeneration/nest_builder.py
```diff
@@ -0,0 +1,124 @@
...
+                                           cwd=str(os.path.join(target_path)))
+        except subprocess.CalledProcessError as e:
+            raise GeneratedCodeBuildException('Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
+
+        # now execute make all
+        try:
+            subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
+                                  cwd=str(os.path.join(target_path)))
+        except subprocess.CalledProcessError as e:
+            raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
```
med-ayssar, line 117: I thought the PR containing the `add_libraries_to_sli` has modified the exceptions:
```suggestion
            msg = "Error during 'make all'. More detailed error messages can be found in stdout."
            raise GeneratedCodeBuildException(msg)
```

.github/workflows/nestml-build.yml
med-ayssar: `NestBuilder` does this also implicitly. Check [build function](https://github.com/clinssen/nestml/blob/643510f9ea6cce5081bf84ac7ae3c9ff8d0571a6/pynestml/codegeneration/nest_builder.py#L87), and the command is for compiling and building the generated code outside from the python environment.

pynestml/codegeneration/resources_nest/point_neuron/SynapseHeader.h.jinja2
clinssen: This is to support NEST2 compatibility, so the template in ``common`` can be included from the NEST3 templates directory, or from the NEST2 templates directory.

pynestml/frontend/pynestml_frontend.py
clinssen: It's in the NESTBuilder class.

pnbabu (approved):
Looks good to me!

clinssen:
Thank you for the comments and reviews!

PR #722, opened 2022-01-19
Remove reference to python2 in lexer/parser generation script

clinssen:


pnbabu:
Should we also generate the lexer and parser files with Python 3 and add them?

pnbabu (approved):

PR #723, opened 2022-01-19
Move pre, post and mod input port qualifiers from codegen_opts to synapse models

### LINKED ISSUE #721: Move code generator options into NESTML model
In the synapse, input ports are currently defined agnostic to whether they will connect to a pre- or postsynaptic partner:
```
  input:
    pre_spikes nS <- spike
    post_spikes nS <- spike
  end
```
They are only assigned during code generation, via the code generator options:
```
generate_target(...,
                codegen_opts={"neuron_synapse_pairs": [...,
                                                       "post_ports": ["post_spikes"]}]})
```
This works fine and is a flexible mechanism that was also adapted to map postsynaptic variables (like ``V_m``) onto synaptic continuous-valued input ports with a potentially different name (like ``post_V_m``), where they can be used as third factors for synaptic plasticity rules.

The difficulty is, that this information cannot be extracted from the NEST simulation script, which means it is incompatible with the proposed JIT workflow.

For this reason, we suggest to move the "meaning" of input ports into the NESTML synapse model itself, using keywords such as "pre" and "post" (and "modulatory"):
```
  input:
    pre_spikes nS <- pre spike
    post_spikes nS <- post spike
  end
```
This would be compatible with multisynapses, as there would just be more than one "pre" port.

The most general case, of mapping neuronal variables onto differently-named input ports, could be done by passing the code generator options by explicit call to the JIT mechanism, for instance
```
import nest_JIT as nest
nest.set_codegen_opts(...)
```
The disadvantage of this is that this is very far removed (in terms of lines of code) from the place in the script where the neurons and synapses actually get instantiated.

An external (my_network_codegen_opts.json) file would of course do the same trick.

clinssen:
Fixes #721.

Changes NESTML grammar to allow one of five predefined keywords (`pre`, `post`, `mod` in addition to the existing `excitatory` and `inhibitory`) as input port qualifiers.


pnbabu (commented):
I have just one comment. Looks good to me otherwise.

pnbabu (approved):
Looks good to me!

clinssen:
An alternative semantics for these input port qualifier keywords is to not define them like an enumeration in the grammar, but allow any "name-like" string (``[a-zA-Z_][a-zA-Z0-9_]*``), and leave the semantics up to the code generator used. For instance, the "excitatory" and "inhibitory" keywords are only relevant for the NEST target, as NEST uses negative weights to identify inhibitory spikes.

clinssen:
Rejected as these should by design not be part of the models themselves, but be specified outside of the models. These options will be passed through PyNEST instead when instantiating the models.

Issue #725, opened 2022-02-08
`ASTNeuron.clone` is not working properly


med-ayssar:
**Describe the bug**
Assuming directly after parsing the `ǹestml` files, you create a `deepcopy` with `ASTNeuron.clone` function, and then we proceed with calling the `CodeGenerator` to `transfrom` the `neurons` and `synapses`. Using the `clone` function, we should assume that the first instance won't be modified if the cloned version introduces new changes, but the problem here that the original instance will be modified as well.
**To Reproduce**
1. Get the parsed `ASTNeuron` and `ASTSynapse` instances without applying any modification
2. Call `neuron.get_internal_symbols() ` 
2. Call the `clone` function on each instance
3. Call the `analyse_transform` function on the cloned instances
4. Call again `neuron.get_internal_symbols() ` on the original version and the cloned version
5. Result: in Step 4, both original and cloned version return same objects, whereas in Step 2 we see different object
 * We expect the original to always return the same result

**Additional context**
Using `copy.deepcopy` doesn't fix the problem, as we get the following error when trying to run the `deepcopy` on the `Scope` instance: `cannot pickle '_io.textiowrapper' object`

clinssen:
``clone()`` only clones the AST nodes, but not the symbols. You should run a symbol table update after modifying the AST.

PR #726, opened 2022-02-09
Clone Scope in all ASTNode instances

med-ayssar:
Fixes [#725 ](https://github.com/nest/nestml/issues/725)

- Use `ASTSymbolTableVisitor` in the cloning function of `ASTNeuron` and `ASTSynapse`

### DIFF
--- a/pynestml/meta_model/ast_neuron.py
+++ b/pynestml/meta_model/ast_neuron.py
@@ -82,6 +82,10 @@ class ASTNeuron(ASTNeuronOrSynapse):
                         post_comments=[s for s in self.post_comments],
                         implicit_conversion_factor=self.implicit_conversion_factor)
 
+        from pynestml.visitors.ast_symbol_table_visitor import ASTSymbolTableVisitor
+        symtable_vistor = ASTSymbolTableVisitor()
+        symtable_vistor = ASTSymbolTableVisitor()
+        dup.accept(symtable_vistor)
         return dup
 
     def get_name(self):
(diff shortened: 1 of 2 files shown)


pynestml/meta_model/ast_neuron.py
```diff
@@ -82,6 +82,10 @@ def clone(self):
                         post_comments=[s for s in self.post_comments],
                         implicit_conversion_factor=self.implicit_conversion_factor)
 
+        from pynestml.visitors.ast_symbol_table_visitor import ASTSymbolTableVisitor
+        symtable_vistor = ASTSymbolTableVisitor()
+        symtable_vistor = ASTSymbolTableVisitor()
```
pnbabu, line 87:
```suggestion
```

pynestml/meta_model/ast_synapse.py
```diff
@@ -80,6 +80,10 @@ def clone(self):
                          post_comments=[s for s in self.post_comments],
                          implicit_conversion_factor=self.implicit_conversion_factor)
 
+        from pynestml.visitors.ast_symbol_table_visitor import ASTSymbolTableVisitor
+        symtable_vistor = ASTSymbolTableVisitor()
+        symtable_vistor = ASTSymbolTableVisitor()
```
pnbabu, line 85:
```suggestion
```

clinssen:
``clone()`` should strictly clone only the AST ("meta model"). Scope updates should be done outside of the cloning (by the calling function, just after ``clone()`` itself is called).

PR #728, opened 2022-02-15
Support for delay differential equations

### LINKED ISSUE #685: Add support for delay-based variables
Some state variables, for example, variables dependent on the membrane potential in the Clopath synapse model, can be updated based on some `delay` value. Whenever these variables are used later in the model, the `delay` should be accounted for. A new syntax for mentioning this `delay` should be added to NESTML.

For example, the state variable `u_bar_plus` is updated and stored with an associated time delay in the neuron. This should be later accounted for while updating the variable `x` that uses `u_bar_plus`.

```
u_bar_plus' = (-u_bar_plus + v_clamp) / tau_plus

x' = u_bar_plus(t - delay) + ....
```

pnbabu:
Fixes #685 

Implementation for supporting differential equations with delay-based variables.

Note: The PR will be in a draft state until all the test cases are added.

Test cases
- [ ] Delay differential equation with analytical solution
- [ ] Delay differential equation with numerical solution
- [ ] Delay differential equation with mixed (both analytical and numeric) solvers
- [ ] Equations with multiple delay-based variables


pynestml/codegeneration/resources_nest/point_neuron/common/NeuronHeader.jinja2
```diff
@@ -527,6 +527,15 @@ private:
     State_();
   };
 
+{%- if has_delay_variables %}
```
clinssen, line 530: Would these fit better in the ``State_`` struct? (Or maybe in ``Buffers_`` — @jougs?)

pynestml/meta_model/ast_expression.py
```diff
@@ -392,6 +394,7 @@ def equals(self, other):
             return False
         if self.is_ternary_operator() and other.is_ternary_operator() and \
                 not (self.get_condition().equals(other.get_condition())
-                     and self.get_if_true().equals(other.get_if_true()) and self.get_if_not().equals(other.get_if_not())):
+                     and self.get_if_true().equals(other.get_if_true())
+                     and self.get_if_not().equals(other.get_if_not())):
```
clinssen, line 398: Add condition on ``has_delay``.

pynestml/meta_model/ast_simple_expression.py
```diff
@@ -52,25 +52,21 @@ class ASTSimpleExpression(ASTExpressionNode):
...
-        :type boolean_literal: bool
         :param numeric_literal: a numeric value.
-        :type numeric_literal: float/int
         :param is_inf: is inf symbol.
-        :type is_inf: bool
         :param variable: a variable object.
-        :type variable: ASTVariable
         :param string: a single string literal
-        :type string: str
+        :param has_delay: is this simple expression node has a delay variable
```
clinssen, line 69:
```suggestion
        :param has_delay: whether this simple expression node has a delay variable
```

pynestml/meta_model/ast_simple_expression.py
```diff
@@ -207,6 +205,24 @@ def is_variable(self):
...
 
+    def is_delay_variable(self):
+        """
+        Returns whether it is a delay variable or not
+        :return: True if the variable has a delay parameter, False otherwise
+        """
+        if self.is_variable() and self.has_delay \
+                and self.get_variable().get_delay_parameter() is not None:
+            return True
+        else:
```
clinssen, line 216: ``else`` not necessary (return-early pattern)

pynestml/meta_model/ast_variable.py
```diff
@@ -160,8 +164,23 @@ def set_size_parameter(self, vector_parameter):
...
+
+    def set_delay_parameter(self, delay: str):
+        """
+        Updates the current delay parameter to the handed over value
+        :param delay: delay parameter
+        """
+        assert (delay is not None), '(PyNestML.AST.Variable) No delay parameter provided'
+        self.delay_parameter = delay
+
+    def get_parent(self, ast):
```
clinssen, line 182: type annotation lost

pynestml/utils/ast_utils.py
```diff
@@ -859,3 +858,44 @@ def get_statements_from_block(cls, var_name, block):
...
                 stmts.append(node)
         return stmts
+
+    @classmethod
+    def has_delay_variable(cls, node: ASTFunctionCall) -> bool:
+        """
+        Checks if the given function call is actually a variable with a delay differential part
+        :param node: The expression
+        """
+        # Check if the function name is a state variable
```
clinssen, line 868: Could you re-write this function as a one-liner that just checks whether ``get_delay_variable_symbol(node)`` returns ``None``?

pynestml/visitors/ast_function_call_visitor.py
```diff
@@ -66,10 +81,11 @@ def visit_simple_expression(self, node):
 
             if isinstance(return_type, TemplateTypeSymbol):
                 # error: return type template not found among parameter type templates
-                assert(False)
+                assert (False)
```
clinssen, line 84:
```suggestion
                assert False, "return type template not found among parameter type templates"
```
Should this raise an exception or be an assertion? (Programmer error or user error?)

tests/nest_tests/nest_delay_based_variables_test.py
```diff
@@ -0,0 +1,199 @@
...
+    def setUp(self):
+        PredefinedUnits.register_units()
+        PredefinedTypes.register_types()
+        PredefinedFunctions.register_functions()
+        PredefinedVariables.register_variables()
+        SymbolTable.initialize_symbol_table(
+            ASTSourceLocation(start_line=0, start_column=0, end_line=0, end_column=0))
+        Logger.init_logger(LoggingLevel.INFO)
+
+        self.target_path = 'target_delay'
```
clinssen, line 60: All double quotes, please.

tests/nest_tests/resources/DelayBasedVariablesWithNumericalSolver.nestml
```diff
@@ -0,0 +1,88 @@
...
+neuron izhikevich_delay:
+
+  state:
+    V_m mV = V_m_init         # Membrane potential
+    U_m real = b * V_m_init   # Membrane potential recovery variable
+  end
+
+  equations:
+    V_m' = ( 0.04 * V_m * V_m / mV + 5.0 * V_m + ( 140 - U_m ) * mV + ( (I_e + I_stim) * GOhm ) ) / ms
+    U_m' = a * (b * V_m(t-delay) - U_m * mV) / (mV*ms)
```
clinssen, line 46:
```suggestion
    U_m' = a * (b * V_m(t - delay) - U_m * mV) / (mV*ms)
```

clinssen (changes requested):
This is a really awesome feature!

Could you please also writing some documentation, and/or perhaps an iPython notebook?

It should define what happens when you do ``GetStatus()`` or ``SetStatus()`` on a delayed variable.

pynestml/cocos/co_co_function_argument_template_types_consistent.py
```diff
@@ -73,6 +74,13 @@ def visit_simple_expression(self, node):
             return
         function_name = node.get_function_call().get_name()
         method_symbol = scope.resolve_to_symbol(function_name, SymbolKind.FUNCTION)
+
+        if method_symbol is None and ASTUtils.has_delay_variable(node.get_function_call()):
+            code, message = Messages.get_function_is_delay_variable(function_name)
+            Logger.log_message(code=code, message=message, error_position=node.get_source_position(),
+                               log_level=LoggingLevel.WARNING)
```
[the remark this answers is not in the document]
pnbabu: This coco is run before the function call (which is actually a delayed variable) is transformed to an AST variable. Since it is not technically a function call, further checks are not required. Hence, a warning is issued and returned.

pynestml/visitors/ast_function_call_visitor.py
pnbabu: This change is a result of the auto-format of the code by the editor.  The same if condition is also present on line 76. Isn't this redundant?

pynestml/codegeneration/nest_code_generator.py
```diff
@@ -974,10 +1001,11 @@ def _get_synapse_model_namespace(self, synapse: ASTSynapse) -> Dict:
         namespace["tracing"] = FrontendConfiguration.is_dev
         namespace["has_state_vectors"] = False
         namespace["vector_symbols"] = []
+        namespace['has_delay_variables'] = synapse.has_delay_variables()
         namespace['names_namespace'] = synapse.get_name() + "_names"
 
         # event handlers priority
-        # XXX: this should be refactored in case we have additional modulatory (3rd-factor) spiking input ports in the synapse
+        # XXX: this should be refactored in case we have additional modulatory (3rd-factor) spiking input ports in
```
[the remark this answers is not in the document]
pnbabu: Lost in auto-formatting. Restored now.

pynestml/codegeneration/nest_code_generator.py
```diff
@@ -1133,16 +1164,24 @@ def _get_neuron_model_namespace(self, neuron: ASTNeuron) -> Dict:
...
                 expr_ast = ModelParser.parse_expression(expr_str)
-                # pretend that update expressions are in "equations" block, which should always be present, as differential equations must have been defined to get here
+                # pretend that update expressions are in "equations" block, which should always be present,
+                # as differential equations must have been defined to get here
                 expr_ast.update_scope(neuron.get_equations_blocks().get_scope())
                 expr_ast.accept(ASTSymbolTableVisitor())
                 namespace["update_expressions"][sym] = expr_ast
 
+                # Check if the update expression has delay variables
+                if has_delay_variables(neuron.equations_with_delay_vars, sym):
```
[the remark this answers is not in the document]
pnbabu: The new visitor `ASTMarkDelayVarsVisitor` needs to be run on the new `expr_ast` which contains the solver's update expression. This expression is in turn stored in `namespace["update_expressions"]` to be used in templates.

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronHeader.jinja2
pnbabu: It makes sense to add the state variable's corresponding delayed vector in the `State_` struct. What about the helper variables used to maintain the index and size of the vector?

pynestml/visitors/ast_function_call_visitor.py
clinssen: The variable ``return_type`` can change inbetween.

pynestml/cocos/co_co_function_argument_template_types_consistent.py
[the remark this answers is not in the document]
clinssen: Is it necessary to warn though, if everything checks out but we are just swapping out an ``ASTFunctionCall`` for an ``ASTVariable`` "behind the scenes"? Perhaps INFO or DEBUG logging level instead?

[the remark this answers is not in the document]
pnbabu: Yes, you are right. There is no need for a warning. I have changed it to DEBUG.

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronHeader.jinja2
pnbabu: Added all the helper variables for delay-based variables into a new struct named `DelayedVariables_`.

tests/nest_tests/nest_delay_based_variables_test.py
```diff
@@ -0,0 +1,199 @@
...
+        # Run the test
+        yield
+
+        # clean up
+        import shutil
+        if self.target_path:
+            try:
+                shutil.rmtree(self.target_path)
+            except Exception:
+                pass
```
clinssen, line 199: This test does not actually contain any assertions. It would be great if we had separate delayed and non-delayed variables, so that we could then numerically check that when we remove the delay again from the curves in Python, they overlap to floating-point precision. Or you could numerically check some simple cases of DDEs (like oscillation); see for example https://fse.studenttheses.ub.rug.nl/14274/1/Delay_Differential_Equations.pdf.

pnbabu: I added simpler test cases for numeric and mixed solver cases. I have also added assert for the analytical solver case as the equations are linear and the results can be easily compared. However, I haven't found a way to assert for the numeric and mixed solvers yet. Do you have any ideas?

clinssen:
We would not be able to simulate the chaotic Mackey-Glass equation, right? Could you perhaps elaborate a little bit in the documentation what this feature can and cannot do? As our original goal is to implement the Clopath synapse, I feel we should focus on that rather than let this feature take on too much of a life of its own.

pnbabu:
> We would not be able to simulate the chaotic Mackey-Glass equation, right? Could you perhaps elaborate a little bit in the documentation what this feature can and cannot do? As our original goal is to implement the Clopath synapse, I feel we should focus on that rather than let this feature take on too much of a life of its own.

Yes, this is currently not supported. I have updated the documentation.

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -846,6 +846,45 @@ has to be defined in the ``state`` block. Otherwise, an error message is generat
...
+
+Here, the ``delay`` variable is defined in the ``parameters`` block as:
+
+.. code-block:: nestml
+
+   parameters:
+     tau ms = 3.5 ms
+     delay ms = 5.0 ms
+   end
+
```
clinssen, line 876: Could you link to the unit test so that people can use it as an example?

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -846,6 +846,45 @@ has to be defined in the ``state`` block. Otherwise, an error message is generat
...
+   The value of the delayed variable (``bar`` in the above example) returned by the node's ``get()`` function in
+   PyNEST is always the non-delayed version, i.e., the value of the derivative of ``bar`` at time ``t``. Similarly, the
+   ``set()`` function sets the value of the actual state variable ``bar`` without the ``delay`` into consideration.
+
+.. note::
+
+   - Delay differential equations where the derivative of a variable is dependent on the derivative of the same
+     variable at previous times, for example, `The Mackey-Glass equation <http://www.scholarpedia.org/article/Mackey-Glass_equation>`_, are not supported currently.
+   - Delay differential equations with multiple delay values for the same variable are also not supported.
```
clinssen, line 887: What does this mean exactly? Could you also indicate as the first point that ``delay`` has to be either a (constant) parameter or a numeric constant? Indicate what happens when someone tries to set the delay parameter via PyNEST.

pnbabu: Done. The doc build can be found here: https://nestml-doc.readthedocs.io/en/latest/index.html

clinssen:
Great, thanks!

PR #732, opened 2022-02-28
Raise proper exception for non-existing code generator/builder options

### LINKED ISSUE #731: Passing `codegen_opts` from `generate_nest_target` to `NESTBuilder`
When calling the `generate_nest_target` with  the `codegen_opts`, the options might include `neuron_parent_class` that is only required for the code generation. However, `NESTBuilder` asks the `FrontendConfiguration` for these options and compare them with the `_default_options` that only include ` {"nest_path": None }`, and then prints that the  `neuron_parent_class`  does not exist in the builder, which might make the user a bit unsure if the shown information is of importance to the execution of the model.

clinssen:
Fixes #731.

### DIFF
--- a/pynestml/codegeneration/nest_code_generator.py
+++ b/pynestml/codegeneration/nest_code_generator.py
@@ -213,6 +213,12 @@ class NESTCodeGenerator(CodeGenerator):
 
         return _abs_template_paths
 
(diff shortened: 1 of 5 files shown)


med-ayssar:
It looks good.

med-ayssar (approved):

clinssen:
Cheers!

PR #738, opened 2022-03-14
Split code generation phases

### LINKED ISSUE #739: Split `nestml` code generation phases
Executing `generate_taget`, it calls the `nestml` code generation pipeline without having independent functions to separate the code generation phases. Mainly, the pipeline should  have four stages, `parse,  transform, generate_code`, and `build`. For the `JIT` project, we need to have control over these phases and not calling them in one block. Therefore, `JIT` must split between two (parsing, transforming) and (generating code, building). 

A use case would be generating the `neuron_synapse` code from a previously seen `neuron` model, however in that particular face we have `moving` variables from the `synapse` model to `ǹeuron` model, and the responsible function for inserting new variable in the `ǹeuron` class is done inside the transformation phase, and thus we must have a separate function to transform the `AstNeuron" before going to the code generation.

med-ayssar:
Fixes #739


clinssen:
Could you provide the background motivation for why ``transform()`` should be part of the "public API" rather than an internal affair of the code generator(s)?

med-ayssar:
> Could you provide the background motivation for why `transform()` should be part of the "public API" rather than an internal affair of the code generator(s)?
Yeah, once I make sure that the pipeline is green, I will create an issue and link it to the PR

clinssen:
Continued in #764.

PR #741, opened 2022-03-14
Redirect stdout and stderr in the building phase

### LINKED ISSUE #740: Allow `nest_builder` to forward `cmake` outputs to file
For the `JIT` project, we don't need to show the user the outputs of `cmake` while compiling and building the generated code in the background, therefore we need to forward the `stderr` and the `stdout` to a file in a user (`JIT`) defined locations.

med-ayssar:
Fixes #740

### DIFF
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -68,16 +68,18 @@ class Builder(WithOptions, metaclass=ABCMeta):
         redirect = None
         error_location = ""
 
-        print(options)
         if options and len(options) > 0 and require_redirect_key in options and options.get(require_redirect_key, False):
             if redirection_path_key in options:
-                stdout = os.path.join(options[redirection_path_key], "build_output.txt")
-                stderr = os.path.join(options[redirection_path_key], "build_error.txt")
-                redirect = True
-                error_location = stderr
+                if os.path.isdir(options[redirection_path_key]):
+                    stdout = os.path.join(options[redirection_path_key], "build_output.txt")
+                    stderr = os.path.join(options[redirection_path_key], "build_error.txt")
+                    redirect = True
+                    error_location = stderr
 
-                stdout = open(stdout, "w")
-                stderr = open(stderr, "w")
+                    stdout = open(stdout, "w")
+                    stderr = open(stderr, "w")
+                else:
+                    raise Exception(f"The provided directory {options[redirection_path_key]} does not exist in your system!")
 
             else:
                 target_path = FrontendConfiguration.get_target_path()
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -20,12 +20,15 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import annotations
+import subprocess
+import os
 
 from typing import Any, Mapping, Optional
 
 from abc import ABCMeta, abstractmethod
 
 from pynestml.exceptions.invalid_target_exception import InvalidTargetException
+from pynestml.frontend.frontend_configuration import FrontendConfiguration
 from pynestml.utils.logger import LoggingLevel, Logger
 from pynestml.utils.messages import Messages
 from pynestml.utils.with_options import WithOptions
@@ -34,8 +37,12 @@ from pynestml.utils.with_options import WithOptions
 class Builder(WithOptions, metaclass=ABCMeta):
     r"""Compile, build and install the code for a given target platform. Runs after the CodeGenerator."""
 
-    def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
+    redirect_build_output = {"redirect_build_output": False,
+                             "build_output_dir": None}
+
+    def __init__(self, target, options: Optional[Mapping[str, Any]] = None):
         super(Builder, self).__init__(options)
+        self.process_output_redirection_(options)
         from pynestml.frontend.pynestml_frontend import get_known_targets
 
         if not target.upper() in get_known_targets():
@@ -46,6 +53,54 @@ class Builder(WithOptions, metaclass=ABCMeta):
 
         self._target = target
 
+    def __del__(self):
+        if self.get_option("redirect"):
+            self.get_option("stdout").close()
+            self.get_option("stderr").close()
+
+    def process_output_redirection_(self, options):
+        dict_keys = list(Builder.redirect_build_output.keys())
+        require_redirect_key = dict_keys[0]
+        redirection_path_key = dict_keys[1]
+
+        stdout = None
+        stderr = None
+        redirect = None
+        error_location = ""
+
+        print(options)
+        if options and len(options) > 0 and require_redirect_key in options and options.get(require_redirect_key, False):
+            if redirection_path_key in options:
+                stdout = os.path.join(options[redirection_path_key], "build_output.txt")
+                stderr = os.path.join(options[redirection_path_key], "build_error.txt")
+                redirect = True
+                error_location = stderr
+
+                stdout = open(stdout, "w")
+                stderr = open(stderr, "w")
+
+            else:
+                target_path = FrontendConfiguration.get_target_path()
+                stdout = os.path.join(target_path, "build_output.txt")
+                stderr = os.path.join(target_path, "build_error.txt")
+                redirect = True
+                error_location = stderr
+
+                stdout = open(stdout, "w")
+                stderr = open(stderr, "w")
+        else:
+            stderr = subprocess.STDOUT
+            redirect = False
+            error_location = "stderr"
+
+        self.add_options({"stdout": stdout, "stderr": stderr, "redirect": redirect, "error_location": error_location})
+
     @abstractmethod
     def build(self) -> None:
         pass
+
+    def set_options(self, options: Mapping[str, Any]) -> Mapping[str, Any]:
+        ret = super().set_options(options)
+        ret.pop("redirect_build_output", None)
+        ret.pop("build_output_dir", None)
+        return ret
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -63,42 +63,40 @@ class Builder(WithOptions, metaclass=ABCMeta):
         require_redirect_key = dict_keys[0]
         redirection_path_key = dict_keys[1]
 
+        # default values. The output will be printed to the console.
         stdout = None
-        stderr = None
-        redirect = None
-        error_location = ""
+        stderr = subprocess.STDOUT
+        redirect = False
+        error_location = "stderr"
 
         if options and len(options) > 0 and require_redirect_key in options and options.get(require_redirect_key, False):
+            output_file_name = f"{self.get_builder_name()}_output.txt"
+            error_file_name = f"{self.get_builder_name()}_error.txt"
+
             if redirection_path_key in options:
                 if os.path.isdir(options[redirection_path_key]):
-                    stdout = os.path.join(options[redirection_path_key], "build_output.txt")
-                    stderr = os.path.join(options[redirection_path_key], "build_error.txt")
-                    redirect = True
-                    error_location = stderr
-
-                    stdout = open(stdout, "w")
-                    stderr = open(stderr, "w")
+                    stdout = os.path.join(options[redirection_path_key], output_file_name)
+                    stderr = os.path.join(options[redirection_path_key], error_file_name)
                 else:
                     raise Exception(f"The provided directory {options[redirection_path_key]} does not exist in your system!")
-
             else:
                 target_path = FrontendConfiguration.get_target_path()
-                stdout = os.path.join(target_path, "build_output.txt")
-                stderr = os.path.join(target_path, "build_error.txt")
-                redirect = True
-                error_location = stderr
+                stdout = os.path.join(target_path, output_file_name)
+                stderr = os.path.join(target_path, error_file_name)
 
-                stdout = open(stdout, "w")
-                stderr = open(stderr, "w")
-        else:
-            stderr = subprocess.STDOUT
-            redirect = False
-            error_location = "stderr"
+            error_location = stderr
+            stdout = open(stdout, "w")
+            stderr = open(stderr, "w")
+            redirect = True
 
         self.add_options({"stdout": stdout, "stderr": stderr, "redirect": redirect, "error_location": error_location})
 
     @abstractmethod
     def build(self) -> None:
+        pass
+
+    @abstractmethod
+    def get_builder_name(self) -> str:
         pass
 
     def set_options(self, options: Mapping[str, Any]) -> Mapping[str, Any]:
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -51,7 +51,6 @@ class Builder(WithOptions, metaclass=ABCMeta):
         self._target = target
 
     def __del__(self):
-        print("======= Testting __del__ features +======")
         if self.get_option("redirect"):
             self.get_option("stdout").close()
             self.get_option("stderr").close()
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -50,11 +50,6 @@ class Builder(WithOptions, metaclass=ABCMeta):
 
         self._target = target
 
-    def __del__(self):
-        if self.get_option("redirect"):
-            self.get_option("stdout").close()
-            self.get_option("stderr").close()
-
     def process_output_redirection_(self, options):
         require_redirect_key = options.get("redirect_build_output", False)
         redirection_path_key = options.get("build_output_dir", "")
@@ -66,8 +61,8 @@ class Builder(WithOptions, metaclass=ABCMeta):
         error_location = "stderr"
 
         if options and len(options) > 0 and require_redirect_key:
-            output_file_name = f"{self.get_builder_name()}_output.txt"
-            error_file_name = f"{self.get_builder_name()}_error.txt"
+            output_file_name = f"{self.get_builder_name()}_stdout.txt"
+            error_file_name = f"{self.get_builder_name()}_stderr.txt"
 
             if redirection_path_key != "":
                 if not os.path.isdir(redirection_path_key):
@@ -81,8 +76,6 @@ class Builder(WithOptions, metaclass=ABCMeta):
                 stderr = os.path.join(target_path, error_file_name)
 
             error_location = stderr
-            stdout = open(stdout, "w")
-            stderr = open(stderr, "w")
             redirect = True
 
         self.add_options({"stdout": stdout, "stderr": stderr, "redirect": redirect, "error_location": error_location})
--- a/pynestml/codegeneration/builder.py
+++ b/pynestml/codegeneration/builder.py
@@ -37,9 +37,6 @@ from pynestml.utils.with_options import WithOptions
 class Builder(WithOptions, metaclass=ABCMeta):
     r"""Compile, build and install the code for a given target platform. Runs after the CodeGenerator."""
 
-    redirect_build_output = {"redirect_build_output": False,
-                             "build_output_dir": None}
-
     def __init__(self, target, options: Optional[Mapping[str, Any]] = None):
         super(Builder, self).__init__(options)
         self.process_output_redirection_(options)
@@ -54,14 +51,14 @@ class Builder(WithOptions, metaclass=ABCMeta):
         self._target = target
 
     def __del__(self):
+        print("======= Testting __del__ features +======")
         if self.get_option("redirect"):
             self.get_option("stdout").close()
             self.get_option("stderr").close()
 
     def process_output_redirection_(self, options):
-        dict_keys = list(Builder.redirect_build_output.keys())
-        require_redirect_key = dict_keys[0]
-        redirection_path_key = dict_keys[1]
+        require_redirect_key = options.get("redirect_build_output", False)
+        redirection_path_key = options.get("build_output_dir", "")
 
         # default values. The output will be printed to the console.
         stdout = None
@@ -69,16 +66,16 @@ class Builder(WithOptions, metaclass=ABCMeta):
         redirect = False
         error_location = "stderr"
 
-        if options and len(options) > 0 and require_redirect_key in options and options.get(require_redirect_key, False):
+        if options and len(options) > 0 and require_redirect_key:
             output_file_name = f"{self.get_builder_name()}_output.txt"
             error_file_name = f"{self.get_builder_name()}_error.txt"
 
-            if redirection_path_key in options:
-                if os.path.isdir(options[redirection_path_key]):
-                    stdout = os.path.join(options[redirection_path_key], output_file_name)
-                    stderr = os.path.join(options[redirection_path_key], error_file_name)
+            if redirection_path_key != "":
+                if not os.path.isdir(redirection_path_key):
+                    raise Exception(f"The provided directory {redirection_path_key} does not exist in your system!")
                 else:
-                    raise Exception(f"The provided directory {options[redirection_path_key]} does not exist in your system!")
+                    stdout = os.path.join(redirection_path_key, output_file_name)
+                    stderr = os.path.join(redirection_path_key, error_file_name)
             else:
                 target_path = FrontendConfiguration.get_target_path()
                 stdout = os.path.join(target_path, output_file_name)
@@ -95,9 +92,8 @@ class Builder(WithOptions, metaclass=ABCMeta):
     def build(self) -> None:
         pass
 
-    @abstractmethod
     def get_builder_name(self) -> str:
-        pass
+        return self.__class__.__name__
 
     def set_options(self, options: Mapping[str, Any]) -> Mapping[str, Any]:
         ret = super().set_options(options)
--- a/pynestml/codegeneration/nest_builder.py
+++ b/pynestml/codegeneration/nest_builder.py
@@ -105,6 +105,11 @@ class NESTBuilder(Builder):
         InvalidPathException
             If a failure occurs while trying to access the target path or the NEST installation path.
         """
+
+        stdout = self._options["stdout"]
+        stderr = self._options["stderr"]
+        error_location = self._options["error_location"]
+
         cmake_cmd = ["cmake"]
         target_path = FrontendConfiguration.get_target_path()
         install_path = FrontendConfiguration.get_install_path()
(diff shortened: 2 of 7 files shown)


pynestml/codegeneration/nest_builder.py
```diff
@@ -85,14 +85,15 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
...
+                Logger.log_message(
+                    None, -1, "An error occurred while importing the `nest` module in Python. Please check your NEST installation-related environment variables and paths.", None, LoggingLevel.ERROR)
                 sys.exit(1)
 
             nest_path = nest.ll_api.sli_func("statusdict/prefix ::")
             self.set_options({"nest_path": nest_path})
             Logger.log_message(None, -1, "The NEST installation was automatically detected as: " + nest_path, None, LoggingLevel.INFO)
 
-    def build(self) -> None:
+    def build(self, stdout: TextIO = None, stderr: TextIO = None) -> None:
```
clinssen, line 96: Could you add docstrings for these new parameters and perhaps a little instruction on how they can be used?

clinssen (changes requested):

pynestml/codegeneration/nest_builder.py
```diff
@@ -103,6 +104,13 @@ def build(self) -> None:
         InvalidPathException
             If a failure occurs while trying to access the target path or the NEST installation path.
         """
+        error_location = ""
+        if stderr is None:
+            stderr = subprocess.STDOUT
+            error_location = "stdout"
+        else:
+            error_location = os.path.abspath(stderr.name)
```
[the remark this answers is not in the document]
med-ayssar: You mean to have a default value in the function deflation instead?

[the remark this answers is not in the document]
pnbabu: No, I meant the variable `error_location`.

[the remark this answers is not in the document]
med-ayssar: It is used to show the user where to find the error, in case there is one. So either it will be `stdout` or the file name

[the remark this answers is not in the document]
pnbabu: I agree, but it is assigned and not used anywhere.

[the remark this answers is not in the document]
med-ayssar: Oh! sorry, yeah it is missing and must be used when showing the error. Thank you!

pynestml/codegeneration/nest_builder.py
```diff
@@ -85,14 +85,15 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
             try:
                 import nest
             except ModuleNotFoundError:
-                Logger.log_message(None, -1, "An error occurred while importing the `nest` module in Python. Please check your NEST installation-related environment variables and paths.", None, LoggingLevel.ERROR)
+                Logger.log_message(
+                    None, -1, "An error occurred while importing the `nest` module in Python. Please check your NEST installation-related environment variables and paths.", None, LoggingLevel.ERROR)
```
jougs, line 89:
```suggestion
                msg = "An error occurred while importing the 'nest' module in Python. Please check your NEST installation-related environment variables and paths."
                Logger.log_message(None, -1, msg, None, LoggingLevel.ERROR)
```

pynestml/codegeneration/nest_builder.py
```diff
@@ -140,21 +148,24 @@ def build(self) -> None:
...
 
         # finally execute make install
         try:
-            subprocess.check_call(make_install_cmd, stderr=subprocess.STDOUT, shell=shell,
+            subprocess.check_call(make_install_cmd, stderr=stderr, shell=shell, stdout=stdout,
                                   cwd=str(os.path.join(target_path)))
         except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
+            raise GeneratedCodeBuildException(
+                'Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
```
jougs, line 171:
```suggestion
            msg = "Error occurred during 'make install'! More detailed error messages can be found in stdout."
            raise GeneratedCodeBuildException(msg)
```

Yes, I hate line breaks!

pynestml/codegeneration/nest_builder.py
```diff
@@ -140,21 +148,24 @@ def build(self) -> None:
...
 
         # now execute make all
         try:
-            subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
+            subprocess.check_call(make_all_cmd, stderr=stderr, shell=shell, stdout=stdout,
                                   cwd=str(os.path.join(target_path)))
         except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
+            raise GeneratedCodeBuildException(
+                'Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
```
jougs, line 163:
```suggestion
            msg = "Error occurred during 'make all'! More detailed error messages can be found in stdout."
            raise GeneratedCodeBuildException(msg)
```

pynestml/codegeneration/nest_builder.py
```diff
@@ -140,21 +148,24 @@ def build(self) -> None:
 
         # first call cmake with all the arguments
         try:
-            result = subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT, shell=shell,
+            result = subprocess.check_call(cmake_cmd, stderr=stderr, shell=shell, stdout=stdout,
                                            cwd=str(os.path.join(target_path)))
         except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
+            raise GeneratedCodeBuildException(
+                'Error occurred during \'cmake\'! More detailed error messages can be found in stdout.')
```
jougs, line 155:
```suggestion
            msg = "Error occurred during 'cmake'! More detailed error messages can be found in stdout."
            raise GeneratedCodeBuildException(msg)
```

jougs (changes requested):

pynestml/codegeneration/nest_builder.py
```diff
@@ -141,21 +161,25 @@ def build(self) -> None:
 
         # first call cmake with all the arguments
         try:
-            subprocess.check_call(cmake_cmd, stderr=subprocess.STDOUT, shell=shell,
-                                  cwd=str(os.path.join(target_path)))
+            result = subprocess.check_call(cmake_cmd, stderr=error_file, shell=shell, stdout=output_file,
```
clinssen, line 164: ``result`` is anyway ignored, right?
```suggestion
            subprocess.check_call(cmake_cmd, stderr=error_file, shell=shell, stdout=output_file,
```

pynestml/codegeneration/nest_builder.py
```diff
@@ -86,24 +86,44 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
...
                 sys.exit(1)
 
             nest_path = nest.ll_api.sli_func("statusdict/prefix ::")
             self.set_options({"nest_path": nest_path})
-            Logger.log_message(None, -1, "The NEST installation was automatically detected as: " + nest_path, None, LoggingLevel.INFO)
+            msg = "The NEST installation was automatically detected as: " + nest_path
+            Logger.log_message(None, -1, msg, None, LoggingLevel.INFO)
 
-    def build(self) -> None:
+    def build(self, output_file: TextIO = None, error_file: TextIO = None) -> None:
```
clinssen, line 98: Types should be ``Optional[TextIO]`` to allow for ``None`` value.

pynestml/codegeneration/nest_builder.py
```diff
@@ -86,24 +86,44 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
...
+
+
         Raises
         ------
         GeneratedCodeBuildException
             If any kind of failure occurs during cmake configuration, build, or install.
         InvalidPathException
             If a failure occurs while trying to access the target path or the NEST installation path.
         """
+        error_location = ""
```
clinssen, line 120:
```suggestion
```

pynestml/codegeneration/nest_builder.py
```diff
@@ -86,24 +86,44 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
...
-    def build(self) -> None:
+    def build(self, output_file: TextIO = None, error_file: TextIO = None) -> None:
         r"""
         This method can be used to build the generated code and install the resulting extension module into NEST.
 
+        Parameters
+        ----------
+        output_file
+            A file object containing the summary of the 'cmake' executed command.
+            By default the output_file will be set to the stderr output
```
clinssen, line 106: Defaults are the other way around (stderr is redirected to stdout), no?

pynestml/codegeneration/nest_builder.py
```diff
@@ -86,24 +86,44 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
...
         Raises
         ------
         GeneratedCodeBuildException
             If any kind of failure occurs during cmake configuration, build, or install.
         InvalidPathException
             If a failure occurs while trying to access the target path or the NEST installation path.
         """
+        error_location = ""
+        if error_file is None:
+            error_file = subprocess.STDOUT
```
clinssen, line 122: Any particular reason for redirecting stderr to stdout by default?

clinssen:
@med-ayssar : gentle ping! Could you also check if this means that when running NESTML from JupyterLab or a Jupyter notebook, we can capture stdout/stderr and not overflow the screen with a lot of compiler/linker info and warning messages?

pynestml/codegeneration/nest_builder.py
med-ayssar: If you take a look at the original code before introducing this modification in the `subprocess.check_call` function call, you will see that the ´stderr´ argument in the function was set to the `STDOUT`. Since I am also printing everything to a file, In line 121, I am just checking either to print everything in the STDOUT or in the file. Correct me if wrong.

pynestml/codegeneration/nest_builder.py
```diff
@@ -142,21 +165,36 @@ def build(self) -> None:
...
 
         # now execute make all
         try:
-            subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
+            subprocess.check_call(make_all_cmd, stderr=error_file, shell=shell, stdout=output_file,
                                   cwd=str(os.path.join(target_path)))
         except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
+            self.__close_files(output_file, error_file)
+            msg = f"Error occurred during 'make all. More detailed error messages can be found in {error_location}."
```
pnbabu, line 182:
```suggestion
            msg = f"Error occurred during 'make all'. More detailed error messages can be found in {error_location}."
```

pynestml/codegeneration/nest_builder.py
```diff
@@ -87,24 +88,46 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
             try:
                 import nest
             except ModuleNotFoundError:
-                Logger.log_message(None, -1, "An error occurred while importing the `nest` module in Python. Please check your NEST installation-related environment variables and paths, or specify ``nest_path`` manually in the code generator options.", None, LoggingLevel.ERROR)
+
+                msg = "An error occurred while importing the 'nest' module in Python. Please check your NEST installation-related environment variables and paths."
```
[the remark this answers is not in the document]
med-ayssar: Sure!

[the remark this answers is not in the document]
med-ayssar: Sure!

pynestml/frontend/pynestml_frontend.py
```diff
@@ -313,7 +315,16 @@ def process():
 
     # perform build
     if not errors_occurred and _builder is not None:
-        _builder.build()
+        target_path = FrontendConfiguration.get_target_path()
+        stdout_path = os.path.join(target_path, "cmake_output.txt")
```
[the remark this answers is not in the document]
med-ayssar: I extended the `frontend_configuration` and `pynestml_frontend` with a new argument to support redirecting CMake output. The default behavior will be simply printing out the CMake output in the terminal, otherwise when the flag (argument) is provided, the CMake output will be redirected to a file.

doc/running.rst
```diff
@@ -55,6 +55,9 @@ The following default values are used, corresponding to the command line default
    * - codegen_opts
      - Optional[Mapping[str, Any]]
      - (Optional) A JSON equivalent Python dictionary containing additional options for the target platform code generator. These options are specific to a given target platform, see for example :ref:`Running NESTML with custom templates`.
+   * - redirect_cmake_output
```
pnbabu, line 58: I think it is better to have this in `codegen_opts` as this is specific to NEST target

pynestml/frontend/pynestml_frontend.py
```diff
@@ -313,7 +318,20 @@ def process():
...
-        _builder.build()
+
+        if FrontendConfiguration.get_redirect_cmake_output():
+            target_path = FrontendConfiguration.get_target_path()
+            stdout_path = os.path.join(target_path, "cmake_output.txt")
+            stderr_path = os.path.join(target_path, "cmake_error.txt")
+            stdout = open(stdout_path, "w")
+            stderr = open(stderr_path, "w")
+
+            _builder.build(stdout, stderr)
```
clinssen, line 329: What happens in case of failure? Use ``try..finally`` or ``with``.

clinssen:
@med-ayssar: could you also pull master? Cheers!

med-ayssar:
The `redirection` is now part of all `Builders`. 
The `redirection` is either set to `true` or `false`. When set to `false` (default), the Cmake output (including the errors) will be printed to the user in the terminal or in `jupyter`, and when set to `true`, the output will be redirected to two different files (one for the normal output, and the other for the errors). The generated files will be created under the `target_path` directory

pynestml/utils/with_options.py
```diff
@@ -20,6 +20,8 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 from typing import Any, Mapping, Optional
+import subprocess
```
pnbabu, line 23: Remove this

pynestml/frontend/pynestml_frontend.py
```diff
@@ -107,14 +108,18 @@ def builder_from_target_name(target_name: str, options: Optional[Mapping[str, An
 
     if target_name.upper() == "NEST":
         from pynestml.codegeneration.nest_builder import NESTBuilder
-        return NESTBuilder(options)
+        if options.pop("redirect_build_output", None):
+            nest_builder = NESTBuilder(options)
+            nest_builder.redirect_build_output(FrontendConfiguration.get_target_path())
```
pnbabu, line 113: I think it would be simpler to have the `redirect_build_output` option as part of `codegen_opts`. The `if` condition would be avoided and the `options` would be parsed in the `Builder` class. See the option `nest_path` for example.

pynestml/codegeneration/builder.py
```diff
@@ -44,8 +57,29 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
             self._target = ""
             raise InvalidTargetException()
 
+        self._redirect_output = copy.deepcopy(self.__class__._redirect_build_output)
         self._target = target
 
-    @abstractmethod
     def build(self) -> None:
+        if (self._redirect_output["default"]):
```
pnbabu, line 64: Could this be moved to the `__init__` function instead? Then there is no need for the `main()` function.

pynestml/codegeneration/builder.py
```diff
@@ -34,7 +39,15 @@
...
     r"""Compile, build and install the code for a given target platform. Runs after the CodeGenerator."""
 
-    def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
+    _redirect_build_output = {
+        "error_location": "stderr",
+        "default": True,
+        "stdout": None,
+        "stderr": subprocess.STDOUT
+
+    }
```
clinssen, line 48: As Builder inherits from WithOptions, can you just use the get and set-option methods that already exists? The defaults in this ``_redirect_build_output`` dict should be moved to the ``_default_options``.

pynestml/codegeneration/builder.py
clinssen: It should probably stay in ``build()`` so it redirects only stdout and stderr output that belongs strictly to the build stage.

``main()`` is very nondescriptive, how about ``build_()`` to indicate "private" method?

pynestml/codegeneration/builder.py
```diff
@@ -44,8 +57,29 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
             raise InvalidTargetException()
 
+        self._redirect_output = copy.deepcopy(self.__class__._redirect_build_output)
         self._target = target
 
-    @abstractmethod
     def build(self) -> None:
+        if (self._redirect_output["default"]):
+            self.main()
+        else:
```
clinssen, line 66: Can we change the if-then-else logic here so ``build_()`` is unconditionally called and we can get rid of the "else" entirely?

pynestml/codegeneration/nest_builder.py
```diff
@@ -148,21 +169,36 @@ def build(self) -> None:
...
-            subprocess.check_call(make_install_cmd, stderr=subprocess.STDOUT, shell=shell,
+            subprocess.check_call(make_install_cmd, stderr=stderr, shell=shell, stdout=stdout,
                                   cwd=str(os.path.join(target_path)))
         except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'make install\'! More detailed error messages can be found in stdout.')
+            self.__close_files(stdout, stderr)
+            msg = f"Error occurred during 'make install'. More detailed error messages can be found in {error_location}.'"
+            raise GeneratedCodeBuildException(msg)
+
+    def __close_files(self, output, error):
```
clinssen, line 198: Would it be possible to move this into the base class Builder?

clinssen (changes requested):
Do you think it would be the most logical to have a filename where stdout would be written to and a filename where stderr would be written to as part of the command-line parameters to NESTML instead? Or we could leave these to the options (due to WithOptions) so they can be specified in the codegen_opts JSON file (or dict). Then we wouldn't even have to add any command-line arguments at all. The documentation could go right after this section: https://nestml.readthedocs.io/en/latest/running.html#custom-templates (see also https://nestml.readthedocs.io/en/latest/extending.html#running-nestml-with-custom-templates).

Also please add logging INFO-level comments about the redirections including the file names.

med-ayssar:
I am against the concept of using the `WithOptions::set_options` that returns a dictionary with the remaining unused data.
I think that each Builder/Generator should use the `codegen_opt` to initiate it its own state. And before executing the main routine of the Builder/Generator, we can simply give a warning to the user of the extra keys, like the way we have in CMake with nest.

med-ayssar:
### Print to the console
```python
generate_nest_target(input_path="prefix/models/izhikevich.nestml",
                     target_path="/tmp/nestml_console",
                     logging_level="ERROR",
                     module_name="test_module",
                     codegen_opts={"nest_path": NEST_SIMULATOR_INSTALL_LOCATION)
```

### Forward to `target_dir` (Default)
```python
generate_nest_target(input_path="prefix/models/izhikevich.nestml",
                     target_path="/tmp/nestml_default",
                     logging_level="ERROR",
                     module_name="test_module",
                     codegen_opts={"nest_path": NEST_SIMULATOR_INSTALL_LOCATION, "redirect_build_output": True})
```

### Forward to any EXISTING directory
```python
generate_nest_target(input_path="prefix/models/izhikevich.nestml",
                     target_path="/tmp/nestml_default",
                     logging_level="ERROR",
                     module_name="test_module",
                     codegen_opts={"nest_path": NEST_SIMULATOR_INSTALL_LOCATION, "redirect_build_output": True, 
                     "build_output_dir": os.getcwd()})
```

pynestml/frontend/pynestml_frontend.py
```diff
@@ -313,7 +315,7 @@ def process():
                                                                            options=codegen_and_builder_opts)
     _codeGenerator = code_generator_from_target_name(FrontendConfiguration.get_target_platform())
     codegen_and_builder_opts = _codeGenerator.set_options(codegen_and_builder_opts)
-    _builder = builder_from_target_name(FrontendConfiguration.get_target_platform())
+    _builder = builder_from_target_name(FrontendConfiguration.get_target_platform(), options=codegen_and_builder_opts)
```
pnbabu, line 318: Do the options need to be set here since it's again set on line 321?

pynestml/codegeneration/builder.py
```diff
@@ -46,6 +53,54 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
+
+            error_location = stderr
+            stdout = open(stdout, "w")
+            stderr = open(stderr, "w")
+            redirect = True
+
+        self.add_options({"stdout": stdout, "stderr": stderr, "redirect": redirect, "error_location": error_location})
+
     @abstractmethod
     def build(self) -> None:
```
clinssen, line 95: How about we rename this to ``build_()`` everywhere, and implement a new, non-abstract method ``build()`` in the base class that handles opening and closing the files safely and calls ``build_()`` from within the right context? Something like:
```py
def build():
    stdout, stderr = process_output_redirection_(self.options)
    with open(..., 'w') as stdout, open(..., 'w') as stderr:
        self._build(stdout, stderr)
```
Do you think it makes sense for ``_build()`` to accept the stdout, stderr parameters (in all ``Builder``s)? This means that we would be more or less explicitly requiring all builders to implement this functionality (of redirecting stdout and stderr in their subprocess or other calls).

pynestml/frontend/pynestml_frontend.py
```diff
@@ -65,7 +65,8 @@ def transformers_from_target_name(target_name: str, options: Optional[Mapping[st
...
+        variable_name_rewriter = IllegalVariableNameTransformer({"forbidden_names": ["alignas", "alignof", "and", "and_eq", "asm", "auto", "bitand", "bitor", "bool", "break", "case", "catch", "char", "char8_t", "char16_t", "char32_t", "class", "compl", "concept", "const", "const_cast", "consteval", "constexpr", "constinit", "continue", "co_await", "co_return", "co_yield", "decltype", "default", "delete", "do", "double", "dynamic_cast", "else", "enum", "explicit", "export", "extern", "false", "float", "for", "friend",
```
clinssen, line 68: Could you please re-run pycodestyle with the options as in https://github.com/nest/nestml/blob/master/.github/workflows/nestml-build.yml#L48? Inserting a line break here (and also in some other files in the PR) is not necessary.

pynestml/codegeneration/builder.py
```diff
@@ -46,6 +53,54 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
+                target_path = FrontendConfiguration.get_target_path()
+                stdout = os.path.join(target_path, output_file_name)
+                stderr = os.path.join(target_path, error_file_name)
+
+            error_location = stderr
+            stdout = open(stdout, "w")
+            stderr = open(stderr, "w")
+            redirect = True
+
+        self.add_options({"stdout": stdout, "stderr": stderr, "redirect": redirect, "error_location": error_location})
```
clinssen, line 92: Could you assert (here or in ``add_options()``) that the keys are not yet in the options dict?

pynestml/codegeneration/builder.py
```diff
@@ -46,6 +50,52 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
+                    stdout = os.path.join(redirection_path_key, output_file_name)
+                    stderr = os.path.join(redirection_path_key, error_file_name)
+            else:
+                target_path = FrontendConfiguration.get_target_path()
+                stdout = os.path.join(target_path, output_file_name)
+                stderr = os.path.join(target_path, error_file_name)
+
+            error_location = stderr
+            stdout = open(stdout, "w")
+            stderr = open(stderr, "w")
```
clinssen, line 85: If something goes wrong opening stderr, is stdout safely closed?

pynestml/codegeneration/builder.py
```diff
@@ -46,6 +50,52 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
+        redirection_path_key = options.get("build_output_dir", "")
+
+        # default values. The output will be printed to the console.
+        stdout = None
+        stderr = subprocess.STDOUT
+        redirect = False
+        error_location = "stderr"
+
+        if options and len(options) > 0 and require_redirect_key:
+            output_file_name = f"{self.get_builder_name()}_output.txt"
```
clinssen, line 69:
```suggestion
            output_file_name = f"{self.get_builder_name()}_stdout.txt"
```

pynestml/codegeneration/builder.py
```diff
@@ -46,6 +50,52 @@ def __init__(self, target, options: Optional[Mapping[str, Any]]=None):
...
+            error_file_name = f"{self.get_builder_name()}_error.txt"
```
clinssen, line 70:
```suggestion
            error_file_name = f"{self.get_builder_name()}_stderr.txt"
```

med-ayssar:
Now the files are opened locally inside the `build` function, and I merged the exceptions to each of the `CMake` stages in a single thrown exception using a dictionary depending on the current stage.

pynestml/codegeneration/nest_builder.py
```diff
@@ -151,24 +148,35 @@ def build(self) -> None:
...
+                    subprocess.check_call(cmake_cmd, stderr=stderr, stdout=stdout, shell=shell,
+                                          cwd=str(os.path.join(target_path)))
+                    current_stage = "build"
+                    subprocess.check_call(make_all_cmd, stderr=stderr, stdout=stdout, shell=shell,
+                                          cwd=str(os.path.join(target_path)))
+                    current_stage = "install"
+                    subprocess.check_call(make_install_cmd, stderr=stderr, stdout=stdout, shell=shell,
+                                          cwd=str(os.path.join(target_path)))
+
+            else:
```
clinssen, line 167: Lot of code duplication between the if and the else parts. I would be happier with the old structure here. If no forwarding is requested, ``stdout`` and ``stderr`` could default to None, so the calls to ``subprocess.check_call()`` do not have to be repeated.

clinssen (changes requested):
Thanks, I think this is almost there! Could you also add a few notes in the documentation and resolve the last open comments from the earlier review?

pynestml/codegeneration/nest_builder.py
med-ayssar: I have removed the duplicated lines, and now we check if we should open the new files or not, and then execute the same block for both cases.

At the end, we close the files if they were opened already, either after finishing the CMake stages or after an exception.

med-ayssar:
Putting the process of opening two files in the try Block will ensure that either are opened or the execution of the process will be stopped. So I think it is safe to use it as it is right now.

pynestml/codegeneration/nest_builder.py
```diff
@@ -146,23 +148,36 @@ def build(self) -> None:
...
-        try:
-            subprocess.check_call(make_all_cmd, stderr=subprocess.STDOUT, shell=shell,
-                                  cwd=str(os.path.join(target_path)))
-        except subprocess.CalledProcessError as e:
-            raise GeneratedCodeBuildException('Error occurred during \'make all\'! More detailed error messages can be found in stdout.')
+        stdout = self._options["stdout"]
+        stderr = self._options["stderr"]
+
+        if self._options["redirect"]:
+            try:
```
clinssen, line 159: It's probably not anything too catastrophic, but putting two things that can fail independently inside the same ``try`` block might not be best practice. What if stdout is opened successfully but stderr fails to open with an unhandeled exception, and we drop into the iPython prompt with a still-open stdout stream? To avoid having to consider all these edge cases, and because the opening and closing of the streams is now anyway very close to where the streams are being used, can we just use a context manager here? See my earlier comment (https://github.com/nest/nestml/pull/741#discussion_r1101229678).

med-ayssar: Sorry, I just made that comment, when I was in the train, and it seems that I have not given it enough thoughts.

I think the best, as you said, to keep opening the file locally close to where they are needed. Therefore, I have split the creation of the files into two different `try-catch` blocks.

clinssen (approved):

PR #743, opened 2022-03-15
Fix `clone` function in `ASTForStmt`

### LINKED ISSUE #742: `Clone` function in `ASTForStmt` isn't implemented properly
The class `ASTForStmt`  has two non-cloneable fields ((`ìnt`, `float`) and `str` don't implement the clone function). Given then that the nestml file has a `for-loop`, cloning the `AstNeuron` (in general the `ASTNode`) causes an exception to be thrown.

med-ayssar:
Fixes #742

### DIFF
--- a/pynestml/meta_model/ast_for_stmt.py
+++ b/pynestml/meta_model/ast_for_stmt.py
@@ -69,7 +69,7 @@ class ASTForStmt(ASTNode):
         """
         variable_dup = None
         if self.variable:
-            variable_dup = self.variable.clone()
+            variable_dup = str(self.variable)
         start_from_dup = None
         if self.start_from:
             start_from_dup = self.start_from.clone()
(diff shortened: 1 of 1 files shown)


pynestml/meta_model/ast_for_stmt.py
```diff
@@ -69,7 +69,7 @@ def clone(self):
         """
         variable_dup = None
         if self.variable:
-            variable_dup = self.variable.clone()
+            variable_dup = str(self.variable)
```
clinssen, line 72: Can we just assign ``variable_dup = self.variable``? Strings are immutable in Python. Same goes for the float and int below.

med-ayssar:
@clinssen, aside note, can you check the pipeline, it is failing installing some dependencies.

pynestml/meta_model/ast_for_stmt.py
med-ayssar: Sure!

clinssen (approved):
Many thanks!

PR #745, opened 2022-03-15
Remove unnecessary log messages

clinssen:
The "ODE has been updated" message was pointless and has been removed.

The "Start updating symbol table" message has been downgraded to DEBUG level, but I'm happy to remove it as well.


clinssen:
Possibly, we should also remove "The model <xyz> will be analysed!" It is only emitted by an internal method in the NEST code generator. This is redundant with "Starts processing of the model <xyz>", which is always printed right before.

pnbabu (approved):
Thanks!

PR #748, opened 2022-03-18
Run multithreading test for various number of threads

clinssen:


tests/nest_tests/nest_multithreading_test.py
```diff
@@ -132,6 +139,10 @@ def test_neuron_synapse_multithreading(self, number_of_threads: int) -> None:
...
-        V_m = nest.GetStatus(mm, "events")[0]["V_m"]
-        print(V_m)
-        np.testing.assert_almost_equal(V_m[-4], -59.17946541)
+        connections = nest.GetConnections(synapse_model="stdp_nestml_rec")
+        gid_post = np.unique(np.array(connections.get("target")))[0]
+        events = mm.get("events")
+        senders = events["senders"]
+        V_m = events["V_m"]
+        V_m_sender = V_m[senders == gid_post]
+        np.testing.assert_almost_equal(V_m_sender[-4], -59.17946541)
```
clinssen, line 148: Actually, where does this -4 come from? Can it also be just -1 (with a slightly updated numeric value)?

pnbabu: Yes, it can be. The value at -1 was -70.0 which is the reset value of the membrane potential. Thus I chose a value that is not equal to the reset value.

clinssen:
Many thanks!

PR #756, opened 2022-04-07
Use @NEURON_NAME@ escape sequence in template filenames

clinssen:
This is necessary because the compartmental platform target needs to render several templates per model, e.g. "cm_main_@NEURON_NAME@.cpp", "cm_tree_@NEURON_NAME@.cpp", etc.


pnbabu (approved):
Thanks, looks good to me.

clinssen:
Cheers!

PR #759, opened 2022-04-13
Fix bug in vectors when there is a single vector state variable

### LINKED ISSUE #758: Uncompleted code generation when using `array` in nestml
The following problem happens only when using `array` state attributes in nestml.

- Assuming we have this simple nestml code:
```
neuron simple_model:
  state:
   g_ex [20] real = 1.2
  end
end
```
the resulted code has a small issue in the header file and precisely in the `get_state_element` function:

```
inline double get_state_element(size_t elem)
  {
    if
    {
      return S_.g_ex[ elem - State_::G_EX];
    }
  }
```
The condition seems to be missing in the generated code.

pnbabu:
Fixes #758


clinssen:
This condition is a bit hard to reverse engineer, could you make the first term just a check for if the length is one?

pnbabu:
@clinssen I have updated the condition. Could you check if it is better now?

clinssen:
OK, looks good! Just one question, could you add @med-ayssar's example model in #758 as a unit test, just to generate and build the code?

PR #760, opened 2022-04-13
Fix CI failure

pnbabu:
Update Antrl4 version to the latest in order to avoid CI failure due to the upstream changes in antlr4-python package


clinssen:
This is great, thank you very much!

PR #766, opened 2022-04-24
Fix upper case state variable name naming clash

### LINKED ISSUE #765: Can't have variable with upcased name when having arrays
It seems when we have arrays and variables with `upcased` names, `nestml` generates the code without any problem, but the code won't compile for the following reasons, but first let's demonstrate the problem.
Again, we start with a simple model that should cause the problem:
```
neuron names_conflicts:
    state:
      Y_1 real = 42.
      x[N] real = 1.
    end
   
    parameters:
      N integer = 1  # array size
    end   
  end
```
Here, we have an array `x` and a double variable named `Y_1`, and the generated `C++` block for the state struct is the following:
```
 struct State_ {
    enum StateVecVars {
      Y_1 = 0,
      X = 1,
    };
    double Y_1;
    std::vector<double> x;

    State_();
  };
```
`Y_1` was declared **twice**, first as part of `enum` and then as double in the same `State_` struct, thus the compilation of the code won't be successful, and the compiler throws the following error: ` double names_conflicts::State_::Y_1 conflicts with a previous declaration´.

clinssen:
Fixes #765.

### DIFF
--- a/pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
+++ b/pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
@@ -87,7 +87,7 @@ template <> void RecordablesMap<{{neuronName}}>::create()
 {%- if has_state_vectors %}
 {%-   for sym in recordable_state_variables %}
 {%-     if not sym.has_vector_parameter() %}
-    insert("{{sym.get_symbol_name()}}", host.get_data_access_functor( {{neuronName}}::State_::{{names.name(sym).upper()}} ));
+    insert("{{sym.get_symbol_name()}}", host.get_data_access_functor( {{neuronName}}::State_::STATE_VEC_VAR_{{names.name(sym).upper()}} ));
 {%-     endif %}
 {%-   endfor %}
 {%- else %}
@@ -127,7 +127,7 @@ template <> void RecordablesMap<{{neuronName}}>::create()
 {%- for variable in neuron.get_vector_state_symbols() %}
       for (size_t i = 0; i < {{printer.print_vector_size_parameter(variable)}}; i++)
       {
-        size_t elem = {{neuronName}}::State_::{{names.name(variable).upper()}} + i;
+        size_t elem = {{neuronName}}::State_::STATE_VEC_VAR_{{names.name(variable).upper()}} + i;
         recordablesMap_.insert(get_var_name(i, "{{names.name(variable).upper()}}_"), this->get_data_access_functor(elem));
       }
 {%- endfor %}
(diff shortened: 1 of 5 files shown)


pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -127,7 +127,7 @@ template <> void RecordablesMap<{{neuronName}}>::create()
 {%- for variable in neuron.get_vector_state_symbols() %}
       for (size_t i = 0; i < {{printer.print_vector_size_parameter(variable)}}; i++)
       {
-        size_t elem = {{neuronName}}::State_::{{names.name(variable).upper()}} + i;
+        size_t elem = {{neuronName}}::State_::STATE_VEC_VAR_{{names.name(variable).upper()}} + i;
         recordablesMap_.insert(get_var_name(i, "{{names.name(variable).upper()}}_"), this->get_data_access_functor(elem));
```
med-ayssar, line 131: Thanks for the changes. The code will compile, but we still have an unchecked runtime problem.  The function `insert_recordables` at line 131 inserts in the map as key the array's name in an upper case form and adds a `_{i}` as the entry name in the recordables. So if we have an array `x`  with `n` elements, we will insert `X_{i}` for ` i in [1, n]`. With this in mind, the user might then want to create two variables in his nestml file, one array variable named `x` and another double variable named `X_1`. What will happen when inserting the recordables is that the double variables `X_1` will be inserted first, and then the array entries, but the map already contains `X_1`, and therefore the first entry of the array will never be recorded.

med-ayssar: I was thinking about that, and I think the best way is to override the `find` function in the generic `RecordableMap` as the `create` function. Instead of inserting each entry of the array, we should only register a constant pointer to the array and only store its name as `x`, and then we have to implement a logic that can resolve the global name to a local name (mapping `x` to `x_i`). 
Assume we have the following map : map : { "x" : [x1, x2,...., xn]}, and let's say we want to deliver `x_i` to the `DataLoggingRequest`. Calling map.find("x_i"), should return a pair containing `x` as the name and a `callable` that can access the array at position `i`.

With this, we can also save us runtime overhead. Assume that we have an array with size `n >M` (big number), and we are only interested in the result of the last entry to be recorded, but at the same time that entry depends on the previous entries. Firstly, the constructor will be very expensive as it must iterate over the entries and register them in the map, but from the python side, the user will only want to record the last entry. With the new suggested changes, the entries will be registered all together at once by having the whole array representing them, and then when setting the `record_from` in `pynest` to `x_i`, the find function should parse the input and split it to `(x, i)` and then returns an access function to the array at position `i`.

clinssen: We could keep it simple and add a check that if a vector variable ``foo`` exists, there exists no declaration in the model of a variable named ``FOO``.

med-ayssar: hmm, but u must prevent then not just `FOO` but also every `FOO_i` for `i < len(foo)`

clinssen:
Perhaps we could eliminate the use of ``.upper()`` to uppercase variable names in the generated code (and also move to C++ enum class for ``StateVecElems`` to ensure proper scoping of the enum members).

PR #771, opened 2022-05-04
Compile NESTML generated code multithreaded

clinssen:


pynestml/codegeneration/nest_builder.py
```diff
@@ -123,9 +123,11 @@ def build(self) -> None:
                 install_path = os.path.abspath(install_path)
             install_prefix = f"-DCMAKE_INSTALL_PREFIX={install_path}"
 
+        n_cpu = len(os.sched_getaffinity(0))  # compile multithreaded
```
pnbabu, line 126: This function is unfortunately not present for Mac OS :(
Could we use `os.cpu_count()` instead?

clinssen: ``sched_getaffinity()`` has some advantages on *nix, so I'd like to keep it in. Do you get an exception when running this under the Mac, or just a return of ``None``? We would then fall back to ``os.cpu_count()``.

pnbabu: I get this exception: `AttributeError: module 'os' has no attribute 'sched_getaffinity'`. Probably, we can add different functions depending on the platform?

jougs (approved):
Just a minor comment for readability.

PR #735, opened 2022-03-09
Add support for compartmental models

pythonjam:
That's the most recent branch now

PR #772, opened 2022-05-05
Compartmental models in NESTML

WillemWybo:
Support for customizable compartmental models in NESTML. Supersedes PR #735


WillemWybo:
@clinssen I've corrected the code-style issues in the nest-3 codecheck, and it doesn't show any errors anymore, yet the code check still fails on the static code analysis. From my side, the `compartmental_model_test.py` runs, so in principle we should be good to merge.

PR #773, opened 2022-05-05
Fix vector resize

### LINKED ISSUE #747: Avoiding  `undefined behavior` when using arrays in `nestml`
When using arrays in nestml, it seems that the generated code has missing logic checking the integrity between the array size and the assigned array.

Given the following simple nestml model:
``` 
neuron array_size_problem:

    state:
      # acumulator for x values
      y real = 0.
   
      x[N] real = 1.
    end
   
    parameters:
      N integer = 1  # array size
    end
   
    update:
 
      j integer = 0
      y = 0
      for j in 0 ... N step 1:
        y += x[j]
      end
      print ("y= {y}\n")
    
    end
   
  end
   ```
and the corresponding  generated `set_status` function:
```
inline void array_size_problem::set_status(const DictionaryDatum &__d)
{
  // parameters
  long tmp_N = get_N();
  updateValue<long>(__d, nest::array_size_problem_names::_N, tmp_N);
  // Resize vectors
  if (tmp_N != get_N())
  {
    std::vector< double >  _tmp_x = get_x();
    _tmp_x.resize(tmp_N, 0.);
    set_x(_tmp_x);
  }

  // initial values for state variables in ODE or kernel
  double tmp_y = get_y();
  updateValue<double>(__d, nest::array_size_problem_names::_y, tmp_y);
  std::vector< double >  tmp_x = get_x();
  updateValue<std::vector< double > >(__d, nest::array_size_problem_names::_x, tmp_x);

  // We now know that (ptmp, stmp) are consistent. We do not
  // write them back to (P_, S_) before we are also sure that
  // the properties to be set in the parent class are internally
  // consistent.
  ArchivingNode::set_status(__d);

  // if we get here, temporaries contain consistent set of properties
  set_N(tmp_N);
  set_y(tmp_y);
  set_x(tmp_x);


};
```
We observe that firstly the size of the array is being checked, and in case if there is a new given size, we resize our current array. Having the array resized, we read the stored array in the dictionary and  assign it to our array with calling the `set_x(tmp_x)` method. Unfortunately, the set function doesn't check if the size of the passed array is equal to the current array size, and therefore the following would happen: The user sets `N=200` and sets `x` by calling`neuron.set(x=[1.0, 1.0])`. Now the neurons thinks it has an array of size `N`, but in reality it has an array of size two, and in the `update` function the neuron iterates `N=200` times over the array calling the`operator[]` function. Consequently, this leads to accessing a nonexistent element in the array and results in an `undefined behavior.`

Running the following the script, we expect nest to throw an exception when reaching the second line.
```
neuron = nest.Create("array_size_problem", params={"N": 256})
neuron.set(x=[1.0, 1.0])   # must throw an exception
nest.Simulate(10)
```
Depending on the set value for `N`, the outputs might be `zero`, `some_big_number` and `-nan`.

To avoid all these problems, we must use the `vector::at` function instead of the `operator[]`, and provide the `neuron::set_array` without condition to check if the passed array has the same size as the current expected size.

### PR DESCRIPTION
**Pull request by pnbabu**
Fixes #747 

This PR fixes the case when the vector is resized and assigned a vector of incorrect size. An exception is thrown now when there is a mismatch in its size.

### CODE REVIEW

File: pynestml/codegeneration/resources_nest/point_neuron/directives/ReadFromDictionaryToTmp.jinja2 [hunk 1]
```diff
@@ -19,9 +22,24 @@ if (tmp_{{names.name(variable)}} != {{names.getter(variable)}}())
...
-{%- elif not variable.is_inline_expression and variable.is_state() %}
+
+{%- else %}
 {{declarations.print_variable_type(variable)}} tmp_{{names.convert_to_cpp_name(variable.get_symbol_name())}} = {{names.getter(variable)}}();
 updateValue<{{declarations.print_variable_type(variable)}}>(__d, nest::{{names_namespace}}::_{{variable.get_symbol_name()}}, tmp_{{names.convert_to_cpp_name(variable.get_symbol_name())}});
+{%- endif %}
+
+{%- if variable.has_vector_parameter() %}
+{%- set vector_param = variable.get_vector_parameter() %}
+{%- if vector_param|int == 0 %}
```
clinssen, line 33: I'm a little confused as to what is happening here, why the special case for 0, and what happens when a parameter is a variable and cannot be cast to int? Also the if-statement on line 36 below I don't understand. Could you perhaps explain it for slow me or add some comments?

clinssen (approved):
Thank you!

PR #784, opened 2022-05-30
Make code generator throw exception on failure; fix tests

clinssen:
The way that some tests were using the code generator bypassed the frontend, which caused code generation errors not to result in a nonzero exit code. This code adds exception throwing to the ``CodeGenerator`` method ``generate_code()`` and fixes the tests that had been quietly failing.


pnbabu (approved):
Looks good to me. Please resolve conflicts before merging. Thanks!

PR #787, opened 2022-06-10
Add STDP-dopamine synapse tutorial notebook

clinssen:


pnbabu:
I was going through the notebook and found the following lines which need changes.

> As discussed in section 1, when a post-synaptic spike arrives very shortly after a pre-synaptic spike, a standard STDP rule would immediately potentiate the synaptic strength. However, as Figure 2 illustrates, when using the three-factor STDP rule, this potentiation would instead be applied to the eligibility trace.

`Section 1` and `figure 2` should be removed here.

```
BASE_DIR = os.path.join(os.path.abspath(""), os.pardir, os.pardir, os.pardir)
os.chdir(BASE_DIR)
```
Since we are adding the notebook separately to the OCNS repository for CNS tutorials, this must be removed and the models have to be downloaded separately.

models/neurons/iaf_psc_exp_alt.nestml
```diff
@@ -0,0 +1,93 @@
+"""
+iaf_psc_exp - Leaky integrate-and-fire neuron model with exponential PSCs
```
pnbabu, line 2: Is this required? Could we use the original `iaf_psc_exp` instead?

pnbabu:
@clinssen I have commented out some plots and retained only the ones that are required. Could you have a look?

clinssen:
I'd be somewhat in favour of keeping these plots in, as it improves the didactic value of the notebook: it makes it easier to see in one go what's going on in terms of the dynamics (even if all the dopamine trace plots look identical, this is easier to immediately see by glancing at the plots than to read in the text) and to be able to point to things when demonstrating in front of a classroom.

pnbabu:
I still have those plots, only that all the trace values (eligibility, weight, dopamine) per neuron are displayed together instead of displaying each trace for all neurons separately. You can see this in the notebook under `Plot the trace values`. Plotting traces separately for all the neurons are commented for now.

clinssen:
I cleaned things up a little and removed the now-unused ``plot_traces()`` function. Could you give it a final review? Cheers!

pnbabu (commented):
The changes look good to me except for the `iaf_psc_exp_alt.nestml`. Is this file required? Also, the CI is failing, although I couldn't reproduce the error locally.

clinssen:
I have removed the "_alt" model that used an absolute membrane potential, and reverted to just using the pre-existing iaf_psc_exp model with a relative membrane potential. Sorry about diffs being hard to generate/follow for ipython notebook files. Could you have another look? Cheers!

pnbabu (approved):
Looks good to me. Thanks!

PR #791, opened 2022-06-24
Add NEST code generator helper as workaround for module reloading

clinssen:
This adds a helper function that generates modules and modules with unique IDs, bypassing the issue that NEST cannot yet do module unloading/reloading.

This will be used in interactive environments: from Jupyter notebooks and NEST Desktop.

## Todo

- [x] Move function to ``NESTCodeGeneratorUtils`` class (as static method)
- [ ] <s>Use ``os.tmpfile`` to create temporary NESTML model files.</s>
- [x] Adapt all Python notebook tutorials to use this function.


pynestml/codegeneration/nest_code_generator_utils.py
```diff
@@ -19,36 +19,119 @@
...
+        -------
+        If a synapse is specified, returns a tuple (module_name, mangled_neuron_name, mangled_synapse_name) containing the names that can be used in ``nest.Install()``, ``nest.Create()`` and ``nest.Connect()`` calls. If no synapse is specified, returns a tuple (module_name, mangled_neuron_name).
+        """
+
+        # generate unique ID
+        if uniq_id is None:
+            uniq_id = str(uuid.uuid4().hex)
+
+        # read neuron model from file?
+        if not "\n" in nestml_neuron_model and ".nestml" in nestml_neuron_model:
```
pnbabu, line 84: Why is there a check for `\n` here?

clinssen: It's a heuristic check to see if we are being passed a file name, or a complete NESTML model as one big long string.

pnbabu (approved):
Looks good to me!

clinssen:
Thanks for the review!

PR #794, opened 2022-07-08
Omit `FrontendConfiguration.suffix` while extracting neuron name

### LINKED ISSUE #793: Neuron-synapse code generation fails when `suffix` option is empty in `codegen_opts`
During the co-generation of neuron and synapse models, if the `suffix` option is missing in the `generate_nest_target` function, the code generation fails with the following error:

```
>       assert len(namespace["pre_ports"]) <= 1, "Synapses only support one spiking input port"
E       AssertionError: Synapses only support one spiking input port
```
This is due to the way the `base_neuron_name` is extracted from `neuron.get_name()` in the below line:

```
base_neuron_name = neuron.get_name()[:-len(FrontendConfiguration.suffix)]
```
When the `suffix` is empty, the `len(FrontendConfiguration.suffix)` is `0` which results in `base_neuron_name` to be an empty string.

pnbabu:
Fixes #793 by not considering the `suffix` while extracting the neuron name. This doesn't affect further execution as neuron name with and without a suffix is checked in the function `is_special_port()` in `synapse_post_neuron_transformer.py`

### DIFF
--- a/pynestml/transformers/synapse_post_neuron_transformer.py
+++ b/pynestml/transformers/synapse_post_neuron_transformer.py
@@ -464,8 +464,9 @@ class SynapsePostNeuronTransformer(Transformer):
         new_synapse.paired_neuron = new_neuron
         new_neuron.paired_synapse = new_synapse
 
-        base_neuron_name = neuron.get_name()
-        base_synapse_name = synapse.get_name()
+        base_neuron_name = neuron.get_name().removesuffix(FrontendConfiguration.suffix)
+        base_synapse_name = synapse.get_name().removesuffix(FrontendConfiguration.suffix)
+
         new_synapse.post_port_names = self.get_post_port_names(synapse, base_neuron_name, base_synapse_name)
         new_synapse.spiking_post_port_names = self.get_spiking_post_port_names(synapse, base_neuron_name, base_synapse_name)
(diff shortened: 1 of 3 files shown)


pynestml/transformers/synapse_post_neuron_transformer.py
```diff
@@ -464,8 +464,8 @@ def mark_post_port(_expr=None):
         new_synapse.paired_neuron = new_neuron
         new_neuron.paired_synapse = new_synapse
 
-        base_neuron_name = neuron.get_name()[:-len(FrontendConfiguration.suffix)]
-        base_synapse_name = synapse.get_name()[:-len(FrontendConfiguration.suffix)]
+        base_neuron_name = neuron.get_name()
```
clinssen, line 467:
```suggestion
        base_neuron_name = neuron.get_name().removesuffix(FrontendConfiguration.suffix)
```

pynestml/transformers/synapse_post_neuron_transformer.py
```diff
@@ -464,8 +464,8 @@ def mark_post_port(_expr=None):
...
+        base_synapse_name = synapse.get_name()
```
clinssen, line 468:
```suggestion
        base_synapse_name = synapse.get_name().removesuffix(FrontendConfiguration.suffix)

```

clinssen (changes requested):

clinssen (approved):

PR #796, opened 2022-07-11
Update python version to 3.9

pnbabu:
With the new changes from #794, the `removesuffix()` function is available only from python version `3.9` and higher. Since we are already running the CI on python `3.9`, it is better to have the version consistent everywhere.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,6 +7,6 @@ sphinx:
     configuration: doc/sphinx-apidoc/conf.py
 
 python:
-    version: 3.8
+    version: 3.9
     install:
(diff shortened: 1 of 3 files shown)


clinssen:
@terhorstd @heplesser: is it acceptable to require Python 3.9 as a minimum version at this point in time? I ask this as it might not be available in all Linux OS package repositories yet.

clinssen:
N.B. the NESTML CI pipeline is already running Python 3.9.

clinssen:
Sorry, I was looking at an older local clone; NEST Simulator is also already at 3.9. Hence, merging.

clinssen (approved):

PR #800, opened 2022-07-28
Add automatic NEST version detection and update CI

clinssen:
There have recently been several recent user support requests that were resolved to (API) version incompatibilities between NEST and NESTML. In general, the remedy has been to recommend using the latest NEST-master and NESTML-master, but NESTML-latest should be made more compatible with earlier versions of NEST. This PR implements that functionality and cleans up the earlier NEST 2 compatibility workaround that used a separate ``CodeGenerator`` class.

The CI is adjusted so most tests now run on NEST 2.20.2, NEST 3.3, as well as master. It is easy to add additional versions by adding the branch (or tag) name to the GitHub actions file. Jupyter notebooks only run on latest master.

The NEST Simulator version is now automatically detected in ``NESTTools.detect_nest_version()``. A related upstream issue is https://github.com/nest/nest-simulator/issues/2116. This is not a blocker for this PR, as we can (to some degree) determine the NEST Simulator version by inspecting which API methods are defined.


pnbabu:
@clinssen I installed a new environment with NEST 3.3 and executed the command `nest-config --version` which indeed prints `3.3` as output. Moreover, for the latest master, the same command prints `master@<git-hash>`. Can we use this to check the version number instead of deciding based on the output of `dir(nest)`?

clinssen:
@pnbabu: the issue is that some people might have installed from git, say, shortly after the v2.20.2 release, when the "master@<hash>" would have corresponded best to a 2.20.2 detected version.

.github/workflows/nestml-build.yml
```diff
@@ -1,10 +1,26 @@
...
+    #        * * * * *
+#    - cron: '30 5 * * 0'
+
 jobs:
-  nest-3-latest:
+  build_and_test:
     runs-on: ubuntu-latest
     strategy:
+      matrix:
+        nest_branch: ["v2.20.2", "v3.3", "master"]
```
pnbabu, line 23: Should we actively test for all other versions of NEST?

pynestml/codegeneration/nest_code_generator.py
```diff
@@ -138,6 +161,12 @@ def raise_helper(self, msg):
         raise TemplateRuntimeError(msg)
 
     def set_options(self, options: Mapping[str, Any]) -> Mapping[str, Any]:
+        # insist on using the old Archiving_Node class for NEST 2
+        if self.get_option("nest_version").startswith("v2"):
```
pnbabu, line 165: Is this required here again? It is already mentioned in line number 120.

tests/nest_tests/nest_custom_templates_test.py
```diff
@@ -18,19 +18,26 @@
...
 
+nest_version = NESTTools.detect_nest_version()
+
 
 class NestCustomTemplatesTest(unittest.TestCase):
     """
     Tests the code generation and installation with custom NESTML templates for NEST
     """
 
+    @pytest.mark.skipif(nest_version.startswith("v2"),
```
pnbabu, line 39: Why don't these tests support NEST 2?

tests/nest_tests/nest_integration_test.py
```diff
@@ -108,7 +115,10 @@ def test_nest_integration(self):
                               {"tau_syn_rise_E": 2., "tau_syn_decay_E": 10., "tau_syn_rise_I": 2.,
                                "tau_syn_decay_I": 10.}))  # XXX: TODO: does not work yet when tau_rise = tau_fall (numerical singularity occurs in the propagators)
 
-        neuron_models.append(("izhikevich", "izhikevich_nestml", 1E-3, 1))        # large tolerance because NEST Simulator model does not use GSL solver, but simple forward Euler
+        if nest_version.startswith("v2"):
+            neuron_models.append(("izhikevich", "izhikevich_nestml", 1E-3, 1, {}, {}, {"V_m": -70., "U_m": .2 * -70.}))        # large tolerance because NEST Simulator model does not use GSL solver, but simple forward Euler
```
pnbabu, line 119: Why are the values different for different NEST versions?

tests/nest_tests/nest_loops_integration_test.py
```diff
@@ -18,20 +18,28 @@
...
 
+nest_version = NESTTools.detect_nest_version()
+
+
 class NestLoopsIntegrationTest(unittest.TestCase):
     """
     Tests the code generation and working of for and while loops from NESTML to NEST
     """
 
+    @pytest.mark.skipif(nest_version.startswith("v2"),
```
pnbabu, line 41: The `while` and `for` loops should be supported for all NEST versions, correct?

tests/nest_tests/stdp_neuromod_test.py
```diff
@@ -72,6 +77,8 @@ def setUp(self):
                              codegen_opts={"neuron_parent_class": "ArchivingNode",
                                            "neuron_parent_class_include": "archiving_node.h"})
 
+    @pytest.mark.skipif(nest_version.startswith("v2"),
```
pnbabu, line 80: I saw some changes to the `SynapseHeader` template for volume transmitter to make it compatible with NEST version 2. Since this test does not support NEST 2, does this mean that those changes can be removed?

tests/nest_tests/stdp_synapse_test.py
```diff
@@ -50,26 +54,35 @@ class NestSTDPSynapseTest(unittest.TestCase):
...
-                                           "neuron_synapse_pairs": [{"neuron": "iaf_psc_exp",
-                                                                     "synapse": "stdp",
-                                                                     "post_ports": ["post_spikes"]}]})
+                             codegen_opts=jit_codegen_opts)
+
+        if nest_version.startswith("v2"):
+            non_jit_codegen_opts = {"neuron_parent_class": "Archiving_Node",
+                                    "neuron_parent_class_include": "archiving_node.h"}
+        else:
+            non_jit_codegen_opts = {"neuron_parent_class": "ArchivingNode",
```
pnbabu, line 76: Shouldn't this be `StructuralPlasticityNode`?

pnbabu (commented):
This looks great! I have a question - could you please comment on what basis you decide the compatibility of tests for NEST 2?

.github/workflows/nestml-build.yml
```diff
@@ -149,91 +136,16 @@ jobs:
...
-          sed -i 's|%NEST_PATH%|${{ env.NEST_INSTALL }}|' tests/nest_tests/resources/nest2_allmodels_codegen_opts.json
-          cat tests/nest_tests/resources/nest2_allmodels_codegen_opts.json
-          nestml --input_path $ALL_MODEL_FILENAMES --target_path target --target_platform NEST2 --suffix _nestml --logging_level INFO --module_name nestml_allmodels_module --codegen_opts tests/nest_tests/resources/nest2_allmodels_codegen_opts.json
-
-      # Integration tests
-      - name: Run integration tests
-        env:
-          LD_LIBRARY_PATH: ${{ env.NEST_INSTALL }}/lib/nest
+      # Run IPython/Jupyter notebooks
+      - name: Run Jupyter notebooks
```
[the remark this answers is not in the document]
clinssen: Because we need to do ugly things in some tests like revert to ``GetStatus()`` rather than ``get()``, which is all the more visible in the notebooks.

.github/workflows/nestml-build.yml
clinssen: Let's discuss this with @jougs after the merge. Compatibility is anyway easy to add now, but I'd like to get this in before waiting for the outcome of that discussion.

pynestml/codegeneration/nest_code_generator.py
clinssen: I agree it's not ideal, but it's duplicated to keep things consistent. It might be that it's set by the constructor, but overwritten by the ``generate_code()`` invocation with custom code generator options (e.g. setting ``StructuralPlasticityNode`` as the parent class), which needs to be overridden again when the actual code is generated.

tests/nest_tests/nest_custom_templates_test.py
clinssen: No particular reason, except that the code of a lot of tests becomes considerably more ugly when adding NEST 2 support back in (``GetStatus()`` instead of ``get()`` and so on). So I restricted the set of NEST 2 tests to those testing the core functionality. We can adapt more tests on a case-by-case basis if we deem it necessary.

tests/nest_tests/nest_integration_test.py
clinssen: The initial values of the Izhikevich neuron have changed from NEST 2 to NEST 3. This line is to make them consistent between both NEST versions and NESTML model defaults.

tests/nest_tests/nest_loops_integration_test.py
clinssen: Probably, but see the comment above, not a test for core functionality.

tests/nest_tests/stdp_neuromod_test.py
clinssen: The compatibility could still be used by NESTML users even though it's not unit tested right now.

tests/nest_tests/stdp_synapse_test.py
clinssen: It's the "non-jit" version, which tests things using the "old" ``ArchivingNode``.

pnbabu: Is there a need for the if condition then? The if and else have exactly the same statements here.

clinssen: Note the presence/lack of underscoren in the class name ``Archiving_Node``.

pnbabu: Got it, my bad!

pnbabu (approved):

PR #801, opened 2022-08-16
Modify input files in tests

pnbabu:
Modify integration tests to run from any directory by providing absolute paths of input files to `generate_nest_target()` function.


tests/nest_tests/stdp_triplet_synapse_test.py
```diff
@@ -18,6 +18,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+import os
```
clinssen, line 21: we should move towards import statements ordered according to:

[import python standard library]

[import our own things, like nest and pynestml]

so in this case probably something like

```python
import numpy as np
import os
import pytest
```

clinssen (commented):
Thanks for the PR! Just a few minor cleanups for consistency.

tests/nest_tests/stdp_triplet_synapse_test.py
pnbabu: Do you also mean to order them alphabetically?

clinssen:
Great, many thanks!

PR #802, opened 2022-08-18
Add NEST Desktop target

pnbabu:
This PR adds NEST Desktop as a target platform to NESTML. The output of the code generator, in this case, will be a `.json` file containing the details of the neuron model file(s). These json files will be used by the NEST Desktop GUI to select models from a model repository and use them for further simulations.

Note: A part of the documentation for adding new targets is borrowed from #729

PR #811, opened 2022-09-13
Move templates ``raise_helper()`` to base class

clinssen:
This is just a small PR containing the templates ``raise_helper()`` bugfix from #802.


pnbabu (commented):
Isn't this already present in #802? Why is it made into a separate PR?

pnbabu (approved):

PR #820, opened 2022-10-03
Allow solver selection of numeric vs. analytic solver in NEST code generator

clinssen:
Allow users the option of disabling the analytic solver. This could be used, for instance, to avoid problems with numerical singularities (https://github.com/nest/ode-toolbox/pull/62).


pnbabu (approved):
Looks good!

PR #822, opened 2022-10-10
Modernize C++ coding styles

clinssen:
Follow-up to https://github.com/nest/nest-simulator/pull/2493

> 1. Using nullptr instead of NULL or 0 to initialize empty pointers or to check for an undefined pointer.
> 2.  Removing the redundant void argument from function with empty argument lists.
> 3.  Add missing override specifiers for overridden functions in derived classes and, if applicable, remove the redundant virtual specifier when override is set as well.


pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
```diff
@@ -648,7 +648,7 @@ public:
     // use accessor functions (inherited from Connection< >) to obtain delay and target
 {%- if paired_neuron is not none and paired_neuron|length > 0 %}
     {{paired_neuron}}* __target = static_cast<{{paired_neuron}}*>(get_target(tid));
-    assert(__target != NULL);
+    assert(__target != nullptr);
```
jougs, line 651:
```suggestion
    assert(__target);
```
See #2496

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -351,7 +351,7 @@ void {{neuronName}}::init_buffers_()
 {%- endif %}
 {%- if uses_numeric_solver %}
 
-  if ( B_.__s == 0 )
+  if ( B_.__s == nullptr )
```
JanVogelsang, line 354:
```suggestion
  if ( not B_.__s )
```

JanVogelsang: See #2496

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -360,7 +360,7 @@ void {{neuronName}}::init_buffers_()
     gsl_odeiv_step_reset( B_.__s );
   }
 
-  if ( B_.__c == 0 )
+  if ( B_.__c == nullptr )
```
JanVogelsang, line 363:
```suggestion
  if ( not B_.__c )
```
See above

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -369,7 +369,7 @@ void {{neuronName}}::init_buffers_()
     gsl_odeiv_control_init( B_.__c, P_.__gsl_error_tol, 0.0, 1.0, 0.0 );
   }
 
-  if ( B_.__e == 0 )
+  if ( B_.__e == nullptr )
```
JanVogelsang, line 372:
```suggestion
  if ( not B_.__e )
```
See above

pynestml/codegeneration/resources_nest/point_neuron/setup/@MODULE_NAME@.h.jinja2
```diff
@@ -82,7 +82,7 @@ public:
   /**
    * Return the name of your model.
    */
-  const std::string name( void ) const;
+  const std::string name() const;
```
pnbabu, line 85: The `void` keyword must also be removed from the function in the .cpp file.

JanVogelsang (changes requested):
1. There would be some changes needed in the NEST2 templates as well. Should we ignore those as its legacy code anyway?

2. In `NeuronHeader.jinja2` line 741, there is an `||` which should be replaced by an `or`. I couldn't find a way to add a suggestion to that line, probably because it was not part of the changed lines sections.  
3. I could also find 6 occurrences of `&&`, which should be replaced by `and`.

4. In `NeuronHeader.jinja2` line 209, the destructor is missing an `override` specifier.

5. In `SynapseHeader.jinja2`, `handles_test_event` functions are missing `override` specifiers.

pynestml/codegeneration/resources_nest/point_neuron/setup/@MODULE_NAME@.h.jinja2
JanVogelsang: I am confused.. on the _modernize_ branch of @clinssen I can't see the void keyword being present there..

pnbabu: See here: https://github.com/clinssen/nestml/blob/c879b09c7531357904c7dd1bc7a2f63d6a5272b9/pynestml/codegeneration/resources_nest/point_neuron/setup/%40MODULE_NAME%40.cpp.jinja2#L116

JanVogelsang: Ah nevermind, I just didn't read your comment properly. Overlooked the part where you said "in the .cpp file", just read "the void keyword must be removed", looked at the file where you left the comment and was confused. But true, I will create a PR against Charl's branch with the changes I suggested below and also add this suggestion.

clinssen:
Thanks all for the suggestions! @JanVogelsang I merged your PR branch. Could you re-review?

JanVogelsang (approved):
LGTM!

clinssen:
Many thanks!

PR #825, opened 2022-10-11
Fix NEST version compatibility

pnbabu:
NEST `v3.0` check is missing in some condition statements in the templates which results in compilation errors while building the code for NEST 3.0. This PR adds the missing condition.


clinssen (approved):
Many thanks!

Issue #609, opened 2021-01-27
Fix use of predefined "simulation time" variable t


clinssen:
The predefined NESTML variable ``t`` should refer to the current simulation time in the context where it is used, but proper code is not generated in all contexts.

For example, in the numerical (GSL) stepping function context, the first argument is unnamed, but should be ``double t`` if ``t`` is referred to in a time-dependent ODE:
```C++
extern "C" inline int iaf_psc_exp_nestml_dynamics(double, const double ode_state[], double f[], void* pnode){
```

In the neuron model ``update()`` method, "the current simulation time" should instead be expressed in terms of the ``origin``, ``from`` and ``to`` parameters:
```c++
void iaf_psc_exp_nestml::update(nest::Time const & origin,const long from, const long to){
  for ( long lag = from ; lag < to ; ++lag ) {
```

clinssen:
Suggestion is to use a private member variables "curr_time_", which is updated immediately after the ``for (...`` in the update() method. All occurrences of *t* can then be simply converted as ``this->curr_time_``.

PR #826, opened 2022-10-12
Fix accessing predefined time variable t in synapse event handlers

clinssen:
This allows using the predefined time variable ``t`` in the synapse pre and post spike event handlers (in the ``onReceive`` blocks).

### DIFF
--- a/pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
+++ b/pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
@@ -716,6 +716,7 @@ public:
         timestep += (start->t_ + __dendritic_delay) - t_lastspike_;
 
         const double _tr_t = start->t_;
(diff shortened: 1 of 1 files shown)


clinssen:
See also #609.

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -512,6 +512,7 @@ double {{neuronName}}::get_delayed_{{variable.get_symbol_name()}}() const
 
 void {{neuronName}}::update(nest::Time const & origin,const long from, const long to)
 {
+  auto get_t = [](){ return nest::Time( nest::Time::step( nest::kernel().simulation_manager.get_slice_origin().get_steps() ) ).get_ms(); };
```
clinssen, line 515: Perhaps this should be a function of the ``origin`` parameter (+ lag) rather than making a call to the kernel.

mlober (approved):
Thank's for this useful extension. The install and the test_time_variable test work for me. Compiling and accessing 't' with a synapse model of my own works as well. The code itself looks good to me.

clinssen:
Thanks for the review!

PR #828, opened 2022-10-13
Handle parser error

pnbabu:
Currently, if there is a parser error, the execution continues to the building phase without exiting. This PR handles the error and exits the execution.


clinssen:
Originally, NESTML was designed to continue processing even after an error occurred. This resulted in a lot of unnecessary errors and churn, because downstream code would have to try its best to accept the garbage it was receiving from the upstream code which encountered an error. This is why we now prefer to an "exit on first error" strategy, however, this is not fully implemented yet. Could you perhaps see if the ``errors_occurred`` variable can be removed and replaced by return statements?

clinssen:
Notes from meeting November 2nd:
- remove ``errors_occurred`` flag from ``process()``; replace with return-early pattern
- remove "development mode" that tries to continue generating code in spite of errors; "development mode" should be retained but only set the additional debugging flag for the templates

pynestml/frontend/pynestml_frontend.py
```diff
@@ -290,35 +291,31 @@ def process():
...
-                                       log_level=LoggingLevel.WARNING)
-                    models.remove(model)
-                    errors_occurred = True
+        for model in models:
+            if Logger.has_errors(model):
+                code, message = Messages.get_model_contains_errors(model.get_name())
+                Logger.log_message(node=model, code=code, message=message,
+                                   error_position=model.get_source_position(),
+                                   log_level=LoggingLevel.WARNING)
+                models.remove(model)
```
clinssen, line 300:
```suggestion
```

clinssen (changes requested):

clinssen (approved):

PR #829, opened 2022-10-13
Eliminate unnecessary target from CMakeLists

clinssen:
This makes compiling/building the generated NESTML code precisely twice as fast.

@jougs: Could you comment on why the other .so was (is?) necessary?


jougs (approved):
@clinssen: I *assume* that the one without the `lib` prefix is meant for dynamic loading the module at run-time through the extension module mechanism, while the one that has it is meant for dynamically linking the module against NEST at compile-time using its `-Dexternal-modules` CMake flag (which probably no one knows about).

That said, I am still fine with the changes at hand and suggest to revisit this once the new module system is in place and we know exactly how static and dynamic linking at compile- and run-time looks then.

PR #835, opened 2022-11-01
Make ``get_blocks()`` functions consistently return lists

clinssen:
This is a purely internal refactoring.

Make ``get_equations_blocks()``, ``get_parameters_blocks()`` and other block-getter function on the ``ASTNeuronOrSynapse`` return lists in a consistent manner. Also de-duplicate some code and fix imports.

Motivation for preferring the list option rather than single elements, is because this leaves the way open in the future, where NESTML could have an "import" or "include" mechanism, to make models more modular. Then, two different blocks of the same type (e.g. ``state``, ``parameters``) from two different .nestml files might end up in the same model.

### DIFF
--- a/pynestml/codegeneration/nest_code_generator.py
+++ b/pynestml/codegeneration/nest_code_generator.py
@@ -40,7 +40,6 @@ from pynestml.codegeneration.printers.nest_reference_converter import NESTRefere
 from pynestml.codegeneration.printers.ode_toolbox_reference_converter import ODEToolboxReferenceConverter
 from pynestml.frontend.frontend_configuration import FrontendConfiguration
 from pynestml.meta_model.ast_assignment import ASTAssignment
-from pynestml.meta_model.ast_equations_block import ASTEquationsBlock
 from pynestml.meta_model.ast_input_port import ASTInputPort
 from pynestml.meta_model.ast_neuron import ASTNeuron
 from pynestml.meta_model.ast_kernel import ASTKernel
@@ -64,9 +63,7 @@ from pynestml.visitors.ast_random_number_generator_visitor import ASTRandomNumbe
 def find_spiking_post_port(synapse, namespace):
     if "paired_neuron" in dir(synapse):
         for post_port_name in namespace["post_ports"]:
-            if synapse.get_input_blocks() \
-                    and synapse.get_input_blocks().get_input_ports() \
-                    and ASTUtils.get_input_port_by_name(synapse.get_input_blocks(), post_port_name).is_spike():
+            if ASTUtils.get_input_port_by_name(synapse.get_input_blocks(), post_port_name).is_spike():
                 return post_port_name
     return None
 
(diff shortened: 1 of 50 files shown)


pynestml/utils/ast_utils.py
```diff
@@ -1878,13 +1902,20 @@ def find_model_by_name(cls, model_name: str, models: Iterable[Union[ASTNeuron, A
         return None
 
     @classmethod
-    def get_convolve_function_calls(cls, ast):
+    def get_convolve_function_calls(cls, nodes: Union[ASTNode, List[ASTNode]]):
         """
         Returns all sum function calls in the handed over meta_model node or one of its children.
         :param ast: a single meta_model node.
```
pnbabu, line 1908:
```suggestion
        :param nodes: a single or list of AST nodes.
```

pynestml/utils/ast_utils.py
```diff
@@ -1878,13 +1902,20 @@ def find_model_by_name(cls, model_name: str, models: Iterable[Union[ASTNeuron, A
...
         :type ast: ASTNode
```
pnbabu, line 1909: Can be removed.

pynestml/utils/ast_utils.py
```diff
@@ -1900,7 +1931,7 @@ def contains_convolve_function_call(cls, ast):
     @classmethod
     def get_function_calls(cls, ast_node, function_list):
         """
-        For a handed over list of function names, this method retrieves all functions in the meta_model.
+        For a handed over list of function names, this method retrieves all function calls in the meta_model.
         :param ast_node: a single meta_model node
         :type ast_node: ASTNode
```
pnbabu, line 1936: Could you remove the type here and add that in the function signature?

pynestml/utils/ast_utils.py
```diff
@@ -1917,3 +1948,11 @@ def get_function_calls(cls, ast_node, function_list):
         vis = ASTHigherOrderVisitor(visit_funcs=fun)
         ast_node.accept(vis)
         return res
+
+    @classmethod
+    def resolve_to_variable_symbol_in_blocks(cls, variable_name: str, blocks: List[ASTBlock]):
```
pnbabu, line 1953: Could you add a docstring for this function?

pnbabu (commented):
Thanks for this clean-up! I have some minor comments.

pynestml/codegeneration/nest_code_generator.py
```diff
@@ -423,7 +425,7 @@ def _get_synapse_model_namespace(self, synapse: ASTSynapse) -> Dict:
             and self.analytic_solver[synapse.get_name()] is not None
         if namespace["uses_analytic_solver"]:
             namespace["analytic_state_variables"] = self.analytic_solver[synapse.get_name()]["state_variables"]
-            namespace["variable_symbols"].update({sym: synapse.get_equations_block().get_scope().resolve_to_symbol(
+            namespace["variable_symbols"].update({sym: synapse.get_equations_blocks()[0].get_scope().resolve_to_symbol(
```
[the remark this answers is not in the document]
clinssen: I prefer just passing the synapse object as the function parameter and grabbing it again from there.

pynestml/codegeneration/nest_code_generator.py
```diff
@@ -64,9 +63,7 @@
 def find_spiking_post_port(synapse, namespace):
```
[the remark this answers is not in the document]
clinssen: Yes! But this is already being refactored in one of the other PRs, so I'd like to keep this separate for now.

pynestml/utils/ast_utils.py
clinssen: ?

pnbabu: I meant the type in docstring can be removed.

pnbabu (approved):
Looks good to me!

pynestml/utils/ast_utils.py
clinssen: Oh, of course, sorry.

clinssen:
Cheers!

PR #836, opened 2022-11-01
Unify the implementation of input ports

### LINKED ISSUE #823: Support vectorized input ports
Currently, the implementation for vectorized input ports (see below for example) is rudimentary. The implementation needs to be unified for all kinds of input ports -- vectorized and non-vectorized, with and without the `excitatory` or `inhibitory` qualifier keyword.

```
neuron foo:
  input:
    AMPA_spikes pA <- excitatory spike
    GABA_spikes pA <- inhibitory spike
    NMDA_spikes pA <- spike
    foo[2] pA <- spike
    my_spikes[42] pA <- excitatory spike 
    my_spikes2[42] pA <- inhibitory spike
  end  
end
```

pnbabu:
This PR unifies the implementation of input ports in NESTML for the NEST target. All the spike inputs are now combined in a single vector of `RingBuffer`s. Since we route `excitatory` and `inhibitory` inputs into different buffers in NESTML, this PR also contains an implementation for mapping the NEST `rport` to its NESTML port.

Note: The support for vector variables in the differential equations is still pending. This is a precursor to that PR as is currently being worked upon.


clinssen:
This looks great, could you also add a paragraph in the documentation?

pnbabu:
Generated readthedocs: https://nestml-doc.readthedocs.io/en/latest/nestml_language/neurons_in_nestml.html

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,34 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+  end
+
+The mapping of NEST ``rport`` and NESTML input port numbers for the above example would be as shown below:
+
+.. list-table::
+   :header-rows: 1
+
+   * - Input port name
+     - NEST ``rport``
+     - NESTML port number
```
clinssen, line 90: Is there such a thing as a NESTML port number?

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,34 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+   * - GABA_spikes
+     - 1
+     - 2
+   * - NMDA_spikes
+     - 2
+     - 3
+
 
 Integrating current input
 ^^^^^^^^^^^^^^^^^^^^^^^^^
```
clinssen, line 103: Please document ``RportToPortMapEntry`` and its usage too (or link to an example).

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -336,8 +360,12 @@ void {{neuronName}}::init_state_(const Node& proto)
 
 void {{neuronName}}::init_buffers_()
 {
-{%- for inputPort in neuron.get_input_ports() %}
-{{ buffer_initialization.BufferInitialization(inputPort) }}
+{% if (neuron.get_spike_input_ports())|length > 0 -%}
+  get_spike_inputs_().clear(); // includes resize
```
clinssen, line 364: what does "include resize" mean here?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronHeader.jinja2
```diff
@@ -378,22 +379,43 @@ private:
...
+    {%- set ns.count = ns.count + 1 -%}
+  {%- endif -%}
+{%- endfor %}
+    SUP_SPIKE_RECEPTOR = {{ns.count}}
   };
+
+  static const size_t NUM_SPIKE_RECEPTORS = SUP_SPIKE_RECEPTOR - MIN_SPIKE_RECEPTOR;
+
+{% if neuron.get_spike_input_ports()|length > 1 -%}
+  static std::map<std::tuple< int, std::string >, int> rport_to_nestml_port;
```
clinssen, line 414:
```suggestion
  static std::map< std::tuple< int, std::string >, int > rport_to_nestml_port;
```

doc/nestml_language/neurons_in_nestml.rst
pnbabu: The usage is in the example mentioned above. The usage in PyNEST will be linked in the next PR where the vector input ports can be actually used in the differential equations.

doc/nestml_language/neurons_in_nestml.rst
pnbabu: No, it's just the mapping that we have between the NEST rport and its corresponding index in NESTML. Do you want me to change it?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
pnbabu: This was copied over from an old NESTML template code. I think it means that the `clear` function clears all the elements of the vector and sets the size to 0. I will remove it to avoid any confusion.

clinssen: Oh, I see, thanks for the explanation.

doc/nestml_language/neurons_in_nestml.rst
clinssen: I guess the "NESTML port number" column can be removed entirely. Can we make this example more complex by also having a vectorized port (or would this be part of the other PR)?

pnbabu: Added a detailed example.

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,69 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+The input ports can also be defined as a vector. For example,
+
+.. code-block:: nestml
+
+   parameters:
+    TWO integer = 2
+   end
+
+   input:
+     foo[TWO] pA <- spike
```
clinssen, line 82: Do we define what happens when the value of ``TWO`` is written to with a different number?

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,69 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+     exc_spikes[3] pA <- excitatory spike
+     inh_spikes[3] pA <- inhibitory spike
+   end
+
+In this example, all the spiking input ports are defined as vectors. The integer surrounded by ``[`` and ``]`` determines the size of the vector. The size can either be directly specified as an integer or defined as a parameter of type integer in the ``parameters`` block. The name of the receptors of these ports are denoted by suffixing the ``vector index + 1`` to the port name. In the above example, the receptor name for ``foo[0]`` would be ``FOO_1``.
```
clinssen, line 87: I don't understand the ``FOO_1`` notation, is this within the NESTML context? Can we just refer to ``foo[0]`` inside the NESTML model?

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,69 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+
+Each connection in NEST is denoted by a receiver port or ``rport`` number which is an integer that starts with 0. As discussed above, NESTML routes the spikes with ``excitatory`` and ``inhibitory`` qualifiers into separate input buffers, whereas NEST identifies them with the same ``rport`` number. Thus during code generation for NEST, NESTML maintains an internal mapping between NEST ``rports`` and NESTML input ports. A list of receptor names and their corresponding ``rport`` numbers defined in a model can be queried using the NEST API ``nest.GetStatus("neuron, receptor_types")``.
+
+.. code-block:: nestml
+
+   parameters:
+    TWO integer = 2
+   end
+
+   input:
```
clinssen, line 97: use just one single NESTML code example for here and just above

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,69 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+
+.. list-table::
+   :header-rows: 1
+
+   * - Input port name
+     - NEST ``rport``
+   * - AMPA_spikes
+     - 0
+   * - GABA_spikes
+     - 0
```
clinssen, line 116: This should be 1 (AMPA_spikes); 2 (GABA_spikes), etc., no?

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -70,6 +70,69 @@ It is equivalent if either both `inhibitory` and `excitatory` are given, or neit
...
+
+   input:
+     foo[TWO] pA <- spike
+     exc_spikes[3] pA <- excitatory spike
+     inh_spikes[3] pA <- inhibitory spike
+   end
+
+In this example, all the spiking input ports are defined as vectors. The integer surrounded by ``[`` and ``]`` determines the size of the vector. The size can either be directly specified as an integer or defined as a parameter of type integer in the ``parameters`` block. The name of the receptors of these ports are denoted by suffixing the ``vector index + 1`` to the port name. In the above example, the receptor name for ``foo[0]`` would be ``FOO_1``.
+
+Each connection in NEST is denoted by a receiver port or ``rport`` number which is an integer that starts with 0. As discussed above, NESTML routes the spikes with ``excitatory`` and ``inhibitory`` qualifiers into separate input buffers, whereas NEST identifies them with the same ``rport`` number. Thus during code generation for NEST, NESTML maintains an internal mapping between NEST ``rports`` and NESTML input ports. A list of receptor names and their corresponding ``rport`` numbers defined in a model can be queried using the NEST API ``nest.GetStatus("neuron, receptor_types")``.
```
clinssen, line 89: Please try to write the documentation here completely agnostic to the simulation platform target (inhibitory/excitatory distinction can be explained just on the sign of the incoming weight). Could you move all NEST Simulator specific material to "Running NESTML -> NEST Simulator target"?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -537,45 +561,34 @@ void {{neuronName}}::handle(nest::DataLoggingRequest& e)
...
-  // this port receives excitatory spikes
+  double weight = e.get_weight();
+  int nestml_port = 0;
+{%- if neuron.get_spike_input_ports()|length > 1 %}
   if ( weight >= 0.0 )
   {
-    get_{{port.get_symbol_name()}}().
-        add_value(e.get_rel_delivery_steps( nest::kernel().simulation_manager.get_slice_origin()),
-                       weight * multiplicity );
+    if( rport_to_nestml_port.find(std::make_pair(e.get_rport(), "spike")) != rport_to_nestml_port.end() )
```
clinssen, line 571: Why not make the mapping the other way around, so the call to ``find()`` is not necessary?

doc/nestml_language/neurons_in_nestml.rst
clinssen, line 89:
```suggestion
Each connection in NEST is denoted by a receiver port or ``rport`` number which is an integer that starts with 0. As discussed above, NESTML routes the spikes with ``excitatory`` and ``inhibitory`` qualifiers into separate input buffers, whereas NEST identifies them with the same ``rport`` number. Thus during code generation for NEST, NESTML maintains an internal mapping between NEST ``rports`` and NESTML input ports. A list of receptor names and their corresponding ``rport`` numbers defined in a model can be queried using the NEST API ``nest.GetStatus("neuron", "receptor_types")``.
```
Could you paste inline what this command would return for the running example?

doc/nestml_language/neurons_in_nestml.rst
pnbabu: Do you mean to say something like `TWO = 3`?

doc/nestml_language/neurons_in_nestml.rst
pnbabu: The `FOO_1` notation is only in the context of NEST where the `receptor_type` is provided in `nest.Connect` calls. It is not in the NESTML context, so I will move it.

doc/nestml_language/neurons_in_nestml.rst
pnbabu: Yes, it should start with 1, my bad. I will fix it.

doc/nestml_language/neurons_in_nestml.rst
pnbabu: Added a CoCo to check if the size of the vector is an integer.

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
pnbabu: Since we have to route spikes based on `rport` in the event via `e.get_rport()`, the map has the `rport` as the key. How do you suggest otherwise?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -116,6 +118,26 @@ template <> void RecordablesMap<{{neuronName}}>::create()
   }
 }
 
+{%- if neuron.get_spike_input_ports()|length >1 %}
+std::map<std::tuple< int, std::string >, int> {{neuronName}}::rport_to_nestml_port =
+{
+    {%- for key, values in utils.get_spike_input_ports_in_pairs(neuron).items() %}
```
clinssen, line 124: Would you mind changing the indentation here so that the ``{%`` is always at the start of the line, followed by as much indentation is necessary for the jinja bits? E.g.
```Jinja
{% for foo in bar %}
{%   if baz == "xyzzy" %}
{%     set doge = 42 %}
```

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -116,6 +118,26 @@ template <> void RecordablesMap<{{neuronName}}>::create()
   }
 }
 
+{%- if neuron.get_spike_input_ports()|length >1 %}
```
clinssen, line 121:
```suggestion
{%- if neuron.get_spike_input_ports()|length > 1 %}
```

tests/nest_tests/input_ports_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class TestInputPorts:
+    """
+    Tests the different kind of input ports supported in NESTML.
+    """
+    def test_input_ports(self):
+        input_path = os.path.join(os.path.realpath(os.path.join(
+            os.path.dirname(__file__), "resources", "InputPorts.nestml")))
+        target_path = "../target"
```
clinssen, line 45:
```suggestion
        target_path = "target"
```

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
clinssen: OK, sorry, my previous suggestion was not very helpful, but I'm still worried about the runtime performance hit of using ``std::map.find()``, because this method will be called whenever an incoming spike is handled, so quite a lot during runtime. Possibly, we have to do some more low-level C-style array indexing magic.

For instance, ``rport_to_nestml_port`` could be a plain C array where the integer index is the NEST rport. There could be a second boolean array of the same length which says if this input port uses the same or separate entries for exc/inh spikes. In case of inhibitory spikes, we would just add 1 to the index. This is less "clean" coding-wise but may be necessary.

@jougs: do you have any suggestion here, should we benchmark the current implementation first? Or do you agree to reject the use of ``find()`` offhand?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronHeader.jinja2
```diff
@@ -382,22 +383,43 @@ private:
 {%- endif %}
 
 private:
-  {% if (neuron.get_multiple_receptors())|length > 1 -%}
+  {% if has_multiple_synapses -%}
   /**
    * Synapse types to connect to
-   * @note Excluded upper and lower bounds are defined as INF_, SUP_.
+   * @note Excluded lower and upper bounds are defined as MIN_, SUP_.
```
clinssen, line 389: I'm fine using either infimum and supremum or minimum and maximum, but please use either one or the other combination.

tests/invalid/CoCoVectorInputPortSizeAndType.nestml
```diff
@@ -0,0 +1,44 @@
...
+neuron CoCoVectorInputPortSizeAndType:
+  parameters:
+    TWO integer = 2
+  end
+
+  input:
+    foo[TWO] pA <- spike
+    bar[-3] pA <- spike
+  end
+end
```
clinssen, line 44:
```suggestion
end

```

tests/nest_tests/resources/InputPorts.nestml
```diff
@@ -0,0 +1,51 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+"""
+neuron input_ports:
+
+  state:
+    bar pA = 1.5 pA
+    vec[3] real = 3.3
```
clinssen, line 36: remove?

tests/valid/CoCoVectorInputPortSizeAndType.nestml
```diff
@@ -0,0 +1,41 @@
...
+You should have received a copy of the GNU General Public License
+along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+"""
+neuron CoCoVectorInputPortSizeAndType:
+  input:
+    foo[2] pA <- spike
+    bar1[3] pA <- excitatory spike
+    bar2[3] pA <- inhibitory spike
+  end
+end
```
clinssen, line 41:
```suggestion
end

```

clinssen (commented):
Let's discuss the main issue (having to use ``find()`` on the mapping) with @jougs before fixing the smaller comments.

pynestml/cocos/co_co_vector_input_port_correct_size_type.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class InputPortsVisitor(ASTVisitor):
+    """
+    This visitor checks if the size of the vector input port is of the type integer and its value is greater than 0
+    """
+    def visit_input_port(self, node: ASTInputPort):
+        size_parameter = node.get_size_parameter()
+        if size_parameter:
+            # For negative numbers, the size_parameter is an expression
```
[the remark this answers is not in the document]
pnbabu: For negative numbers, the `size_parameter` is parsed as an `ASTExpression`. So we have to check for the `unary_operator` along with the `numeric_literal`. See also: https://github.com/nest/nestml/blob/cf8b13145f8460c062ca730944393a48495a206f/pynestml/cocos/co_co_vector_declaration_right_size.py#L60

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
pnbabu: I have now modified the mapping of `rport` numbers with `std::vector` instead of a `std::map`.
@clinssen Could you please have a look?

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
```diff
@@ -544,45 +565,24 @@ void {{neuronName}}::handle(nest::DataLoggingRequest& e)
...
-{%-   for port in neuron.get_spike_input_ports() %}
-{%-     if port.is_excitatory() and port.is_inhibitory() %}
-  // this port receives both excitatory and inhibitory spikes (where inhibitory spikes are those where weight < 0)
-  get_{{port.get_symbol_name()}}().
-      add_value(e.get_rel_delivery_steps( nest::kernel().simulation_manager.get_slice_origin()),
-                     weight * multiplicity );
-{%-     elif port.is_excitatory() %}
-  // this port receives excitatory spikes
+  double weight = e.get_weight();
+  int nestml_buffer_idx = 0;
```
clinssen, line 571: should this be a size_t?

clinssen (changes requested):
Thanks, this looks really good. Just a few more small points!

doc/running.rst
```diff
@@ -155,6 +155,95 @@ Custom templates
...
+
+The above code querying for ``receptor_types`` gives a list of port names and NEST ``rport`` numbers as shown below:
+
+.. list-table::
+   :header-rows: 1
+
+   * - Input port name
+     - NEST ``rport``
+   * - AMPA_spikes
+     - 0
```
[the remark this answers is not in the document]
pnbabu: This is NEST `rport` and it starts from 0.

pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
pnbabu: Yes, changed it to `size_t`.

doc/running.rst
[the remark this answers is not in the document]
pnbabu: This should start from 1. Sorry for the confusion. I have changed it now.

clinssen:
Great, thank you!

clinssen (approved):

PR #839, opened 2022-11-08
Allow synapse-neuron co-generation with no equations block defined in neuron

clinssen:
@tomtetzlaff: could I ask you to check if this works for your "ignore and fire" neuron? Cheers!


tomtetzlaff (approved):
Yes, this works. Thanks.

clinssen:
Thanks for the quick review!

PR #841, opened 2022-11-15
Make later CI jobs depend on earlier jobs

clinssen:


med-ayssar (approved):
Cannot we ignore some file from being checked in the link checker to prevent the failing?
Otherwise, looks good to me.

clinssen:
Thank you for the review! The curent failure is due to an upstream issue with https://nest-extension-module.readthedocs.org/, colleagues have already been notified and it should be fixed soon.

PR #842, opened 2022-11-23
Change NEST version detection from multiprocessing to subprocess

clinssen:
The ``multiprocessing`` package was leaking state or environment between the host and child process, causing issues on JURECA. Thanks to @shimoura for reporting this.


clinssen:
@shimoura confirmed via email that this fixes the issue on JURECA.

@pnbabu: could you check if this works OK on MacOS? Thanks!

pnbabu:
@clinssen this works perfectly fine on macOS. This also eliminates the need to enclose the script in the `if __name__ == '__main__'` condition.

pnbabu (approved):

PR #847, opened 2022-12-20
Fix initialization of templates

clinssen:
Make sure entries in the ``_model_templates`` and ``_module_templates`` members are always initialized (even if just an empty list).


pnbabu (approved):
Thanks!

pnbabu (approved):

PR #851, opened 2023-01-14
Fix naming of V_abs to V_rel

clinssen:
In several models, the membrane potential was intended to be defined relative to the resting or reset potential. However, the membrane potential state variable was mistakenly called V_abs instead of V_rel. To prevent confusion, we eliminate the use of relative (to zero) membrane potentials and use only absolute potentials. There could theoretically be a minor impact on performance, as we have to now subtract the leak or reset potential from the right-hand side of the ODE for V_m, but this is assumed to be negligible.


pnbabu (approved):

clinssen:
Cheers!

PR #852, opened 2023-01-15
Fix API doc build environment requirements

clinssen:
The ``requirements.txt`` for the docs is apparently no longer including the dependencies in ``requirements.txt`` of NESTML itself (in the package root). As ``requirements.txt`` files do not support importing one into the other, copy-paste the one into the other while adding some cautionary notes about this as comments.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -3,4 +3,14 @@ nbsphinx
 docutils
 sphinx>=4
 sphinx_rtd_theme>=1.0.0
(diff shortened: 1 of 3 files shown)


clinssen:
Please see the raw/full logs; in particular, the API documentation fails to build due to import exceptions.

pnbabu (approved):
Looks good to me.

clinssen:
Cheers!

PR #865, opened 2023-02-16
Clean up input and target path processing in frontend

clinssen:


pynestml/frontend/frontend_configuration.py
```diff
@@ -250,25 +255,41 @@ def handle_target_platform(cls, target_platform: Optional[str]):
...
-            if os.path.isabs(path):
-                cls.target_path = path
-            # a relative path, reconstruct it. get the parent dir where models, pynestml etc. is located
-            else:
-                pynestml_dir = os.getcwd()
-                cls.target_path = os.path.join(pynestml_dir, path)
+    def handle_target_path(cls, path: Optional[str]) -> None:
+        r"""Process the target path parameter.
+
+        Create the target path if it is specified as a string, but does not exist. Its parent directory has to exist already.
```
pnbabu, line 261:
```suggestion
        Create the target path directory if it is specified as a string, but does not exist. Its parent directory has to exist already.
```

pynestml/frontend/frontend_configuration.py
```diff
@@ -250,25 +255,41 @@ def handle_target_platform(cls, target_platform: Optional[str]):
...
+
+        code, msg = Messages.get_target_path_info(cls.target_path)
+        Logger.log_message(code=code, message=msg, log_level=LoggingLevel.INFO)
 
     @classmethod
-    def handle_install_path(cls, path):
+    def handle_install_path(cls, path: Optional[str]) -> None:
+        r"""Process the installation path parameter.
+
+        Create the installation path if it is specified as a string, but does not exist. Its parent directory has to exist already. The path will not be created if ``path`` is None."""
```
pnbabu, line 292:
```suggestion
        Create the installation path directory if it is specified as a string, but does not exist. Its parent directory has to exist already. The path will not be created if ``path`` is None."""
```

pnbabu (approved):
Thanks!

PR #870, opened 2023-02-22
Update link to available code generation options in docs

clinssen:
Make it more visible where to find the available code generation options.


med-ayssar (approved):
Looks good. Thanks!

PR #866, opened 2023-02-16
Remove `end` keyword from NESTML models

pnbabu:
Fixes #586

PR #875, opened 2023-03-07
Fix code generation for `autodoc` target

pnbabu:
This should be merged before #866


clinssen (approved):
Thank you!

N.B. re-generating the model docs will be done as part of #866.

PR #882, opened 2023-03-17
Remove units from spiking input ports

### LINKED ISSUE #727: Check/fix physical units of input spike train
The unit $[\delta(x)]$ of the Dirac delta "function" $\delta(x)$ follows
from its definition:

$f(y) = \int dx \delta(x-y) f(x)$

Here, f(.) is some arbitrary continuous function. As the f(.) on the lhs
has the same unit as the f(.) on the rhs, the unit of

$dx \delta(x)$

must be equal to 1. Hence, the unit $[\delta(x)]$ of the delta function
must be identical to the inverse $1/[x]$ of the unit $[x]$ of its argument
x, i.e.

$[\delta(x)] = 1/[x]$

This is a very generic property of the delta function. For the special
case where x is a measure of time with unit s, $\delta(x)$ must
consequently have units 1/s.

In the neuroscience context, we often use

$\xi(t) = \sum_k \delta(t-t_k)$

to represent sequences of spikes at times t_k (spike train), because
then we can easily construct other signals that are derived from spikes
by a convolution integral

$g(t)$
$= \int du \xi(u) h(t-u)$
$= \sum_k \int du \delta(u-t_k) h(t-u)$
$= \sum_k h(t-t_k)$

such as the number of spikes in a certain time interval, the synaptic
input current, the membrane potential, etc. If the kernel $h(t)$ on the
rhs has the same unit as $g(t)$, the delta function needs to have units of
1/s. Same reasoning as above.

Reference: https://en.wikipedia.org/wiki/Dirac_delta_function

With credits to @tomtetzlaff for pointing this out!

pnbabu:
Fixes #727 

This is based on #866 which should be merged first.


clinssen:
Thanks for all the work! I think we should stick to the original intention of the pull request, which was to remove units from spiking input ports. It is not clear at first glance what it means to have a unit next to an input port definition, so there is a high potential for mistakes. We should make it clear in the documentation that the units are 1/s, and the type checking system will take care of any potential inconsistencies. Unfortunately this will mean a little bit of extra syntax in the ODEs and convolutions where the units have to now be manually "multiplied in".

doc/models_library/index.rst
```diff
@@ -1,386 +1,49 @@
 Models library
 ==============
 
-Neuron models
```
clinssen, line 4: This file still needs to be hand-tuned, unfortunately. Could you leave it as-is for this PR? We will pick it up again in #920.

clinssen (commented):
Thank you for a lot of work! Just a few minor comments, mostly documentation-related.

clinssen (approved):
Great work, thank you!

PR #883, opened 2023-03-20
Change neuron partner preprocessor #DEFINE to typedef

clinssen:
This allows each synapse to have exactly one paired neuron. Previously, preprocessor ``#DEFINE``s could conflict with one another across different generated synapse model files.


pnbabu (approved):

clinssen:
Thanks!

PR #886, opened 2023-03-24
Fix OU noise tutorial

### LINKED ISSUE #876: Ornstein-Uhlenbeck noise neuron fails to spike
The neuron model in the Ornstein-Uhlenbeck noise tutorial notebook fails to spike and as a consequence fails to calculate the inter-spike intervals (the last plot in the notebook).

pnbabu:
Fixes #876 

With the new printers in place, the code resurfaced a bug in the notebook. The membrane capacitance was $1{\micro}A$ before which was too high when converted to the neuroscience-standard $pA$ in NEST which produced no spikes. The capacitance value is now changed to 250 $pA$ to match the original `iaf_psc_exp` model.


clinssen (approved):
Great, thank you!

PR #887, opened 2023-03-24
Improve handling of NEST synaptic delay parameter

### LINKED ISSUE #707: Fail with error in NEST code generator if ``@nest::delay`` is not supplied in synapse model

clinssen:
Synaptic delay and weight in NEST are a special case and should not be treated like any other parameter (or state variable). It should be get/set only using ``get_delay()`` and ``set_delay()`` inherited from the ``Connection`` base class. A member variable is already provided by the base class.

``get_weight()`` and ``set_weight()`` should be defined by the synapse model, but a member variable is not yet provided by the base class. We could make it such that no state or parameter marked ``@nest::weight`` is needed for minimalistic models.

When a decorator is specified in the model, the parameter can only be get/set using the name given as part of the decorator, termed the "namespace name" (i.e. "weight" instead of "w" and "delay" for "d").

As part of this PR, ``ASTExternalVariable`` was refactored out and replaced by the ``alternate_name`` (and ``alternate_scope``) fields of a regular ``ASTVariable``.

Fixes #707.


pnbabu:
As these decorators are specific to NEST, could we have these in the `codegen_opts` for the NEST target instead?

PR #893, opened 2023-04-13
Indent output block

pnbabu:


tests/resources/NestMLPrinterTest.nestml
```diff
@@ -70,7 +70,8 @@ neuron aeif_cond_alpha_implicit:
         # input decl post
 
     # output pre
-    output: spike # output in
+    output:
+        spike # output in
     # output post
```
clinssen, line 75: didn't we get rid of post comments?

clinssen:
Looks good, cheers! Could you create a follow-up PR fixing the URLs involving your username (pnbabu -> nest)?

clinssen (approved):

jougs:
@clinssen, @pnbabu: This is probably the wrong place to discuss this, but isn't the `output: spike` block in 100% of cases redundant with the specification `emit_spike()` in the `update` block? I would rather remove the former completely and introduce new functions `emit_rate()`, `emit_current()`, `emit_whatever()` if those become supported in the future and not have the redundancy. What is your opinion on this?

PR #897, opened 2023-04-17
Fix ternary operator handling

clinssen:
The sympy Piecewise() function was not being converted into C++ (cond ? if_true : if_false) syntax properly. This PR fixes the problem (which showed up in the Hill-Tononi model) as well as changing the lexer/parser error strategy to "bail", so this error cannot pass unnoticed in the future.


pnbabu (approved):
Thanks for fixing this! I wonder why this wasn't caught before even if we have a test with the model `BiexponentialPostSynapticResponse.nestml` that has the ternary operator.
The changes look good to me.

clinssen:
The ternary operators there are not in the ODEs or inline expressions themselves.

This issue will be covered by the updated integration tests in #879 (comparing ``hill_tononi.nestml`` to the NEST builtin ``ht_neuron``).

PR #901, opened 2023-05-07
Fix spike buffer origin in GSL stepping function

### LINKED ISSUE #900: Incorrect reference of spike input ports in the result of a numeric solver
When an ODE in the equations block refers to a spiking input port and has a numeric solver, the spiking input buffer is referenced incorrectly in the generated code, thus resulting in a compilation error.

NESTML code:
```
neuron ex_neuron:
  ...
  equations:
    foo' = (-foo/tau * spikes) + ...

  input:
    spikes real <- spike
```

The generated code:
```
extern "C" inline int ex_neuron_dynamics(double, const double ode_state[], double f[], void* pnode)
{
...
f[State_::foo] = ... * node.B_.spike_inputs_grid_sum_[SPIKES - MIN_SPIKE_RECEPTOR] + ....;
...
}
```
The `SPIKES` and `MIN_SPIKE_RECEPTOR` must be referenced with the right scope, in this case, the neuron name `ex_neuron::`

clinssen:
Fixes #900.

@pnbabu: I did not do any functional testing, could you double-check this? Thanks!

### DIFF
--- a/pynestml/codegeneration/printers/gsl_variable_printer.py
+++ b/pynestml/codegeneration/printers/gsl_variable_printer.py
@@ -87,6 +87,6 @@ class GSLVariablePrinter(CppVariablePrinter):
                 vector_parameter = ASTUtils.get_numeric_vector_size(variable_symbol)
                 var_name = var_name + "_" + str(vector_parameter)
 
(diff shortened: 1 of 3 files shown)


pnbabu (approved):
Looks good to me. Thank you so much for taking care of this!

clinssen:
Thanks for the review!

PR #905, opened 2023-05-10
Fix and add test for second-order integration in neuromodulated synapse

clinssen:


pynestml/transformers/synapse_post_neuron_transformer.py
```diff
@@ -277,6 +276,15 @@ def transform_neuron_synapse_pair_(self, neuron, synapse):
                                              for neuron_state_var in syn_to_neuron_state_vars
                                              if new_synapse.get_kernel_by_name(neuron_state_var) is None]
 
+        # all state variables that will be moved from synapse to neuron
+        syn_to_neuron_state_vars = []
```
pnbabu, line 280: The same variable is assigned in line 267. Why is it initialized here again?

clinssen: It's first assigned the variables from an initial scan of the model, then a recursive scan is done to find dependent variables, then it gets overwritten in a way that includes the newly found variables.

tests/nest_tests/resources/dopa_synapse_second_order.nestml
```diff
@@ -0,0 +1,57 @@
...
+        dopa_rate real = 0.
+        dopa_rate_d real = 0.
+
+    parameters:
+        tau_dopa ms = 100 ms
+        d ms = 1 ms    @nest::delay
+
+    equations:
+        dopa_rate' = dopa_rate_d / ms
+        dopa_rate_d' = -dopa_rate / tau_dopa**2 * ms - 2 * dopa_rate_d / tau_dopa
```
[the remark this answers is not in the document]
clinssen: This is just the alpha kernel dynamics as a simple example.

pnbabu (approved):

PR #918, opened 2023-06-15
JIT implementation requirements for NESTML

med-ayssar:


pynestml/meta_model/ast_neuron_or_synapse_body.py
```diff
@@ -68,7 +69,7 @@ def clone(self):
         dup = ASTNeuronOrSynapseBody(body_elements=body_elements_dup,
                                      # ASTNode common attriutes:
                                      source_position=self.source_position,
-                                     scope=self.scope,
+                                     scope=self.scope.clone(),
```
clinssen, line 72: Why does this need to be deepcopied? Can't we run ASTSymbolTableBuilder on the ASTNode after the clone?

PR #921, opened 2023-06-22
Replace NEST typedefs in templates with plain C++ datatypes

pnbabu:
This PR is a consequence of the NEST PR: https://github.com/nest/nest-simulator/pull/2618

PR #924, opened 2023-06-26
Modify NEST version detection

pnbabu:
#921 should be merged before this.


clinssen:
Versions for older NEST releases (<3.5) seem to come out as git strings, e.g. for 3.1:
```
[23,GLOBAL, INFO]: The NEST Simulator version was automatically detected as: vHEAD@512022e
```
I don't know why the tests are all passing. The version check in the Jinja templates needs something of the form "v3.1". We could just leave the older version detection as-is and add a new option for everything 3.5 and following.

clinssen (approved):

PR #929, opened 2023-07-07
Add ceil, floor and round functions

clinssen:


tests/nest_tests/test_nest_math_function.py
```diff
@@ -56,7 +56,10 @@ def test_math_function(self):
         log10_state_specifier = "log10_state"
         erf_state_specifier = "erf_state"
         erfc_state_specifier = "erfc_state"
-        nest.SetStatus(mm, {"record_from": ["x", ln_state_specifier, log10_state_specifier, erf_state_specifier, erfc_state_specifier]})
+        ceil_state_specifier = "ceil_state"
+        floor_state_specifier = "floor_state"
+        round_state_specifier = "round_state"
+        nest.SetStatus(mm, {"record_from": ["x", ln_state_specifier, log10_state_specifier, erf_state_specifier, erfc_state_specifier, ceil_state_specifier, floor_state_specifier, round_state_specifier]})
```
jougs, line 62: I don't quite understand this line. Maybe others feel the same and a comment would help? Thanks!

jougs (approved):
Many thanks for this nice addition. It looks good to me.

Can you please also quickly reply to https://www.nest-initiative.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/ZAMPB3FTGHJANSXNE5ZA6YNE6OFJSVEJ/ when this gets merged?

tests/nest_tests/test_nest_math_function.py
pnbabu: I think the names of the recordables are assigned to variables and they are specified in the `SetStatus()` instead of directly providing their names as literal strings to the function. I also think that the former is not necessary.

jougs: Ah! Thanks for the clarification. Maybe it would be better to remove the variable assignments and use the specifiers directly?

PR #931, opened 2023-07-10
Fix third-factor plasticity buffering and add Bouhadjar network tutorial

clinssen:
Should be merged after #879.

## Todo
- [x] Extend and fix issue with third factor plasticity test: it should also be using a buffer for postsynaptic values, alike to the Clopath neuron's buffer for postsynaptic V_m (and derived) values
- [x] <s>Add coco to check that when specifying ports and variable names as ``post_ports`` code generation option, these variable names and port names exist; give a friendly error message.</s> -- Spun off to #1009.
- [x] <s>Fix issue with resolution()/get_parent() being very slow</s> -- Spun off to #1024
- [ ] Clean up


clinssen:
Superseded by #1027 and #1026.

PR #941, opened 2023-08-08
Change type() check to isinstance()

clinssen:
Fix new pycodestyle guidelines on type checking using ``isinstance()`` rather than ``type()``.


pnbabu (approved):
Thanks!

PR #950, opened 2023-09-07
Fix power operator and `elif` functions in NESTML

pnbabu:


pynestml/visitors/ast_power_visitor.py
```diff
@@ -22,8 +22,13 @@
 """
 rhs : <assoc=right> left=rhs powOp='**' right=rhs
 """
+from pynestml.codegeneration.nest_unit_converter import NESTUnitConverter
+
+from pynestml.symbols.predefined_units import PredefinedUnits
```
clinssen, line 27: could you please alphabetize the imports?

pynestml/visitors/ast_power_visitor.py
```diff
@@ -71,20 +76,49 @@ def calculate_numeric_value(self, expr):
         :rtype: Either
         """
         # TODO write tests for this by PTraeder
```
clinssen, line 78: I think this can now be removed :)

clinssen (changes requested):
Many thanks for the fix! Just some very small points.

clinssen (approved):
Thank you!

PR #953, opened 2023-09-12
Add/fix checks that inline expressions, parameters and internals cannot be assigned to

clinssen:


tests/invalid/CoCoInternalAssignedOutsideBlock.nestml
```diff
@@ -0,0 +1,40 @@
+"""
+CoCoInternalAssignedOutsideBlock.nestml
```
pnbabu, line 2: I think we should do this for the inline expression coco as well. I just found that `CoCoAssignmentToInlineExpression.nestml` is not being called in the `coco_test.py`. According to the new coco, this file should fail.

tests/nest_tests/third_factor_stdp_synapse_test.py
```diff
@@ -304,25 +225,23 @@ def run_synapse_test(self, neuron_model_name,
...
         # verify
         MAX_ABS_ERROR = 1E-6
-        idx = np.where(np.abs(third_factor_trace) < 1E-3)[0]  # find where third_factor_place is (almost) zero
+        idx = np.where(np.abs(third_factor_trace) < 1E-3)[0]  # find where third_factor_trace is (almost) zero
         times_dw_should_be_zero = timevec[idx]
-        for time_dw_should_be_zero in times_dw_should_be_zero:
+        for time_dw_should_be_zero in times_dw_should_be_zero[:-1]:
             _idx = np.argmin((time_dw_should_be_zero - np.array(t_hist))**2)
-            assert np.abs(np.diff(w_hist)[_idx]) < MAX_ABS_ERROR
+            assert np.abs(np.diff(w_hist)[_idx + 1]) < MAX_ABS_ERROR
```
pnbabu, line 245: This test is failing for my PR as well. Why does this need to be changed? Is it because of a change in NEST upstream?

clinssen: I think it's a fluke, due to the fact that our test timing is randomly generated in that test and different for every run, coupled with an off-by-one bug on the actual check.

pnbabu: But the test is consistently failing for the other PRs #950 and #954.

tests/nest_tests/test_ignore_and_fire.py
```diff
@@ -59,30 +59,33 @@ def setUp(self):
...
+        post_neuron.phase_steps = 5
         pre_sr = nest.Create("spike_recorder")
         post_sr = nest.Create("spike_recorder")
         nest.Connect(pre_neuron, pre_sr)
         nest.Connect(post_neuron, post_sr)
 
         nest.Connect(pre_neuron, post_neuron, syn_spec={"synapse_model": self.synapse_model_name})
 
-        nest.Simulate(sim_time)
+        nest.Simulate(sim_time // 2)
```
pnbabu, line 87: Could you add a comment here as to why the simulation is split into two?

pnbabu (changes requested):

pnbabu (approved):
Looks good to me!

clinssen:
Thank you!

PR #956, opened 2023-09-15
Fix third factor STDP synapse test

clinssen:


pnbabu (commented):
The changes look good to me, expect that the test is failing for NEST 2.

pnbabu (approved):
Looks good, thanks for fixing this!

PR #958, opened 2023-09-16
Fix third factor STDP synapse test

clinssen:
Reprise of #956.

Tighter check (1E-12 instead of 1E-5) and clamp I_dend to zero in addition to letting it slowly decay.

### DIFF
--- a/tests/nest_tests/third_factor_stdp_synapse_test.py
+++ b/tests/nest_tests/third_factor_stdp_synapse_test.py
@@ -249,4 +249,3 @@ class NestThirdFactorSTDPSynapseTest(unittest.TestCase):
             np.testing.assert_allclose(0., np.abs(w_hist[_idx] - w_hist[_idx + 1]))   # make sure that weight does not change appreciably
 
         assert np.any(np.abs(np.array(w_hist) - 1) > 0.), "No change in the weight!"
-        0/0
--- a/tests/nest_tests/third_factor_stdp_synapse_test.py
+++ b/tests/nest_tests/third_factor_stdp_synapse_test.py
@@ -167,8 +167,11 @@ class NestThirdFactorSTDPSynapseTest(unittest.TestCase):
                 state = 1
             if t > 2 * sim_time / 6 and state == 1:
                 nest.SetStatus(post_neuron, {"I_dend": 1.})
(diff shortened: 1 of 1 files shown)


tests/nest_tests/third_factor_stdp_synapse_test.py
```diff
@@ -167,8 +161,11 @@ def run_synapse_test(self, neuron_model_name,
                 state = 1
             if t > 2 * sim_time / 6 and state == 1:
                 nest.SetStatus(post_neuron, {"I_dend": 1.})
-            if t > 2 * sim_time / 3. and state == 1:
+            if t > 3 * sim_time / 6. and state == 1:
```
pnbabu, line 164: Could you maybe comment what the states represent here?

tests/nest_tests/third_factor_stdp_synapse_test.py
```diff
@@ -230,19 +226,18 @@ def run_synapse_test(self, neuron_model_name,
...
             fig.savefig("/tmp/stdp_third_factor_synapse_test" + fname_snip + ".png", dpi=300)
 
         # verify
-        MAX_ABS_ERROR = 1E-5
-        idx = np.where(np.abs(third_factor_trace) < MAX_ABS_ERROR)[0]  # find where third_factor_trace is (almost) zero
+        idx = np.where(np.abs(third_factor_trace) < 1E-12)[0]  # find where third_factor_trace is (almost) zero
         times_dw_should_be_zero = timevec[idx]
         assert len(times_dw_should_be_zero) > 0  # make sure we have > 0 datapoints to check
-        for time_dw_should_be_zero in times_dw_should_be_zero[:-1]:
+        for time_dw_should_be_zero in times_dw_should_be_zero[1:]:
```
pnbabu, line 238: Why is the first element from `times_dw_should_be_zero` ignored?

tests/nest_tests/third_factor_stdp_synapse_test.py
clinssen: It's a sequence of statements to apply an externally injected current (and stop applying it) at given times. I don't think the states can be given meaningful names, hence the use of integers.

pnbabu (approved):

PR #960, opened 2023-09-21
Add NEST 3.6.0 to CI

pnbabu:


clinssen:
The version detection method should be updated; 3.6-rc1 is still being detected as 3.4.

clinssen (approved):
Thanks! I am happy with this, but let's not merge yet, but wait for 3.6 to actually be released  so we can drop the "_rc1".

clinssen (approved):

PR #963, opened 2023-09-22
Fix docs for convolve

pnbabu:


clinssen:
Do you think we should change the documentation, or the actual naming (and change "__X__" to "_conv_" throughout the codebase)?

pnbabu:
> Do you think we should change the documentation, or the actual naming (and change "**X**" to "_conv_" throughout the codebase)?

I think having `__conv__` in the variable name is perhaps more descriptive and `__X__` is concise as it is now. I do not actually have a preference. Do you have one?

clinssen:
Shall we replace __X__ with __conv__? It's only three extra characters and much more descriptive.

PR #968, opened 2023-10-07
Extend static synapse test

clinssen:
Simplify: remove multimeter and plot.

Add noisy_synapse to the test.

Add homogeneous parameters test.

Refactoring: remove ``Either`` class for types; this is possible because we now favour erroring out early rather than accumulating all errors and trying to continue processing.


pnbabu (approved):
Looks good to me. I'm fine to merge this after fixing the test for NEST 2.20.

clinssen:
Thank you for the review!

PR #969, opened 2023-10-07
Fix convolutions integration and enhance integrate_odes() to integrate specific ODEs

### LINKED ISSUE #440: Integrate specific ODEs with `integrate_odes()`
Feature request: specify which ODEs to integrate via parameters to the `integrate_odes()` function. Note that there might be interdependencies between the ODEs that need to be respected/checked.

clinssen:
Add feature to integrate specific ODEs with integrate_odes().

Always integrate convolutions in time, independent of whether integrate_odes() is called, and in the case of NEST Simulator, are always assumed to be processed after the subthreshold dynamics but before the condition checks; hence the observed value will always be the one "just after" the instantaneous change due to the spike event.

Fixes #440.

- [ ] NEST code generator
- [x] Python code generator


clinssen:
Sorry, this turns out to be impossible to untie from #879.

PR #973, opened 2023-10-18
Add plastic synapse weight sign test

clinssen:
With thanks to @tomtetzlaff!

### DIFF
--- a/models/synapses/stdp_synapse.nestml
+++ b/models/synapses/stdp_synapse.nestml
@@ -33,22 +33,26 @@ References
        Stable Hebbian learning from spike timing-dependent
        plasticity, Journal of Neuroscience, 20:23,8812--8821
 """
-synapse stdp:
+synapse stdp_synapse:
     state:
-        w real = 1.    @nest::weight   # Synaptic weight
+        w real = 1.    @nest::weight   # Synaptic weight (> 0 for excitatory and < 0 for inhibitory synapses)
         pre_trace real = 0.
         post_trace real = 0.
 
     parameters:
         d ms = 1 ms  @nest::delay   # Synaptic transmission delay
-        lambda real = .01
-        tau_tr_pre ms = 20 ms
-        tau_tr_post ms = 20 ms
-        alpha real = 1
-        mu_plus real = 1
-        mu_minus real = 1
-        Wmax real = 100.
-        Wmin real = 0.
+        lambda      real =  0.01  # (dimensionless) learning rate for causal updates
+        alpha       real = 1      # relative learning rate for acausal firing
+        tau_tr_pre  ms   = 20 ms  # time constant of presynaptic trace
+        tau_tr_post ms   = 20 ms  # time constant of postsynaptic trace
+        mu_plus     real = 1      # weight dependence exponent for causal updates
+        mu_minus    real = 1      # weight dependence exponent for acausal updates
+
+        Wmax        real = 100.     # maximum absolute value of synaptic weight
+        Wmin        real = 0.       # minimum absolute value of synaptic weight
+
+    internals:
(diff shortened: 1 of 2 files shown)


clinssen:
Indeed, inhibitory STDP is an absolutely crucial to model. However, I thought that this could be captured by the postsynaptic neuron having a separate input port for inhibitory spikes, which are coming in with weight >=0, and are subsequently handled with the appropriate sign on the postsynaptic side.

However, if the STDP synapse model is to support postsynaptic neuron models that have only one spiking input port and differentiate excitatory from inhibitory spikes on the basis of the sign of the weight, then indeed an adjustment is needed.

About the implementation: if all is as it should be, then internals should not be allowed to depend on state variables (only on parameters and other internals). Thus the piece of code

```
internals:
    w_sign real = w/abs(w)   # sign of synaptic weight
```

would result in an error. This may also have an issue with division by zero when the weight becomes equal to 0.

We could make ``w_sign`` a user-settable parameter, and additionally verify that so long as the weight is not equal to zero, it holds that ``w_sign==sign(w)`` as a sanity check.

Alternatively, we could allow a signed Wmin and Wmax to indicate the sign of the weight.

clinssen:
Thanks for the comments!

> An additional parameter `w_sign` would be redundant if the weights carry a sign, right? In principle, one could extract the sign just from the initial weight, rather than each time the synapse is updated.

Yes, but in NESTML this runs into the problem that principally, parameter initial values cannot depend on state initial values. This is how the initialisation order is defined right now in the language. So an extra parameter ``W_sign`` would have to be introduced (or we would have to re-consider the initialisation sequence in the language).

> Defining weights to be always positive and then adding the sign of the synapse on the postsynaptic side via the input port is actually a quite nice idea. It's also closer to biology where "weights" primarily correspond to (effective) "conductances" (which are always >0) and the sign is determined by the reversal potential (see above).

Cool! Does that mean that you would be okay with the PR as it is now? It enforces all STDP-type synapse weights to be >= 0 at all times (and the same for Wmin and Wmax).

> Do we actually need to account for cases where the sign of the weight may change during the learning? This is certainly not what we have in nature, but I'm thinking of studies comparing the performance of a network respecting Dale's principle with an alternative (machine learning) solution which doesn't care about nature.

That would be an interesting use-case to research, but at the moment I guess it's exactly the thing we want to prevent from happening inadvertedly. Perhaps we could add a short paragraph mentioning Dale's law in the STDP synapse model docstring?

tomtetzlaff:
> Cool! Does that mean that you would be okay with the PR as it is now? It enforces all STDP-type synapse weights to be >= 0 at all times (and the same for Wmin and Wmax).

Yes, I think that should be fine. But I would suggest to add an example illustrating this strategy, i.e., how to to implement STDP for inhibitory synapses.

> That would be an interesting use-case to research, but at the moment I guess it's exactly the thing we want to prevent from happening inadvertedly. 

I agree that with STDP synapses this usecase will probably never happen, as most users employing an STDP model would certainly assume that synaptic weights cannot simply switch sign. But we may have this issue with other plasticity types. For eprob, for example, @akorgor is currently investigating exactly this, i.e., what the effect of the "fixed-sign" constraint is on the learning performance. But this is a different type of plasticity, so it may not be relevant for this PR, is it?

> Perhaps we could add a short paragraph mentioning Dale's law in the STDP synapse model docstring?

I wouldn't refer to this as "Dale's law", but rather to the fact that an individual synapse cannot simply switch from one type to another, say, from glutamatergic to GABAergic. Dale's principle is an even stronger statement on the ensemble of all outgoing synapses of a given neuron (all of them having the same sign).

tomtetzlaff (approved):
This looks very good. Thanks a lot!

PR #975, opened 2023-10-30
Add non-default install path option to ``NESTCodeGeneratorUtils.generate_code_for()``

### LINKED ISSUE #972: Add non-default install path option to ``NESTCodeGeneratorUtils.generate_code_for()``
On the EBRAINS image, NESTML tries to copy the generated .so user extension module into the NEST Simulator install directory, which is not writeable by the user. 

Add a feature to install this to another location, or use mkstemp to automatically find a suitable location.

clinssen:
Fixes #972.


pnbabu (approved):
Thanks!

PR #994, opened 2023-11-30
Fix readthedocs search

clinssen:
Search on https://nestml.readthedocs.org/ broke spontaneously with the following error message:

```
Uncaught ReferenceError: jQuery is not defined
```

Fix: update version numbers to latest.


clinssen:
Looks like this was fixed by #1034.

PR #998, opened 2023-12-05
Instruction Update re Basic Ubuntu on WSL

pippowell:
Added the cmake package to the conda install command. The cmake package is not installed by default and will result in errors if missing (confirmed on MacOS and Linux). 

Also added information relevant to installing on Windows Linux Subsystem installations of Linux.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -91,7 +91,7 @@ Please make sure to have the latest conda version installed and to create a new 
 
    conda create --name wnestml
    conda activate wnestml
-   conda install -c conda-forge nest-simulator ipython cxx-compiler pyqt wxpython boost boost-cpp libboost cmake
+   conda install -c conda-forge nest-simulator ipython cxx-compiler pyqt wxpython boost boost-cpp libboost cmake make
    pip install nestml
 
 Test the path to ``c++``:
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -83,7 +83,9 @@ After installation, correct operation can be tested by:
 Anaconda installation
 ---------------------
 
-In preparation, `create a conda environment with NEST <https://nest-simulator.readthedocs.io/en/stable/installation/index.html>`_, and install some additional dependencies. Note that if you are working with a basic version of Ubuntu or another Linux distribution (ex. a basic download for use in the Windows Linux Subsystem), you also need to include the ``make`` package in the install command. The ``cmake`` package will not function without this package installed (it is by default on full versions of Linux distributions). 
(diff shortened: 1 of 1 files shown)


pippowell:
@clinssen In general I would say go ahead and include it, since it should just recognize that "make" is already installed and simply install the other packages. The caveat though is that conda is notoriously problematic with environment solving the more packages you include that it has to check dependencies and conflicts for. The "cxx-compiler" and "wxpython" packages were already making the installation tricky when I set up my environment and the addition of the boost packages to the list (which was done Monday) seemed to already make matters worse. Adding "make" could potentially add to the issue and make it even more likely that fresh installations will hang on solving. But if there is a switch to a new installation scheme, such as mamba etc., perhaps that would work. Conda switched to the mamba background installer in the last couple of updates and when I created a Miniconda environment and used the latest conda with that background, it managed to install the previous list (everything including boost, but not cmake or make) without issue as long as you entered the entire list on the first installation. It still didn't like installing "wxpython" after the fact though, which I had tried in an earlier installation attempt. 

Perhaps if you add "make" and make an additional note that new users should make sure they have the latest conda installed and that they create their new environment with the command on the website, i.e. installing all packages together at the start versus installing one by one? Some may still have issues, but it would be a good place for them to start and they can troubleshoot from there.

doc/installation.rst
```diff
@@ -83,13 +83,13 @@ After installation, correct operation can be tested by:
 Anaconda installation
 ---------------------
 
-In preparation, `create a conda environment with NEST <https://nest-simulator.readthedocs.io/en/stable/installation/index.html>`_, and install some additional dependencies:
+In preparation, `create a conda environment with NEST <https://nest-simulator.readthedocs.io/en/stable/installation/index.html>`_, and install some additional dependencies. Note that if you are working with a basic version of Ubuntu or another Linux distribution (ex. a basic download for use in the Windows Linux Subsystem), you also need to include the ``make`` package in the install command. The ``cmake`` package will not function without this package installed (it is by default on full versions of Linux distributions).
```
clinssen, line 86:
```suggestion
In preparation, `create a conda environment with NEST <https://nest-simulator.readthedocs.io/en/stable/installation/index.html>`_, and install some additional dependencies.

Please make sure to have the latest conda version installed and to create a new environment with the command below, i.e. installing all packages together at the start versus installing one by one.
```

clinssen:
@pippowell : thanks for your feedback! Could you review the proposed changes?

pippowell:
@clinssen Looks good. The new list should (fingers crossed) create a functional environment for NEST regardless of platform. With all packages installing at once, nothing worse should happen than conda noting that one of the packages is already there and the user doesn't have to give anything a second thought. If anyone does encounter issues, they'll at least have a good foundation to start from. I'd say these changes can be committed now. 👍

PR #1002, opened 2024-01-08
Change model registration for NEST target

pnbabu:
Changes in model registration and module unloading/reloading corresponding to the NEST PR https://github.com/nest/nest-simulator/pull/3103


tests/nest_tests/stdp_window_test.py
```diff
@@ -75,6 +74,10 @@ def run_stdp_network(pre_spike_time, post_spike_time,
     nest.set_verbosity("M_ALL")
 
     nest.ResetKernel()
+    try:
+        nest.Install("nestmlmodule")
+    except Exception:
+        pass
```
clinssen, line 80:
```suggestion
        # ResetKernel() does not unload modules for NEST Simulator < v3.7; ignore exception if module is already loaded on earlier versions
        pass
```

clinssen (changes requested):

pynestml/codegeneration/nest_code_generator_utils.py
```diff
@@ -18,12 +18,11 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
-
+import uuid
```
clinssen, line 21: please fix import order

pynestml/codegeneration/nest_code_generator_utils.py
```diff
@@ -81,59 +80,56 @@ def generate_code_for(cls,
...
                 nestml_neuron_model = nestml_model_file.read()
 
         # update neuron model name inside the file
-        neuron_model_name_orig = re.findall(r"neuron\ [^:\s]*:", nestml_neuron_model)[0][7:-1]
-        neuron_model_name_uniq = neuron_model_name_orig + uniq_id
-        nestml_model = re.sub(r"neuron\ [^:\s]*:",
-                              "neuron " + neuron_model_name_uniq + ":", nestml_neuron_model)
-        neuron_uniq_fn = neuron_model_name_uniq + ".nestml"
-        with open(neuron_uniq_fn, "w") as f:
+        neuron_model_name = re.findall(r"neuron [^:\s]*:", nestml_neuron_model)[0][7:-1]
```
clinssen, line 89: lines 89-93 are no longer necessary because we're not appending the uuid any longer, right?

pynestml/codegeneration/nest_code_generator_utils.py
```diff
@@ -81,59 +80,56 @@ def generate_code_for(cls,
...
+        nestml_model = re.sub(r"neuron [^:\s]*:",
+                              "neuron " + neuron_model_name + ":", nestml_neuron_model)
+        neuron_fn = neuron_model_name + ".nestml"
+        with open(neuron_fn, "w") as f:
             print(nestml_model, file=f)
 
+        input_fns = [neuron_fn]
+        codegen_opts = {"neuron_parent_class": "StructuralPlasticityNode",
+                        "neuron_parent_class_include": "structural_plasticity_node.h"}
+        mangled_neuron_name = neuron_model_name + "_nestml"
```
clinssen, line 99: same for this line, mangled name is by this point still always equal to the original name, right?

clinssen (commented):
Great, thanks for all the work! Just a few small comments.

pynestml/codegeneration/nest_code_generator_utils.py
pnbabu: I have removed the string substitution but have retained extracting the neuron name since we return the neuron name from this function.

pynestml/codegeneration/nest_code_generator_utils.py
pnbabu: I retained this because we generate the code with the suffix `_nestml` and since we return the neuron name from this function, I append this to the extracted neuron name.

clinssen (approved):

PR #1005, opened 2024-02-22
Add GL model tutorial notebook

clinssen:


clinssen:
From meeting 2024-04-02: we should add a figure that shows how picking various simulation resolutions (between say 0.01 and 10 ms) affects the number of emitted spikes in an extended simulation span.

clinssen:
See here for further notes and comments: https://fz-juelich.sciebo.de/apps/files/?dir=/stochastic_model

Feedback from Renan (many thanks!):

1. As we discussed before, the interpretation of Phi may vary. For instance, the ref [[1]](https://www.nature.com/articles/srep35831) call it "firing function", and in their section "The Model" they explicitly say that Phi(V) in their context is a probability that grows monotonically in the limits of 0 to 1, then in this case Phi(V) is not the firing rate. I'm bringing this context to say that the biophysical parameters in the notebook were estimated using this same interpretation.
   • The reference where we estimated the Phi(U) from experimental data (Figure 5 of REF below) and also tested the reliability of the single neuron (Figure 1 of REF below) is this one:
     • Lima, V., Pena, R. F. O., Shimoura, R. O., Kamiji, N. L., Ceballos, C. C., Borges, F. S., Higa, G. S. V., de Pasquale, R., & Roque, A. C. (2021). Modeling and characterizing stochastic neurons based on in vitro voltage-dependent spike probability functions. The European Physical Journal Special Topics 2021 230:14, 230(14), 2963–2972. https://doi.org/10.1140/EPJS/S11734-021-00160-7
   • An experimental reference for the reliability figure is the Figure 1 from: 
     • Mainen, Z. F., & Sejnowski, T. J. (1995). Reliability of Spike Timing in Neocortical Neurons. Science, 268(5216), 1503–1506. https://doi.org/10.1126/science.7770778
 2. If considering Phi(V) as a probability of spike in function of the membrane potential, then I'm not sure if the conversion in nestml_gl_exp_model fits to this case: if random_uniform(0, 1) <= 1E-3 * resolution() * Phi(U). In my understanding, this is fine if Phi(U) is the firing rate (which I think is the case for the nestml_gl_ca_neuron_model), but maybe not for this one. 
   • This definition makes the model similar to the [Escape noise model](https://lcnwww.epfl.ch/gerstner/SPNM/node35.html#noise-2a), where here comes an example Markus mentioned about a similar existing model. If you check the link, Figure 5.8, shows equivalent Phi(V) curves for different discretizations dt, where you can see it affects Phi(U). So, back to the notebook, I understand that the nestml_gl_exp_model should not have the 1E-3 * resolution() multiplying Phi(U).
 3. A more technical issue, the reliability experiment has to be adjusted a bit. 
   • For instance, the evaluate_neuron has more arguments than needed for the example. I can adjust it later. 
   • Additionally, the idea of the experiment is testing multiple trials under constant current and frozen noise. I used a fixed poisson generator as a workaround, but ideally we should use the exact same noise_generator (frozen noise) across different trials. Is there a way to set a single noise_generator and apply it for multiple neurons? I mean, it is possible of course, but each neuron receives a different noise, ideally it should receive the exact same. If there is a smart way to do it, we should implement and replace the poisson generator.

shimoura:
Thanks a lot @clinssen for generating this notebook and already addressing many points!
I would like to mention some missing issues below. 

## Regarding the first part (Experiments with Biophysical units)

> 1. As we discussed before, the interpretation of Phi may vary. For instance, the ref [[1]](https://www.nature.com/articles/srep35831) call it "firing function", and in their section "The Model" they explicitly say that Phi(V) in their context is a probability that grows monotonically in the limits of 0 to 1, then in this case Phi(V) is not the firing rate. I'm bringing this context to say that the biophysical parameters in the notebook were estimated using this same interpretation.
>        • The reference where we estimated the Phi(U) from experimental data (Figure 5 of REF below) and also tested the reliability of the single neuron (Figure 1 of REF below) is this one:
>        • Lima, V., Pena, R. F. O., Shimoura, R. O., Kamiji, N. L., Ceballos, C. C., Borges, F. S., Higa, G. S. V., de Pasquale, R., & Roque, A. C. (2021). Modeling and characterizing stochastic neurons based on in vitro voltage-dependent spike probability functions. The European Physical Journal Special Topics 2021 230:14, 230(14), 2963–2972. https://doi.org/10.1140/EPJS/S11734-021-00160-7
>        • An experimental reference for the reliability figure is Figure 1 from:
>        • Mainen, Z. F., & Sejnowski, T. J. (1995). Reliability of Spike Timing in Neocortical Neurons. Science, 268(5216), 1503–1506. https://doi.org/10.1126/science.7770778

This point here was not addressed. The "GL model with biophysical units" follows the interpretation of phi being the spiking probability (It is the opposite of the current description in the "Firing rate" subsection). So, $$\phi(V_m)$$ goes from 0 to 1 and represents the spike probability depending on the membrane potential. Additionally, the parameters for the biophysical units were estimated from experimental data in the paper I mentioned above, we could add the ref.

Summary of suggestions for this part:
- Add the reference in Biophysical units: Lima, V., Pena, R. F. O., Shimoura, R. O., Kamiji, N. L., Ceballos, C. C., Borges, F. S., Higa, G. S. V., de Pasquale, R., & Roque, A. C. (2021). Modeling and characterizing stochastic neurons based on in vitro voltage-dependent spike probability functions. The European Physical Journal Special Topics 2021 230:14, 230(14), 2963–2972. https://doi.org/10.1140/EPJS/S11734-021-00160-7
- Add threshold maximum value of $$\phi(V_m)$$ to 1.0, for the "nestml_gl_exp_model"
- The subsection should be renamed to "Firing probability"
        - The description should be adjusted to match it as well
        - In the figure: rename the ylabel as "Firing probability" and xlabel as "V [mV]"; set the y-lim to [0, 1]; 
- In "Computing reliability of the single neuron model", replace all "Poissonian input" with "frozen Poissonian input"
       - In the figure, adjust the Firing rate unit to "spikes/s" instead of "Hz"

## Regarding the second part (GL with calcium dynamics)

In this case, I guess the interpretation of $$\phi(U)$$ as being the firing rate is fine.
Something strange happens here when I run this second part. Specifically, in this line:
`log = measure_postsynaptic_response(neuron_model_name,
                                    synapse_model_name,
                                    module_name,
                                    V_m_specifier="U")`
Here the code keeps running forever (I waited for almost an hour, so I interrupted ). If I stop this cell after a while and run it again, then it runs fast and plot the expected graph. Do you know what could be the explanation? 
In addition to that, I tested skipping the line above to generate the next graphs, sometimes they give the expected results sometimes not. I guess it is related to the random seed because probably the result is chaotic.

Other suggestions:
- Set random seed for nest simulations
- Rename all figures with units "[Hz]" by "[spikes/s]"

clinssen:
### [WORK IN PROGRESS!]


> 1. As we discussed before, the interpretation of Phi may vary. For instance, the ref [[1]](https://www.nature.com/articles/srep35831) call it "firing function", and in their section "The Model" they explicitly say that Phi(V) in their context is a probability that grows monotonically in the limits of 0 to 1, then in this case Phi(V) is not the firing rate. I'm bringing this context to say that the biophysical parameters in the notebook were estimated using this same interpretation.

Fixed the definition.

>        • The reference where we estimated the Phi(U) from experimental data (Figure 5 of REF below) and also tested the reliability of the single neuron (Figure 1 of REF below) is this one:
>        • Lima, V., Pena, R. F. O., Shimoura, R. O., Kamiji, N. L., Ceballos, C. C., Borges, F. S., Higa, G. S. V., de Pasquale, R., & Roque, A. C. (2021). Modeling and characterizing stochastic neurons based on in vitro voltage-dependent spike probability functions. The European Physical Journal Special Topics 2021 230:14, 230(14), 2963–2972. https://doi.org/10.1140/EPJS/S11734-021-00160-7
>        • An experimental reference for the reliability figure is Figure 1 from:
>        • Mainen, Z. F., & Sejnowski, T. J. (1995). Reliability of Spike Timing in Neocortical Neurons. Science, 268(5216), 1503–1506. https://doi.org/10.1126/science.7770778

Inserted reference in the section "Computing reliability of the single neuron model".

> - Add threshold maximum value of $$\phi(V_m)$$ to 1.0, for the "nestml_gl_exp_model"

not needed as Phi(V) is in the NESTML models assumed to be the firing rate. Definition in the text was updated and a comment was added to the NESTML function code.


> - The subsection should be renamed to "Firing probability"
        - The description should be adjusted to match it as well
        - In the figure: rename the ylabel as "Firing probability" and xlabel as "V [mV]"; set the y-lim to [0, 1]; 
- In "Computing reliability of the single neuron model", replace all "Poissonian input" with "frozen Poissonian input"
       - In the figure, adjust the Firing rate unit to "spikes/s" instead of "Hz"

## Regarding the second part (GL with calcium dynamics)

In this case, I guess the interpretation of $$\phi(U)$$ as being the firing rate is fine.
Something strange happens here when I run this second part. Specifically, in this line:
`log = measure_postsynaptic_response(neuron_model_name,
                                    synapse_model_name,
                                    module_name,
                                    V_m_specifier="U")`
Here the code keeps running forever (I waited for almost an hour, so I interrupted ). If I stop this cell after a while and run it again, then it runs fast and plot the expected graph. Do you know what could be the explanation? 
In addition to that, I tested skipping the line above to generate the next graphs, sometimes they give the expected results sometimes not. I guess it is related to the random seed because probably the result is chaotic.

> Other suggestions:
> - Set random seed for nest simulations

In the sense of, set it explicitly to 42 or set it to the system time in ms?

> - Rename all figures with units "[Hz]" by "[spikes/s]"

Done.

PR #1007, opened 2024-02-27
Add copyright notice to documentation

clinssen:


jessica-mitchell (approved):
Ah this looks like a nice solution! We should do this for nest too. thanks!

PR #1017, opened 2024-03-25
Add dependabot to CI

clinssen:
I spoke with @terhorstd and this might be hard to test before we merge it. Perhaps we can give it a sanity check, merge it, and then evaluate if it is working as expected.


clinssen:
Merging this to see if it works well, we will review and evaluate based on the results.

Issue #1016, opened 2024-03-18
Error with `random_normal()` function


pnbabu:
When the `random_normal()` function is called from the `parameters` or `state` block of the NESTML model, the simulation script with the model throws an error:
```
Assertion failed: (tid < static_cast< size_t >( vp_specific_rngs_.size() )), function get_vp_specific_rng, file random_manager.h, line 166.
```

In the generated code, the parameter with the `random_normal()` function is initialized in the constructor
```
P_.internal_val = ((0) + (0.2) * normal_dev_( nest::get_vp_specific_rng( get_thread() ) )); // as real
```

At this point in the code, the thread ID via the `get_thread()` function is not yet available with the node which results in the error.

clinssen:
Based on discussion at the 2024 Q1 NEST hackathon, we concluded that implementing this in NEST Simulator would be difficult. In this case, we should forbid users from using RNGs in the parameters (or internals, or state) initialisation when generating code for the NEST target. (This should be a target platform-specific coco.)

PR #1018, opened 2024-03-25
Update docs for `random_normal()` and `random_uniform()` functions

pnbabu:


clinssen:
This is more a limitation of the NEST target than of NESTML in general. Other platforms could easily support random number generation during initialisation. I would therefore prefer if we add a context condition check for this that only operates on the NEST target.

clinssen:
Already covered by #1016.

PR #1027, opened 2024-04-19
Fix third-factor plasticity buffering and add third factor plasticity tutorial

clinssen:
Extend and fix issue with third factor plasticity test: it should also be using a buffer for postsynaptic values, alike to the Clopath neuron's buffer for postsynaptic V_m (and derived) values


pnbabu (changes requested):
My comments on the jupyter notebook:
- There is an empty graph between the pre-spikes and post-spikes graph. Please remove that
- Can we use `NestCodeGeneratorUtils.generate_code_for()` instead of `generate_nest_target()` to keep all the notebooks consistent?
- can you add some more comments in the `run_synapse_test()` function?
- Could you add a line or two about how NEST updates weights? This is especially for case 6 where the wright update in NEST differs from reference simulation.

pynestml/utils/ast_utils.py
```diff
@@ -1568,6 +1594,27 @@ def get_state_variable_declaration_by_name(cls, node: ASTModel, var_name: str) -
                         return decl
         return None
 
+    @classmethod
+    def replace_post_moved_variable_names(cls, astnode, post_connected_continuous_input_ports, post_variable_names):
```
pnbabu, line 1598: Can you add docs to this function?

pynestml/transformers/synapse_post_neuron_transformer.py
```diff
@@ -496,7 +529,7 @@ def mark_post_port(_expr=None):
 
         for state_var, alternate_name in zip(post_connected_continuous_input_ports, post_variable_names):
```
pnbabu, line 530: Since `alternate_name` is not used in the loop anymore, you can just loop over the `post_connected_continuous_input_ports`.

pnbabu (changes requested):
Comments on the jupyter notebook:

- The NESTML neuron model name is `iaf_psc_exp_active_dendrite_neuron`, but the kernel used is an alpha kernel. Should the neuron model be renamed?
- Why is `I_syn = 0 pA` on the last line of the model? Should the value of `I_syn` be reset when a spike is emitted?
- The `run_synapse_test` function says the simulation is run for two neurons of type `neuron_model_name`, but the pre-neuron is a `parrot` neuron and the post-neuron is `neuron_model_name`. In principle, the pre-neuron can also be of type `neuron_model_name`, right?
- Isn't running the simulation for each timestep costly? Can we use the `step_current_generator` for setting the `I_dAP` currents instead?
- The y-axis for the weight plot in Case 2 is cluttered. Could you fix that?

clinssen:
Thanks for the detailed review! The new commits should address all of your comments.

> Why is I_syn = 0 pA on the last line of the model? Should the value of I_syn be reset when a spike is emitted?

Indeed, I added this because I thought it would make it simpler to make a neuron spike by generating a spike event with a very large weight.

> Isn't running the simulation for each timestep costly? Can we use the step_current_generator for setting the I_dAP currents instead?

I don´t think it can set the value of the state variable directly. This is just for demonstration purposes on a 2-neuron network, so performance is not so important.

pynestml/transformers/synapse_post_neuron_transformer.py
```diff
@@ -365,6 +365,37 @@ def transform_neuron_synapse_pair_(self, neuron, synapse):
 
         # XXX: TODO
```
pnbabu, line 366: Is this `TODO` comment required?

pnbabu (commented):
Could you generate readthedocs and post the link?

pynestml/transformers/synapse_post_neuron_transformer.py
clinssen: yes, it refers to "collect all the internal parameters", which is something that should indeed still be done!

clinssen:
> Could you generate readthedocs and post the link?

Please see: https://nestml--1027.org.readthedocs.build/en/1027/

doc/running/running_nest.rst
```diff
@@ -218,6 +218,19 @@ Simulation of volume-transmitted neuromodulation in NEST can be done using "volu
...
+
+The implementation corresponds to the event-based update scheme in Fig. 4b of [Stapmanns2021]_. There, the authors observe that the storage and management of such a buffer can be expensive in terms of memory and runtime. In each time step, the value of the current dendritic current (or membrane potential, or other third factor) is appended to the buffer. The maximum length of the buffer depends on the maximum inter-spike interval of any of the presynaptic neurons.
+
+To generate code for synapses with a (postsynaptic) plasticity factor, NESTML needs to be invoked so that it generates code for neuron and synapse together (as described in :ref:`Generating code for plastic synapses`). Additionally, specify the ``"post_ports"`` entry to connect the input port on the synapse with the right variable of the neuron (see :ref:`Generating code`). Passing this as a code generator option facilitates combining models from different sources, where the naming conventions can be different between the neuron and synapse model.
```
pnbabu, line 230: The link to `Generating code` here points to SpiNNaker. Could you fix that?

pnbabu:
Could you also resolve the merge conflicts?

pnbabu (approved):
Thank you for adding this feature!

clinssen:
Cheers!

PR #1042, opened 2024-04-24
Fix indexing of vector input ports

pnbabu:
This PR fixes the indexing of vector input ports when used in the `update` block with indexes that are variables. The vector variables and vector input ports are both indexed starting from `0`.


pynestml/codegeneration/printers/gsl_variable_printer.py
```diff
@@ -82,10 +82,11 @@ def _print_buffer_value(self, variable: ASTVariable) -> str:
         variable_symbol = variable.get_scope().resolve_to_symbol(variable.get_complete_name(), SymbolKind.VARIABLE)
         if variable_symbol.is_spike_input_port():
             var_name = variable_symbol.get_symbol_name().upper()
-            if variable_symbol.get_vector_parameter() is not None:
-                vector_parameter = ASTUtils.get_numeric_vector_size(variable_symbol)
-                var_name = var_name + "_" + str(vector_parameter)
-
+            if variable.has_vector_parameter():
+                if variable.get_vector_parameter().is_variable():
+                    var_name += "_0 + " + variable.get_vector_parameter().get_variable().get_name()
```
clinssen, line 87: why the 0 here?

tests/nest_tests/resources/input_ports_in_loop.nestml
```diff
@@ -0,0 +1,61 @@
...
+        # foo spikes
+        i integer = 0
+        for i in 0 ... (N_buf) step 1:
+            foo_spikes += 2.5 * foo[i] * s
+
+        # spike_buf spikes
+        j integer = 0
+        k integer = 0
+        while j < N_buf and k < N_spikes:
+            my_spikes_ip[k+2] += spike_buf[j] * s
```
clinssen, line 59: Very cool!

pynestml/codegeneration/printers/gsl_variable_printer.py
pnbabu: Because the enum corresponding to the first input port in a vector of input ports will have the `_0` suffixed to the enum's name. For example:

```
input:
  foo[42] <- spike

update:
  j integer = foo[23]
```
The enum for the first input port here would be `FOO_0`. So the indexing in this case would be

```
int j = spike_inputs_buffer[FOO_0 + 23]
```

clinssen (approved):

clinssen:
Could you link the PR to the corresponding issue(s), so they are automatically closed when this PR is merged? Cheers!

PR #1043, opened 2024-04-25
Fix typos and links in documentation pages

clinssen:


doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -10,11 +10,11 @@ Modeling neurons in NESTML
 Writing the NESTML model
 ########################
 
-The top-level element of the model is ``neuron``, followed by a name. All other blocks appear inside of here.
```
pnbabu, line 13: The `neuron` keyword is also present in compartmental models docs. Could you also fix that?

pnbabu (changes requested):

doc/nestml_language/neurons_in_nestml.rst
clinssen: Thanks for noticing that! Fixed.

PR #1049, opened 2024-05-08
Refactor ``ASTNode.get_parent()`` to improve runtime performance

### LINKED ISSUE #1024: ``get_parent()`` is excruciatingly slow
There is a context condition check on the NESTML built-in resolution() function, which needs to find the block that each resolution() function call is enclosed in: the update block, equations, parameters, etc. (In the equations block and in functions, it is not allowed.) To check this coco, there is a visitor which finds each resolution() function call, and then tries to find the parent block of the call by calling ``ast_neuron.get_parent(ast_resolution_func_call)``. It turns out that get_parent() is implemented by hand for each AST node, and does a depth-first search on the entire ast_neuron to find ast_resolution_func_call. This works, in principle. However, we now have a model that has several resolution() function calls, which is resulting in tens of millions of calls to get_parent() on various AST nodes. This is making code generation excruciatingly slow.

The simplest and fastest approach could be to simply add a "parent" attribute to each AST node to create a "doubly linked tree". However, this means that this attribute needs to be passed around in each constructor and the tree should be kept consistent. Could we implement this instead by a new "parent aware" visitor pattern? But would we then assign the parent information to each AST node instance, or to a centralised data structure inside the visitor? And would the get_parent() methods go away on all the AST node classes?

If we use the "parent aware" visitor pattern, all visitors would derive from this ParentAwareVisitor. Would it be the case that for each of their methods (visit_statement(), visit_equation(), etc.), they would need to invoke the corresponding visit_...() method in the ParentAwareVisitor? This means that we would have to explicitly make this call for all derived visitors, e.g.

```
class MyVisitor(ParentAwareVisitor):
    def visit_statement(self, node):
        super().visit_statement(node)   # <-- has to be done explictly to get ParentAwareVisitor to update the stack?!
        # ... things here specific to MyVisitor; can now call node.get_parent() here ...
```

See paragraph 8.2.3 ("Parent Aware Visitor") in the MontiCore reference manual.

clinssen:
Add an ASTParentVisitor which sets an attribute ``parent_`` on each ASTNode on the basis of calls to ``get_children()`` on ASTNodes, making the AST a doubly-linked tree. This is (probably) the fastest and easiest solution, but of course has the disadvantage that information is stored redundantly.

As a test for self-consistency, an assert is added to check that ``self`` is in ``parent_.get_children()`` in each call to ``get_parent()``.

Fixes #1024.

### DIFF
--- a/pynestml/meta_model/ast_declaration.py
+++ b/pynestml/meta_model/ast_declaration.py
@@ -19,12 +19,13 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-from typing import Optional, List
+from typing import Optional, List, Union
 
 from pynestml.meta_model.ast_data_type import ASTDataType
 from pynestml.meta_model.ast_expression import ASTExpression
 from pynestml.meta_model.ast_namespace_decorator import ASTNamespaceDecorator
 from pynestml.meta_model.ast_node import ASTNode
+from pynestml.meta_model.ast_simple_expression import ASTSimpleExpression
 from pynestml.meta_model.ast_variable import ASTVariable
 
 
@@ -58,7 +59,7 @@ class ASTDeclaration(ASTNode):
         invariant = None
     """
 
-    def __init__(self, is_recordable: bool = False, is_inline_expression: bool = False, _variables: Optional[List[ASTVariable]] = None, data_type: Optional[ASTDataType] = None, size_parameter: Optional[str] = None,
+    def __init__(self, is_recordable: bool = False, is_inline_expression: bool = False, _variables: Optional[List[ASTVariable]] = None, data_type: Optional[ASTDataType] = None, size_parameter: Optional[Union[ASTSimpleExpression, ASTExpression]] = None,
                  expression: Optional[ASTExpression] = None, invariant: Optional[ASTExpression] = None, decorators=None, *args, **kwargs):
         """
         Standard constructor.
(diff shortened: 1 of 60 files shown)


pynestml/meta_model/ast_declaration.py
```diff
@@ -148,37 +149,31 @@ def get_data_type(self):
...
         return self.data_type
 
-    def has_size_parameter(self):
+    def has_size_parameter(self) -> bool:
         """
         Returns whether the declaration has a size parameter or not.
         :return: True if has size parameter, else False.
-        :rtype: bool
         """
-        return self.size_parameter is not None
```
pnbabu, line 157: This condition should be retained for the function to return `bool`.

pynestml/meta_model/ast_expression.py
```diff
@@ -313,59 +316,48 @@ def get_function_calls(self):
...
-            if self.get_expression().get_parent(ast) is not None:
-                return self.get_expression().get_parent(ast)
+            return [self.get_expression()]
+
         if self.is_unary_operator():
-            if self.get_unary_operator() is ast:
-                return self
-            if self.get_unary_operator().get_parent(ast) is not None:
-                return self.get_unary_operator().get_parent(ast)
+            return [self.get_unary_operator()]
```
pnbabu, line 328: In the case of a unary operator, should the expression associated with the operator be included in the children?

pynestml/meta_model/ast_input_port.py
```diff
@@ -118,9 +120,9 @@ def has_size_parameter(self) -> bool:
         Returns whether a size parameter has been defined.
         :return: True if size has been used, otherwise False.
         """
-        return self.size_parameter is not None
```
pnbabu, line 121: Should this condition be removed?

clinssen:
Thank you for the review, good catch! I have addressed all your comments in the latest commit.

pnbabu (approved):
Looks good to me, thanks!

clinssen:
Thanks for the review!

PR #1060, opened 2024-06-03
Allow neuron models without state and parameters block; fix dependent variables search for co-generation transformer

clinssen:
- Allow neuron models without state or parameters block
- Fix dependent variables search for co-generation transformer: if a variable is assigned to inside an if- or elif-block, it depends on the variables in the condition for that block


pynestml/utils/ast_utils.py
```diff
@@ -449,10 +451,31 @@ def create_state_block(cls, model: ASTModel):
         """
         # local import since otherwise circular dependency
         from pynestml.meta_model.ast_node_factory import ASTNodeFactory
-        if not model.get_internals_blocks():
-            state = ASTNodeFactory.create_ast_block_with_variables(True, False, False, list(),
+        if not model.get_state_blocks():
```
pnbabu, line 454: Could you update the docs of this function?

tests/nest_tests/resources/minimal_neuron.nestml
```diff
@@ -0,0 +1,3 @@
+model minimal_neuron:
```
pnbabu, line 1: What does having this kind of a neuron mean?

pnbabu (changes requested):

pynestml/utils/ast_utils.py
clinssen: Done!

tests/nest_tests/resources/minimal_neuron.nestml
clinssen: Added a documentation header to the model.

pnbabu (approved):

tests/nest_tests/test_minimal_stdp_neuron.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class TestMinimalSTDPNeuron:
+
+    @pytest.fixture(autouse=True,
+                    scope="module")
+    def generate_model_code(self):
+        """Generate the model code"""
+
+        jit_codegen_opts = {"neuron_synapse_pairs": [{"neuron": "minimal_neuron",
```
pnbabu, line 53: The `delay` and the `weight` variables should be passed in `codegen_opts` for the tests to pass.

Issue #703, opened 2021-12-02
Fix edge case bug in triplet synapse


clinssen:
See https://github.com/nest/nest-simulator/issues/1990 for a full description. NESTML now supports the ``priority`` parameter for onReceive blocks, that allow to fix this edge case. The integration test for the triplet synapse should be modified accordingly.

clinssen:
Subsidiary: remove the "naive" triplet STDP synapse model file.

PR #1061, opened 2024-06-03
Fix STDP access counter logic

### LINKED ISSUE #1057: Plasticity rule not implemented correctly in certain cases
Vieth and colleagues [1] have defined a plasticity rule that increases the synaptic weight by a fixed amount if the post-synaptic neuron fires within a fixed time interval after the pre-synaptic neuron, see their table S2.

NESTML generates incorrect code for this model with two types of errors:
1. If, following a pre-post spike pair that should trigger a weight increase, the postsynaptic neuron fires at least twice before the next pre-synaptic spike, plasticity will not occur.
2. If there are multiple pre-post spike pairs separated by min_delay intervals, plasticity does not occur.

The first issue seems to be cause by `unsigned int num_transferred_variables = 0;` in the generated `set_spiketime()` method. This variable should be 1.

The cause of the second issue not clear at present; it is of limited practical concern since the actual refractory time in the model by Vieth et al is more than 10 min delays, but it still seems to be a corner case that should be understood.

Attached below is an archive containing 
- NESTML model specifications for a parrot neuron designed to study plasticity and the synapse model by Vieth et al
- a minimal reproducer for each of the two cases, 
- a script to build the necessary module.

[1] Vieth, M., Rahimi, A., Gorgan Mohammadi, A., Triesch, J., & Ganjtabesh, M. (2024). Accelerating spiking neural network simulations with PymoNNto and PymoNNtorch. Frontiers in Neuroinformatics, 18, 1331220. https://doi.org/10.3389/fninf.2024.1331220

clinssen:
The access counter should not have a multiplicative factor ``num_transferred_variables``, because each synapse only calls ``get_history__()`` once on its postsynaptic partner, regardless of how many variables were transferred. The main change of this PR is to remove ``num_transferred_variables``. This requires postsynaptic variables to be read from the postsynaptic spike buffer iterator (``start->...``) instead of from a postsynaptic getter method (``get...(_tr_t)``).

Also fixes an issue with initialisation of synaptic weight. @heplesser: your test case was added as an integration test and now passes, but I have changed the spike times, could you double-check?

Fixes #1057.


heplesser:
This should fix #1057.

models/synapses/stdp_triplet_synapse.nestml
```diff
@@ -57,7 +57,7 @@ model stdp_triplet_synapse:
     onReceive(post_spikes):
         # potentiate synapse
         #w_ nS = Wmax * ( w / Wmax + tr_r1 * ( A2_plus + A3_plus * tr_o2 ) )
-        w_ nS = w + tr_r1 * ( A2_plus + A3_plus * tr_o2 )
+        w_ nS = w + tr_r1 * ( A2_plus + A3_plus * (tr_o2 - 1) )
```
clinssen, line 60: Add a comment to see #703

tests/nest_tests/resources/test_plasticity_dynamics_neuron.nestml
```diff
@@ -0,0 +1,45 @@
...
+
+NEST is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+"""
+model test_plasticity_dynamics_neuron:
```
heplesser, line 25: I'd insert an empty line above and add a comment describing the logic of the neuron: parrot on one port and a second port to support plasticity.

tests/nest_tests/resources/test_plasticity_dynamics_synapse.nestml
```diff
@@ -0,0 +1,47 @@
...
+
+NEST is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+"""
+model test_plasticity_dynamics_synapse:
```
heplesser, line 25: Add an empty line above and a comment explaining the logic. Further, the synapse model is based on `simple_stdp_synapse` in  https://github.com/saeedark/Fast-SNN-PymoNNto-rch/blob/main/Benchmarks/Simple/nest_native_LIF.py, which is licensed under MIT License. Thus we need to cite the source and include the original license in this file (https://github.com/saeedark/Fast-SNN-PymoNNto-rch/blob/main/LICENSE).

tests/nest_tests/test_plasticity_dynamics.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.mark.skipif(NESTTools.detect_nest_version().startswith("v2"),
+                    reason="This test does not support NEST 2")
+def test_plasticity_dynamics():
+    r"""Test that the time interval-based plasticity rule in ``test_plasticity_dynamics_synapse`` is implemented correctly on a simple sequences of spikes"""
+    nest.ResetKernel()
+    nest.resolution = .01
+    nest.Install("nestmlmodule")
+
+    pre_spikes = nest.Create('spike_train_injector', params={'spike_times': [1, 10]})
+    post_driver_spikes = nest.Create('spike_train_injector', params={'spike_times': [.5]})
```
heplesser, line 69: I do not understand the logic behind your change in spike times. If I remember correctly, I carefully engineered the spike times in the two reproducer scripts to trigger the error.

heplesser (changes requested):
@clinssen I am out of my depth when it comes to NESTML internals. I added some comments to the tests.

tests/nest_tests/test_plasticity_dynamics.py
clinssen: It was a more minimal set of spike times that reproduced the issue, but I have swapped it back out for your original times.

tests/nest_tests/resources/test_plasticity_dynamics_neuron.nestml
clinssen: Done!

tests/nest_tests/resources/test_plasticity_dynamics_synapse.nestml
clinssen: Waiting for @terhorstd to confirm that this is OK.

clinssen: Everything checks out on the copyright front; we're good to go!

heplesser (approved):
@clinssen Fine with me now, just needs a conflict resolution.

PR #1071, opened 2024-06-28
Replace outdated mentions of ``deliver_spike()``

pnbabu:


clinssen:
Could you also remove lines 66-67 from ``spinnaker_c_function_call_printer.py``? These aren't necessary as the SpiNNaker target does not yet support synapses.

clinssen (approved):
Great, thank you!

PR #1078, opened 2024-07-04
Fix third-factor plasticity buffering and add third factor plasticity tutorial

clinssen:
Builds upon #1027 by adding a new code generator option ``continuous_state_buffering_method`` to allow postsynaptic 3rd factor buffering using a spike-based buffer rather than a continuous time (simulation resolution) based buffer.


doc/running/running_nest.rst
```diff
@@ -218,16 +218,16 @@ Simulation of volume-transmitted neuromodulation in NEST can be done using "volu
...
+
+As a computationally more efficient alternative, a spike-based buffer can be selected. In this case, the third factor is not stored every timestep, but only upon the occurrence of postsynaptic (somatic) spikes. Because of the existence of a nonzero dendritic delay, the time at which the somatic spike is observed at the soma is delayed, and the time at which the third factor is sampled should match the time of the spike at the synapse, rather than the soma. When the spike-based buffering method is used, the dendritic delay is therefore ignored, because the third factor is sampled instead at the time of the somatic spike.
```
pnbabu, line 228: Should it be `third factor is sampled instead at the time of the spike at the synapse` in the last line?

clinssen: The last line is correct, but the sentence before it should have read "the time at which the somatic spike is observed at the synapse is delayed", rather than "at the soma", which makes no sense. Cheers!

pnbabu: The sentence is yet to be changed in the `running.rst` file.

pnbabu (approved):
Looks good to me, apart from the small comment. I am fine merging this after the fix.

clinssen:
Thanks for pointing that out! This is fixed in the latest commit. Then I am hereby merging!

PR #1081, opened 2024-07-12
Add pretty code rendering feature

clinssen:
Use for marketing/documentation purposes.


tests/nest_tests/print_function_code_generator_test.py
```diff
@@ -136,7 +136,3 @@ def test_print_statment_in_function(self):
         with open(str(os.path.realpath(os.path.join(os.path.dirname(__file__), os.path.join(
                 os.pardir, 'target', 'print_test_function.cpp')))), 'r') as reader:
             self.assertEqual(reader.read().count('std::cout'), 3)
-
```
pnbabu, line 139: Why is this removed here?

tests/nest_tests/print_statement_test.py
```diff
@@ -44,7 +44,3 @@ def test_print_statement(self):
         self.assertEqual(matches[0], "print: This is a simple print statement\n")
         self.assertEqual(matches[1], "print: Membrane voltage: -0.05 V, threshold: -7e-08 MA Ohm, and V_rel: -50 mV\n")
         self.assertEqual(matches[2], "print: Numeric state variable: 0.048731\n")
-
-    def tearDown(self) -> None:
```
pnbabu, line 48: Same as above

pnbabu (changes requested):
Do we need to merge the images as well? We can generate them whenever required by just running the script, right?

tests/nest_tests/print_statement_test.py
clinssen: Oops! Removed.

clinssen:
> Do we need to merge the images as well? We can generate them whenever required by just running the script, right?

That's true, I thought it might help to show what the output would look like. Perhaps we should add this feature as a kind of mock "code generator" in the documentation, what do you think?

clinssen:
Well, hypothetically we could add an extra code generation target platform which would do the pretty printing, akin to the model docs generator (autodoc). I'm not sure if it's the right thing to do or if it's going too far for such a simple feature.

pnbabu:
> Well, hypothetically we could add an extra code generation target platform which would do the pretty printing, akin to the model docs generator (autodoc). I'm not sure if it's the right thing to do or if it's going too far for such a simple feature.

I wouldn't go as far as to have a new code generator for this as this feature would be rarely used. I think just retaining the scripts and generating the images when required would be enough.

clinssen:
I have refactored the pretty rendering feature as a separate code generator.

The rendered version of the docs can be found here: https://nestml--1081.org.readthedocs.build/en/1081/running/index.html

pynestml/codegeneration/code_generator.py
```diff
@@ -42,18 +42,11 @@
...
-    def __init__(self, target, options: Optional[Mapping[str, Any]] = None):
-        from pynestml.frontend.pynestml_frontend import get_known_targets
+    r"""Generate code for a given target platform."""
 
-        if not target.upper() in get_known_targets():
-            code, msg = Messages.get_unknown_target(target)
-            Logger.log_message(message=msg, code=code, log_level=LoggingLevel.ERROR)
-            self._target = ""
-            raise InvalidTargetException()
+    _default_options: Mapping[str, Any] = {}
```
pnbabu, line 48: Why is this removed?

clinssen: ``self._target`` was apparently no longer being used, and specifying an invalid target is already prevented earlier in the frontend.

pynestml/codegeneration/resources_pretty_render/nestml_code_render_neuron.nestml
```diff
@@ -0,0 +1,48 @@
+model iaf_psc_alpha_neuron:
```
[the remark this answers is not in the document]
clinssen: Unfortunately not, as they have been fine-tuned for pretty rendering.

clinssen:
Thank you for the review! The latest changes should address your comments.

pynestml/codegeneration/python_standalone_code_generator.py
```diff
@@ -68,7 +68,7 @@ class PythonStandaloneCodeGenerator(NESTCodeGenerator):
     }
 
     def __init__(self, options: Optional[Mapping[str, Any]] = None):
-        super(NESTCodeGenerator, self).__init__("python_standalone", {**PythonStandaloneCodeGenerator._default_options, **(options if options else {})})
+        super(NESTCodeGenerator, self).__init__({**PythonStandaloneCodeGenerator._default_options, **(options if options else {})})
```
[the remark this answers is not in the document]
pnbabu: Can the `__init__` function be simplified as suggested?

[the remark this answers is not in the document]
clinssen: I checked this, but it actually has different semantics and sadly doesn't work.

pnbabu (approved):

clinssen:
Many thanks for the review!

PR #1083, opened 2024-07-12
Clean up templates and debug print statements

clinssen:


pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
```diff
@@ -1344,12 +1337,13 @@ inline void
...
+  std::cout << "[synapse " << this << "] {{ synapseName }}::trigger_update_weight(): \tupdating from " << t_lastspike_ << " to " << t_trig + dendritic_delay << std::endl;
 #endif
 
- /* update_internal_state_(t_lastspike_,
-                         t_trig - t_lastspike_,
-                         cp);*/
+  // N.B. call to update_internal state is not needed; this is already taken care of by ``process_{{vt_port}}_spikes_()``
+  // update_internal_state_(t_lastspike_,
+  //                        t_trig - t_lastspike_,
+  //                        cp);
```
pnbabu, line 1346: Do you want to remove this then?

tests/nest_tests/print_function_code_generator_test.py
```diff
@@ -69,7 +69,7 @@ def test_simple_print_statment(self):
 
         with open(str(os.path.realpath(os.path.join(os.path.dirname(__file__), os.path.join(
                 os.pardir, 'target', 'simple_print_test.cpp')))), 'r') as reader:
-            self.assertEqual(reader.read().count('std::cout'), 3)
+            self.assertEqual(reader.read().count('std::cout'), 5)
```
pnbabu, line 72: Why are the `std::cout` counts increased here?

pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
clinssen: I don't really mind very much, but would slightly prefer to keep these lines, to make explicit what would otherwise happen; hopefully the comment explains it sufficiently.

tests/nest_tests/print_function_code_generator_test.py
clinssen: To accommodate the extra ``std::cout``s that were added to the template.

clinssen: They get rendered into the ``.cpp`` file, still enclosed by the ``#ifdef DEBUG`` condition. They will only be ignored once they get to the C++ preprocessor.

pnbabu: Yeah right, the generated code will still have it but will be ignored during runtime depending on the `#DEBUG` flag. Sorry for the confusion, my bad!

pnbabu (approved):
Thanks!

PR #1089, opened 2024-07-20
Postsynaptic variable integration fix

### LINKED ISSUE #1084: Convolution Kernel Reset After Spike Generation
Goodmorning, generating a neuron-synapse model, and comparing its dynamics with a NEST-native synapse, NESTML-generated neuron model, I noticed that, when using Mulisynapse models, the Kernel Convolution between a spiking input (e.g. AMPA or NMDA) is reset after the emit_spike() function only if the synaptic model is generated using NESTML.

I am running with NEST 3.6.0 and NESTML 7.0.2.post0.dev0.

Attached are some plots around an emit_spike() function call, as well as the code usable to reproduce the error. 

around_spike_0
around_spike_8
[Convolution_Issue.zip](https://github.com/user-attachments/files/16231760/Convolution_Issue.zip)

clinssen:
This still requires a test to be added.

Fixes #1084.

### DIFF
--- a/pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
+++ b/pynestml/codegeneration/resources_nest/point_neuron/common/NeuronClass.jinja2
@@ -1124,10 +1124,10 @@ void
       assert(history_.back().t_ == last_spike_);
 
 {%-     for var in purely_numeric_state_variables_moved|sort %}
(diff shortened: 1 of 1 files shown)


pnbabu:
Could you add the test or would it be in a separate PR?

clinssen:
As we will eventually move away from having these separate entries for ``numeric_state_variables_moved`` and such, in favour of a "pure" model transformer, would you be OK merging this bugfix as-is?

pnbabu (approved):

PR #1093, opened 2024-08-13
Refactor inline expressions expansion into a transformer

clinssen:
Fixes a bug where parent was not defined after inline expressions were rewritten.

### DIFF
--- a/pynestml/codegeneration/nest_code_generator.py
+++ b/pynestml/codegeneration/nest_code_generator.py
@@ -323,7 +323,6 @@ class NESTCodeGenerator(CodeGenerator):
         equations_block = neuron.get_equations_blocks()[0]
 
         kernel_buffers = ASTUtils.generate_kernel_buffers(neuron, equations_block)
-        InlineExpressionExpansionTransformer().transform(neuron)
         delta_factors = ASTUtils.get_delta_factors_(neuron, equations_block)
         ASTUtils.replace_convolve_calls_with_buffers_(neuron, equations_block)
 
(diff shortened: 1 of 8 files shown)


pynestml/codegeneration/nest_code_generator.py
```diff
@@ -322,8 +323,7 @@ def analyse_neuron(self, neuron: ASTModel) -> Tuple[Dict[str, ASTAssignment], Di
         equations_block = neuron.get_equations_blocks()[0]
 
         kernel_buffers = ASTUtils.generate_kernel_buffers(neuron, equations_block)
-        ASTUtils.make_inline_expressions_self_contained(equations_block.get_inline_expressions())
-        ASTUtils.replace_inline_expressions_through_defining_expressions(equations_block.get_ode_equations(), equations_block.get_inline_expressions())
+        InlineExpressionExpansionTransformer().transform(neuron)
```
pnbabu, line 326: Could this be moved to `pynestml_frontend.py` where all the other transformers are called?

tests/nest_tests/resources/beta_function_with_inline_expression_neuron.nestml
```diff
@@ -0,0 +1,30 @@
...
+    weighted_input_spikes  <- spike
+
+  output:
+    spike
+
+  update:
+    integrate_odes()
+
+  onReceive(weighted_input_spikes):
+    x_ += alpha * ( 1./tau2 - 1./tau1) * pA
```
pnbabu, line 30: Can you add `weighted_input_spikes` to this expression as Tom suggested?

tests/nest_tests/resources/beta_function_with_inline_expression_neuron.nestml
```diff
@@ -0,0 +1,30 @@
+model beta_function_with_inline_expression_neuron:
```
pnbabu, line 1: Add copyright

pnbabu (changes requested):

pynestml/codegeneration/nest_code_generator.py
clinssen: I agree that this would be nice! Unfortunately, I tested it, and it does not work due to the order of transformations---the inline expansion currently takes place at a specific stage in the code generator. I think we might be able to do as you suggested after completing https://github.com/nest/nestml/pull/1050, but that will probably still be some time into the future.

pnbabu (approved):

PR #1101, opened 2024-09-23
Allow vector index in onReceive blocks

clinssen:
Allow a formulation such as ``onReceive(spikes[42])``.

@pnbabu: do you think it's a good idea to store the vector index inside ASTOnReceiveBlock? We could also change the current property ``port_name`` (a string) into an ASTSimpleExpression (so it can store the complete expression "spikes[42]").


clinssen:
Subsumed by #1137.

PR #1105, opened 2024-09-29
Run context condition checks only once, after model parsing

clinssen:
This is a spin-off from #1050.

Refactor cocos such that they run only once, after model parsing.

Fixes some other bugs that emerged as part of this refactor. For instance, two cocos turned out to actually make needed changes to the model; the modification part has now been spun off to the new AssignImplicitConversionFactorsTransformer. (I can also make the transformer into a pure visitor; please let me know if you prefer this approach.)

Changes order of operations in compartmental code generator: to prevent symbol not found error, add propagator variable first, then add the update expressions that reference this variable.


pynestml/codegeneration/nest_code_generator.py
```diff
@@ -908,8 +910,8 @@ def update_symbol_table(self, neuron) -> None:
         """
         SymbolTable.delete_model_scope(neuron.get_name())
         symbol_table_visitor = ASTSymbolTableVisitor()
-        symbol_table_visitor.after_ast_rewrite_ = True
         neuron.accept(symbol_table_visitor)
+        CoCosManager.check_cocos(neuron, after_ast_rewrite=True)
```
pnbabu, line 914: Should the cococs be checked inside each code generator? Shouldn't it rather happen between transformation and code generation stages?

pynestml/transformers/assign_implicit_conversion_factors_transformer.py
```diff
@@ -0,0 +1,335 @@
...
+from pynestml.utils.type_caster import TypeCaster
+from pynestml.visitors.ast_visitor import ASTVisitor
+
+
+class AssignImplicitConversionFactorsTransformer(Transformer):
+    r"""
+    Assign implicit conversion factors in expressions.
+    """
+
+    def transform(self, models: Union[ASTNode, Sequence[ASTNode]]) -> Union[ASTNode, Sequence[ASTNode]]:
```
pnbabu, line 50: Which are the two cocos that are replaced by this transformer? Do they also have to be removed from the cocos manager class?

pnbabu (changes requested):

pynestml/codegeneration/nest_code_generator.py
clinssen: Indeed, the primary coco checking call (to ``CoCosManager.check_cococs()``) is in the frontend. This is an additional call (with ``after_ast_rewrite=True``) after model transformations, as an extra sanity check.

pynestml/transformers/assign_implicit_conversion_factors_transformer.py
clinssen: Thank you for remarking upon this. The calls  have been effectively removed by setting ``set_implicit_conversion_factor_on_lhs`` to False on https://github.com/nest/nestml/pull/1105/files#diff-e13518a54108e0f87db406aee6c43eadbebc0bb6908943a1f045136bc1c14024R43. This affects ``co_co_illegal_expression``, ``co_co_user_defined_function_correctly_defined`` and ``co_co_function_calls_consistent``.

pynestml/codegeneration/spinnaker_code_generator.py
```diff
@@ -216,6 +215,7 @@ def generate_code(self, models: Sequence[ASTModel]) -> None:
         for model in models:
```
pnbabu, line 215: Can these two for loops be merged into one?

pynestml/transformers/assign_implicit_conversion_factors_transformer.py
pnbabu: Can these cocos be removed then?

pynestml/codegeneration/spinnaker_code_generator.py
clinssen: Done!

pynestml/utils/model_parser.py
```diff
@@ -142,10 +144,14 @@ def parse_file(cls, file_path=None):
         for model in ast.get_model_list():
             model.accept(ASTSymbolTableVisitor())
             SymbolTable.add_model_scope(model.get_name(), model.get_scope())
+            Logger.set_current_node(model)
+            AssignImplicitConversionFactorsTransformer().transform(model)
```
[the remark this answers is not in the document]
clinssen: Done! Changed the transformer into a visitor.

pnbabu (approved):
Looks good to me, thank you!

clinssen:
Thank you for the review! Let's merge as soon as the tests pass.

Issue #946, opened 2023-08-23
Refactoring events


clinssen:
Thanks to #879, we can now unify neuron and synapse models: instead of treating these as fundamentally different things in NESTML (indicated by the "neuron" and "synapse" top-level keywords), we can have a generic top-level "model" type, which can then turn out to be either a neuron or a synapse, depending on how the code generator is invoked. This makes sense as both are fundamentally hybrid dynamical systems (handling continuous-time dynamics in addition to events).

One issue is coming up as a part of this unification. Currently, neurons can emit spikes by calling the NESTML predefined function ``emit_spike()`` (without any arguments). Synapses can emit spikes by calling ``deliver_spike(w, d)``, passing the weight and delay parameters.

It would be great if we could unify these functions. Could we rewrite this in a way where there is a generic "event", to which we can attach attributes (like weight and delay, or multiplicity, or any named parameter at all)? Then we could do something like:

```
emit_spike()    # neuron emits unweighted spikes, no parameters needed
emit_spike(weight=w_, delay=1 ms)   # synapse emits a spike with an attribute "weight" which gets its value from the local variable w_, and an attribute "delay" which is assigned a constant value here
```

These attributes could subsequently be read out in an event handler as follows:

```
onReceive(exc_spikes):
    I_syn_exc += exc_spikes.weight
    foo += exc_spikes.multiplicity
```

as opposed to the currently envisaged:

```
onReceive(exc_spikes):
    I_syn_exc += exc_spikes
```

The latter is in line with our current way of thinking about incoming spikes as a sum of delta pulses (with physical units 1/s). In the former, the event attribute "weight" would simply have the units of whatever we put in on the sending side, but we would still need a way to specify the expected unit on the receiving side, which might be tricky syntactically. Perhaps we could define something like

```
input:
    exc_spikes <- spike(weight: real, delay: ms)
```

-----

Alternatively, we could have a generic ``emit()`` function:

```
emit(spike(weight=w_, delay=1 ms))
emit(rate(rate=99/s))
```

```
state:
  x 1/s = 99 / s

output:
   continuous <- x
```

pnbabu:
As discussed in the NESTML meeting, the syntax for the `onReceive` block with vector input ports will be as follows:

```
input:
  spikes[3] <- spike(weight:real, multiplicity: real)

onReceive(spikes[0]):
  I_syn += spikes[0]

onReceive(spikes[1]):
  # using weight and multiplicity of the spike event
[...]

PR #1124, opened 2024-10-17
Add explicit output parameters to spiking output ports

clinssen:
Part of #946.

Add explicit output parameters to spiking output ports: if ``emit_spike(foo, bar)`` is called, then the spiking output port should be parameterised by two parameters with types equal to (or castable to) those of foo and bar.

Todo:
- [x] update documentation
- [x]  add coco to check that continuous-type output ports do not support attributes
- [ ]  <S>(?) disallow spike attribute names to alias variable names in the model scope (such as in the state or parameters block), e.g. "spike_port.foo" is not allowed when a state variable or parameter named "foo" is defined</s>


pynestml/cocos/co_co_no_attributes_on_continuous_port.py
```diff
@@ -0,0 +1,56 @@
...
+from pynestml.symbols.predefined_functions import PredefinedFunctions
+from pynestml.utils.ast_utils import ASTUtils
+from pynestml.utils.logger import Logger, LoggingLevel
+from pynestml.utils.messages import Messages
+from pynestml.visitors.ast_visitor import ASTVisitor
+
+
+class CoCoNoAttributesOnContinuousPort(CoCo):
+    """
+    This context condition checker ensures that no attributes are defined on continuous time output ports.
```
pnbabu, line 36: If true for all code-generation targets, could we remove the attributes from the grammar for continuous ports?

tests/test_cocos.py
```diff
@@ -324,6 +324,29 @@ def test_valid_output_port_defined_if_emit_call(self):
...
+        assert len(Logger.get_all_messages_of_level_and_or_node(model, LoggingLevel.ERROR)) > 0
+
+    def test_valid_output_port_type_if_emit_call(self):
+        """test that an error is raised when the emit_spike() function is called with different parameter types than are defined in the spiking output port"""
+        model = self._parse_and_validate_model(os.path.join(os.path.realpath(os.path.join(os.path.dirname(__file__), 'invalid')), 'CoCoOutputPortTypeIfEmitCall-3.nestml'))
+        assert model is not None
+        assert len(Logger.get_all_messages_of_level_and_or_node(model, LoggingLevel.ERROR)) == 0
+        assert len(Logger.get_all_messages_of_level_and_or_node(model, LoggingLevel.WARNING)) > 0
+
+    def test_valid_output_port_type_continuous(self):
```
pnbabu, line 344:
```suggestion
    def test_invalid_output_port_type_continuous(self):
```

pnbabu (changes requested):

pynestml/cocos/co_co_no_attributes_on_continuous_port.py
clinssen: Done!

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -907,6 +907,28 @@ Each model can only send a single type of event. The type of the event has to be
...
+
+.. code-block:: nestml
+
+   parameters:
+       weight real = 10.
+
+   update:
+       emit_spike(w, 1 ms)
+
+If spike event attributes are used, their names and types must be given as part the output port specification. The names are only used externally, for other models making reference to this one (it is allowed to have a state variable called ``weight`` and an output port attribute by the same name).
```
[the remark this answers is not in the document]
clinssen: Clarified, please check!

PR #1138, opened 2024-11-06
Bump minimum Python version requirement to 3.9 and update CI

clinssen:
Needed for #1026 due to dependency on a newer matplotlib.

Python 3.9 is the default available on the JSC login node as well as JUSUF.


pnbabu (commented):
Could you also change the `setup.py` file?

clinssen:
Oops, thanks, done!

pnbabu (approved):
Thanks!

PR #1139, opened 2024-11-06
Integrate the entire dynamics for higher-order ODEs

pnbabu:


tests/nest_tests/test_integrate_odes.py
```diff
@@ -232,3 +230,50 @@ def test_integrate_odes_params2(self):
...
+            for _ax in ax:
+                _ax.grid(which="major", axis="both")
+                _ax.grid(which="minor", axis="x", linestyle=":", alpha=.4)
+                _ax.set_xlim(0., simtime)
+                _ax.legend()
+
+            fig.savefig("/tmp/test_integrate_odes_higher_order.png", dpi=300)
+
+        # verify
+        np.testing.assert_allclose(x[-1], 0.10737970490959549)
```
clinssen, line 278: How do we know these values are correct (are they from just running the NESTML model once, or from an independent implementation of the ODEs)?

clinssen (changes requested):

clinssen:
Thanks for the PR! Could you add/update the documentation for the ``integrate_odes()`` function as well?

tests/nest_tests/test_integrate_odes.py
pnbabu: These values are from repeated simulations of the NESTML-generated code, equivalent to providing `integrate_odes(x',x)` in the NESTML model. I have changed the test now to include calls to both `integrate_odes` and compare their output.

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1121,6 +1121,23 @@ Integrating the ODEs needs to be triggered explicitly inside the ``update`` bloc
...
 
+In case of higher-order ODEs of the form ``F(x'', x', x) = 0``, the solution to ``x(t)`` can be integrated by just providing the variable ``x`` to the ``integrate_odes`` function. For example,
```
clinssen, line 1124:
```suggestion
In case of higher-order ODEs of the form ``F(x'', x', x) = 0``, the solution ``x(t)`` can be integrated by just providing the variable ``x`` to the ``integrate_odes`` function. For example,
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1121,6 +1121,23 @@ Integrating the ODEs needs to be triggered explicitly inside the ``update`` bloc
...
+   state:
+     x  real    = 0
+     x' ms**-1  = 0 * ms**-1
+
+   equations:
+     x'' = - 2 * x' / ms - x / ms**2
+
+   update:
+     integrate_odes(x)
+     # integrate_odes(x', x)  # similar to integrate_odes(x)
```
clinssen, line 1137:
```suggestion
     # integrate_odes(x', x)  # equivalent to integrate_odes(x)
```

clinssen (changes requested):

doc/nestml_language/nestml_language_concepts.rst
tomtetzlaff: "...the solution ``x(t)`` is obtained by..."

It's the ODE that is "integrated", not the solution.

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1121,6 +1121,23 @@ Integrating the ODEs needs to be triggered explicitly inside the ``update`` bloc
...
+     x' ms**-1  = 0 * ms**-1
+
+   equations:
+     x'' = - 2 * x' / ms - x / ms**2
+
+   update:
+     integrate_odes(x)
+     # integrate_odes(x', x)  # equivalent to integrate_odes(x)
+
+Here, ``integrate_odes(x)`` integrates the entire dynamics of ``x(t)`` and is equivalent to ``integrate_odes(x', x)``.
```
clinssen, line 1139: Should it be allowed to write variables with derivatives inside ``integrate_odes()``? As in, ``x`` is allowed, but ``x'`` is not. Otherwise, what is the semantics of saving ``integrate_odes(x')``?

pnbabu: I have now updated one of the existing cocos to check this condition.

clinssen (approved):
Great, thank you!

PR #1143, opened 2024-11-18
Update conda installation docs

pnbabu:
Rendered docs here: https://nestml-doc.readthedocs.io


steffengraber (approved):
:+1: 
We will recommend `mamba` as tool, as this also has some advantages when installing larger environments. However, the description is perfectly fine.

PR #1144, opened 2024-11-28
Clean up NESTML grammar definition

clinssen:
One of the unit tests (``nestml_printer_test.py``) was silently failing on master, giving parse errors which were not resulting in exceptions/test failures.

This PR fixes the issue and cleans up (internally refactors) the NESTML grammar definition.

The term "block" only refers to top-level elements in a model (update, state, parameters, onReceive, etc.) The inner "ASTBlock" was renamed to ASTStmtsBody as it includes an INDENT in the grammar rule.


pynestml/grammars/PyNestMLParser.g4
```diff
@@ -54,303 +52,310 @@ parser grammar PyNestMLParser;
...
-  * Procedural-Language
-  *********************************************************************************************************************/
+  // -------------------------------------------------------------------------
+  //  Procedural
+  // -------------------------------------------------------------------------
 
-  block : NEWLINE INDENT stmt+ DEDENT;
-
-  stmt : smallStmt | compoundStmt;
+  stmt : NEWLINE? (smallStmt | compoundStmt);
```
pnbabu, line 154: Would this work if there were multiple newlines at the beginning of the `stmt`?

tests/test_nestml_printer.py
```diff
@@ -32,13 +33,13 @@
 from pynestml.utils.model_parser import ModelParser
 
 
-class NestMLPrinterTest(unittest.TestCase):
+class TestNESTMLMLPrinter:
```
pnbabu, line 36:
```suggestion
class TestNESTMLPrinter:
```

tests/test_nestml_printer.py
```diff
@@ -47,73 +48,73 @@ def setUp(self):
...
-                "    # real pre\n" \
-                "    # real pre2\n" \
-                "    r real = 0\n"
-        model = ModelParser.parse_block_with_variables(block)
+        block = "    # pre1\n" \
+                "    state: # in\n" \
+                "        # real pre\n" \
+                "        # real pre2\n" \
+                "        r real = 0\n"
+        model = ModelParser.parse_model_body(block)
```
pnbabu, line 56: Would this be any different if parsed as `block_with_variables`?

pnbabu (commented):
Thank you for fixing this! I have two minor comments.

tests/test_nestml_printer.py
clinssen: That wouldn't work, because it cannot account for the pre comment (which leaves behind stray "\n" characters as the comments go into the COMMENTS channel during parsing.

pynestml/grammars/PyNestMLParser.g4
clinssen: I changed this so the newline is now in the ``stmtsBody`` rule. ``NEWLINE`` captures multiple newlines (this is already covered by the NESTML printer test).

pnbabu (approved):
Looks good to me, thanks!

clinssen:
Great, thank you!

PR #1149, opened 2024-12-10
Modify the code generator for the NEST Desktop target

pnbabu:
Modify the code generator from NEST to Python Standalone for the NEST Desktop target. Also, add the state variables into the output json file.


clinssen (approved):

clinssen:
Thanks for the updates! I am fine with this if @babsey also gives it the thumbs-up.

PR #1166, opened 2025-02-07
Add codemeta file

clinssen:
This is now required for the EBRAINS Knowledge Graph.


codemeta.json
```diff
@@ -0,0 +1,27 @@
+{
+    "@context": "https://w3id.org/codemeta/3.0",
```
pnbabu, line 2: I think we should also add the version of NESTML here.

pnbabu (changes requested):

codemeta.json
clinssen: Cheers, done!

pnbabu (approved):

PR #1169, opened 2025-02-13
Use single-line comment characters for docstrings

clinssen:
There is no need to use a special ``"""`` token for model docstrings; NESTML already processes a series of single-line comments preceding a model definition as a "pre comment" and belonging to the model AST node.

This change makes docstring comments more consistent particularly when using the ``include`` statement (#1121).


pnbabu (approved):
Looks good to me! I am fine merging this after the CI turns green.

PR #1171, opened 2025-03-05
Make distinction between neuron and synapse model interpretation more clear

clinssen:


doc/running/running_nest.rst
```diff
@@ -66,6 +66,12 @@ Sometimes it can be convenient to directly edit the generated code. To manually
 where ``<nest_install_dir>`` is the installation directory of NEST (e.g. ``/home/nest/work/nest-install``).
 
 
+Generating code for synapses
```
pnbabu, line 69: Could you tell me the reason for this? Isn't it better to be explicit and provide the names of the neurons and synapses separately to avoid any typos, if at all?

clinssen: It is certainly possible to remove the heuristic that looks at the model name suffix. Would you be in favour of doing that? It means adding another parameter/SLOC to each ``generate_nest_target()`` call.

pnbabu (approved):
Thanks for the changes!

PR #1172, opened 2025-03-06
Fix and add test for saturating-type equations in synapse models

clinssen:
Adds an expression ``post_trace += foobar - .1 * post_trace`` where the right-hand side depends on parameters and variables that are moved from synapse to neuron by the co-generation transformer.


pnbabu (approved):
Thanks for the fix!

PR #1174, opened 2025-03-07
Add test for parrot neuron

clinssen:
Test a parrot neuron that repeats all spikes received on its first input port.

This tests that spikes can be emitted from inside onReceive blocks.


pnbabu (approved):
Looks good to me!

PR #1177, opened 2025-03-11
Limit number of compiler processes

clinssen:
This can prevent the compiler process from getting killed due to out-of-memory on systems with many cores but not so much memory available (like EBRAINS-lab).


pnbabu (approved):

clinssen:
Thanks for the review!

PR #1181, opened 2025-03-13
Refactor NESTML printer

clinssen:
Refactor the NESTML printer to be modular, just like the other printers.

This is needed for #1050; spun off from there.

- [x] Should be merged after #1182.


pnbabu (approved):
Thanks for the changes!

clinssen:
Thanks for the review!

Issue #1163, opened 2025-01-23
Issue with triplet STDP triplet synapse model


clinssen:
There is a regression in the STDP triplet synapse model, as is visible in the tutorial notebook on the current master branch.

The regression bug should be fixed, and assertions have to be added to the notebook to prevent this issue from happening again.

PR #1183, opened 2025-03-16
Temporary fix for STDP trace

clinssen:
Temporary fix (but does not fully resolve) #1163.


pnbabu (commented):
I have just one comment: is there a reason to remove `mu_plus` and `mu_minus` from the `stdp` model in the notebook?

clinssen:
Thanks for the review! These were not relevant to the tutorial and always fixed in value, and while I was trying to find the source of the issue, I removed some things to be more minimal/down to the core point. However, I can reinstate them if you prefer.

pnbabu:
I am fine with this minimal version as well, as we anyway have the detailed model in the `models` directory.

pnbabu (approved):
I am fine merging this after resolving the merge conflicts. Thanks!

PR #1202, opened 2025-04-23
Fix printing of numeric literals with units

### LINKED ISSUE #1128: Operator precedence and units
The NESTML string "0 ms**-1" can be read as either "0 (ms**-1)", a rate of change of 0 per millisecond, or as "(0 ms)**-1", a division by zero error.

(To obtain the former, one can write "0 * ms**-1".)

Possibly, operator precedence rules should be changed such that the former interpretation is the default.

pnbabu:
Fixes #1128 

If a variable has a numeric literal assigned to a unit, for example, ``x pA**-1 = 0 pA**-1``, the units are parsed correctly, where both the LHS and the RHS of the expression have a unit of ``1/pA``. The bug was how these expressions were printed before sending them to the ode-toolbox. Before, the above expression was printed as ``x = 0**-1`` where the exponent in the unit was incorrectly considered as the expression itself. This PR fixes this with more precedence to the power operator and printing the correct expression, ``x = 0``.


clinssen:
Thank you very much for tracking this down and for the fix!

Do you think we should add an xfail test for the expression (0 pA)**-1? Or perhaps this is obvious already to do the right thing?

PR #1206, opened 2025-05-06
Remove unused ``module_name`` parameter from SpiNNaker code generator API

mottenmann99:
Added module_name to generate_target() parameters within generate_spinnaker_target(), due following error when generating spinnaker-target-code:
[5,GLOBAL, INFO]: No module name specified; the generated module will be named "nestmlmodule"


clinssen:
@mottenmann99: before merging this, we should test that the pipeline actually works with a different module name.

clinssen (approved):

PR #1207, opened 2025-05-07
Add support for GeNN code generation

clinssen:


tests/genn_tests/test_iaf_psc_exp.py
```diff
@@ -0,0 +1,182 @@
...
+        model.build()
+        model.load(num_recording_timesteps=SIM_TIMESTEPS)
+
+        V_m_log = []
+        while model.timestep < SIM_TIMESTEPS:
+            model.step_time()
+            neuron_pop.vars["V_m"].pull_from_device()
+            V_m_log.append(neuron_pop.vars["V_m"].values)
+
+        model.pull_recording_buffers_from_device()
```
pnbabu, line 168: Since GeNN generates code for NVIDIA GPUs, where does the simulation run in this case?

pnbabu (changes requested):

doc/running/running_genn.rst
```diff
@@ -0,0 +1,37 @@
...
+
+NESTML code generation support for GeNN currently covers neuron models with linear dynamics that can be solved with propagators, as well as neurons that require a numeric solver, which for GeNN is implemented (in the neuron code templates) as a forward Euler solver.
+
+Please see the unit tests in https://github.com/nest/nestml/tree/master/tests/genn_tests for usage examples.
+
+
+Generating code
+~~~~~~~~~~~~~~~
+
+1. Install GeNN
```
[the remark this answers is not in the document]
clinssen: I have added a citation. I could only find a versioned link and I'm afraid it will need to be updated frequently.

tests/genn_tests/test_iaf_psc_exp.py
clinssen: It runs on GeNN's CPU backend instead.

pnbabu (approved):
Thank you for adding this target!

PR #1213, opened 2025-05-10
Fix typesetting in documentation

clinssen:


doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -817,7 +817,7 @@ Block types
 -  ``state`` - This block is composed of a list of variable declarations that describe parts of the model which may change over time. All the variables declared in this block must be initialized with a value.
 -  ``equations`` - This block contains kernel definitions and differential equations. It will be explained in further detail `later on in the manual <#equations>`__.
 -  ``input`` - This block is composed of one or more input ports. It will be explained in further detail `later on in the manual <#input>`__.
--  ``output`` *``<event_type>``* - Defines which type of event the model can send. Currently, only ``spike`` is supported.
+-  ``output`` - Defines which type of event the model can send.
```
pnbabu, line 820: Why is `Currently, only ``spike`` is supported.` removed?

clinssen: It is mentioned in the "Output" section of the same document, and I removed it here to eliminate redundancy. But I don't have a strong opinion about it and would be happy to reinstate it here too.

pnbabu (approved):

PR #1217, opened 2025-05-21
Add model transformer for non-dimensionalisation

### LINKED ISSUE #984: Add model transformer for non-dimensionalisation
Currently, physical units and their factors (nano-, micro-, kilo-, etc.) are stored in the model AST and converted by the printers during code generation. 

I would suggest to replace this architecture with a transformer, that converts a model with quantities in terms of physical units, to a model containing only real-typed quantities (and int, bool, string, etc). Then, the printers and code generator do not have to deal with conversion factors.

See also #608.

clinssen:
Fixes #984.


clinssen:
This PR is replaced by #1234.

Issue #561, opened 2020-08-26
Add MessageCode to Logger to simplify unit testing


clinssen:
Right now, messages such as warnings and errors are recorded in the Logger. So we can do e.g.:

```py
(Pdb) Logger.get_all_messages_of_level_and_or_neuron(model.get_neuron_list()[0], LoggingLevel.WARNING)[0]
(<pynestml.meta_model.ast_neuron.ASTNeuron object at 0x7f4232793898>, <LoggingLevel.WARNING: 1>, 'Input block not defined!')
(Pdb) Logger.get_all_messages_of_level_and_or_neuron(model.get_neuron_list()[0], LoggingLevel.WARNING)[1]
(<pynestml.meta_model.ast_neuron.ASTNeuron object at 0x7f4232793898>, <LoggingLevel.WARNING: 1>, 'Output block not defined!')
(Pdb) Logger.get_all_messages_of_level_and_or_neuron(model.get_neuron_list()[0], LoggingLevel.WARNING)[2]
(<pynestml.meta_model.ast_neuron.ASTNeuron object at 0x7f4232793898>, <LoggingLevel.WARNING: 1>, "Variable 'eV' has the same name as a physical unit!")
```

Unit tests are often checking for the number of warnings (i.e. the length of this list), not what the warnings are saying. See https://github.com/nest/nestml/blob/master/tests/cocos_test.py for an example. This means that when we add a warning for something, we have to touch a lot of unrelated unit tests.

We should additionally record the MessageCode (from https://github.com/nest/nestml/blob/master/pynestml/utils/messages.py#L997) in the Logger tuple, to simplify unit testing. Each test could then check for the presence or absence of a particular MessageCode rather than the length of the warnings (or errors) list.

PR #1220, opened 2025-06-10
Fix fractional exponents in unit types

### LINKED ISSUE #1219: Issue when declaring parameters with non-integer measurement units
When declaring variables in the parameters block that include non-integer exponents an error occurs. This seems to affect expressions where the exponent is a fractional power, e.g., `sigmal_pl nC*s**0.5 = 0.29 nC*s**0.5`. I've tried with parentheses and using different  but none of them worked.

This is the error: `[12,GLOBAL, ERROR, [15:24;15:24]]: Error occurred during parsing: mismatched input '0.5' expecting {'+', '-', UNSIGNED_INTEGER}`.

I solved by defining auxiliary variables to compute the correct measurement units when using the variable in the equations block: `sigmal_pl * C_aux * T_aux**0.5`, with `C_aux = 1 nC` and `T_aux = 1 s`.

clinssen:
Fixes #1219.

### DIFF
--- a/tests/test_cocos.py
+++ b/tests/test_cocos.py
@@ -34,6 +34,7 @@ from pynestml.symbols.predefined_units import PredefinedUnits
 from pynestml.symbols.predefined_variables import PredefinedVariables
 from pynestml.utils.ast_source_location import ASTSourceLocation
 from pynestml.utils.logger import LoggingLevel, Logger
+from pynestml.utils.messages import MessageCode
 from pynestml.utils.model_parser import ModelParser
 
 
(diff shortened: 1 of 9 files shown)


pynestml/grammars/PyNestMLParser.g4
```diff
@@ -43,15 +43,15 @@ parser grammar PyNestMLParser;
   /**
    * ASTUnitType. Represents an unit data type. It can be a plain data type as 'mV' or a complex data type as 'mV/s'
   **/
+  unitTypeExponent : leftParentheses=LEFT_PAREN num=UNSIGNED_INTEGER divOp=FORWARD_SLASH den=UNSIGNED_INTEGER rightParentheses=RIGHT_PAREN
```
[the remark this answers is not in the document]
clinssen: Thank you for pointing out this incompleteness! I have fixed this and added several tests.

tests/test_cocos.py
```diff
@@ -153,6 +153,38 @@ def test_valid_numerator_of_unit_one(self):
         model = self._parse_and_validate_model(os.path.join(os.path.realpath(os.path.join(os.path.dirname(__file__), 'valid')), 'CoCoUnitNumeratorNotOne.nestml'))
         assert len(Logger.get_all_messages_of_level_and_or_node(model, LoggingLevel.ERROR)) == 0
 
+    def test_invalid_fractional_unit_exponent(self):
+        model = self._parse_and_validate_model(os.path.join(os.path.realpath(os.path.join(os.path.dirname(__file__), 'invalid')), 'fractional_unit_exponent.nestml'))
```
pnbabu, line 157: Would it be better to provide the exact number of errors here, similar to other invalid tests?

pnbabu (commented):
Could you merge master into this branch?

tests/test_cocos.py
clinssen: I have instead partially implemented https://github.com/nest/nestml/issues/561 so that we can check even more accurately whether specific message codes appear in the log.

pnbabu (approved):

PR #1228, opened 2025-06-25
Fix vector initialization in synapse models

pnbabu:


tests/nest_tests/resources/vectors_test_synapse.nestml
```diff
@@ -0,0 +1,24 @@
+# vectors_test_synapse
+# ####################
+#
```
clinssen, line 3: Could you add the full copyright header? (We should add a check for this in the CI...)

clinssen (approved):

PR #1234, opened 2025-07-10
Add a Transformer to support multiple backends with different preferred prefixes a.k.a Non-dimensionalisation-transformer

mwhy:
Relates to [Issue 984](https://github.com/nest/nestml/issues/984)


clinssen:
We should consider changing the default units of time in NESTML, e.g. in the return value of the ``resolution()`` function.

.gitignore
```diff
@@ -7,14 +7,20 @@ target/
 build/
 buildNest/
 venv/
-spinnaker-install/
+tests/spinnaker_tests/bak/spinnaker-install/
```
clinssen, line 10: Could you remove this file from the PR and also all of the SpiNNaker-2 related files from this PR? Cheers!

PR #1238, opened 2025-08-08
Add HPC running instructions to the documentation

clinssen:
With thanks to Renan for the suggestion!


doc/running/index.rst
```diff
@@ -245,17 +245,30 @@ This will generate, compile, build, and install the code for a set of specified
...
 
 .. code-block:: python
 
-   from pynestml.frontend.pynestml_frontend import generate_target
+   from pynestml.frontend.pynestml_frontend import generate_nest_target
 
-   generate_target(input_path="/home/nest/work/pynestml/models/neurons/iaf_psc_exp.nestml",
-                   target_platform="NEST_DESKTOP",
-                   target_path="/tmp/nestml_target")
+   nestml_file_path = "~/nestml-neurons"    # this should be a path inside your home directory
```
pnbabu, line 257: Should this comment be on the line where the `nestml_install_path` is declared?

doc/running/index.rst
```diff
@@ -245,17 +245,30 @@ This will generate, compile, build, and install the code for a set of specified
...
+
+   print(f"Generating NEST target code from {nestml_file_path} to {nestml_target_path}")
+
+   generate_nest_target(
+        input_path=str(nestml_file_path),
+        target_path=str(nestml_target_path),
+        install_path=str(nestml_install_path)
+   )
+
+Then, when running the Python simulation script which contains the ``nest.Install(<module_name>)``, make sure to first set the ``LD_LIBRARY_PATH``, such as:
```
pnbabu, line 270: Doesn't teh toolchain do this automatically when the `install_path` is specified?

doc/running/index.rst
clinssen: But the target path should also be writeable, right? Of course it could also be ``/tmp``.

doc/running/index.rst
clinssen: If the simulation script is run in a separate process, after the NESTML toolchain runs, this directory would not be seen unless given in ``LD_LIBRARY_PATH``, right?

PR #1241, opened 2025-08-22
Adding back externally defined constant dendritic delays and other small features and fixes for the compartmental feature. [WIP]

LeanderEwert:
In the normal nestml generation pipeline, the connection between the STDP synapse and the post-synaptic neuron has a minimum delay determined by the simulation resolution. Due to the STDP dynamics being directly integrated into the receptors in the compartmental case, we do not implicitly have such a delay and have initially not regarded it as necessary, as a delay should emerge from the presence of a compartmental dendritic structure. In a simple one-compartment model, this creates a small mismatch to the behaviour in the point neuron case due to the one-timestep delay missing.


clinssen:
Closing this PR as it has been superseded by #1388.

PR #1250, opened 2025-09-24
Fix printing log function for numeric solver expressions

pnbabu:
The predefined functions `ln()` and `log()`, when used in the `equations` block in the ODEs, result in a compilation error when a numeric solver is selected. 

Currently, the `ln()` function is passed as `log()` to the ode-toolbox, which results in the error. According to the [sympy docs](https://docs.sympy.org/latest/tutorials/intro-tutorial/simplification.html#exponentials-and-logarithms), `log` has an alias `ln()`. Thus, replacing `log` with `ln` in the ode-toolbox printer fixes the error.

### DIFF
--- a/tests/nest_tests/test_log_functions_in_odes.py
+++ b/tests/nest_tests/test_log_functions_in_odes.py
@@ -64,7 +64,7 @@ class TestLogFunctionsInODEs:
         sim_time = 100.0
 
         n = nest.Create("equations_log_neuron")
-        params_dict = n.get()
+        params_dict = nest.GetStatus(n)[0]
         initial_state = [params_dict["foo"], params_dict["x"], params_dict["z"]]
         params = [params_dict["tau_d"], params_dict["tau_r"], params_dict["alpha_x"],
                   params_dict["alpha_z"], params_dict["beta_x"], params_dict["beta_z"],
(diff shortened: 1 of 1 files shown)


pnbabu:
> Either `log10` or `ln` should be used to avoid ambiguities about the base (see https://nestml.readthedocs.io/en/latest/nestml_language/nestml_language_concepts.html#predefined-functions). Does using `log` in a NESTML model (without a custom NESTML model function `log` being defined) not result in an error?

The model has an `ln()` function, which gets translated to `log` while sending to ode-toolbox. The ode-toolbox returns the function as `log` as well. I checked that we do not have any models till now that use an `ln()` or `log10()` function inside the `equations` block. It is either used in the `internals` or the `update` block, or inside a user-defined function.

When the solver returned by the ode-toolbox is `analytic`, we do not see this error because the printers print the function as `log`, which is compiled by the C++ compiler. Whereas in the case of a numeric solver, the `log` is appended to the `node.` string, which results in a compilation error. Basically, the `PredefinedFunctions` check doesn't hold good here, as the function `log` doesn't match with `ln` or `log10`.

pynestml/codegeneration/printers/ode_toolbox_function_call_printer.py
```diff
@@ -89,7 +89,7 @@ def _print_function_call_format_string(self, function_call: ASTFunctionCall) ->
             return 'exp({!s})'
 
         if function_name == PredefinedFunctions.LN:
-            return 'log({!s})'
+            return 'ln({!s})'
 
         if function_name == PredefinedFunctions.LOG10:
             return 'log10({!s})'
```
clinssen:
```suggestion
            return 'log({!s, 10})'
```
sympy.log10() does not exist. Do you know why this was working before?

PR #1270, opened 2025-11-18
Clean up model docstring syntax and add stdsp_synapse to models library

### LINKED ISSUE #1269: Enquiring about the absence of stdsp_synapse model in the nestml synapse models library
NESTML sequence learning network tutorial in the documentation mentions that the model description for the synapse can be found at models/synapses/stdsp_synapse.nestml. However, I couldn't find this synapse model in the library but I could find it on GitHub. Just wanted to bring this to attention and ask if there's a reason that this model is not present in the documentation?

This is the link to the tutorial: https://nestml.readthedocs.io/en/latest/tutorials/sequence_learning/sequence_learning.html

If this is not intentional, then perhaps the stdsp synapse model can be added to the library. Thanks!

clinssen:
Fixes #1269.


clinssen:
Thank you for the review!

PR #1272, opened 2025-11-27
Improved getting of numerical initial values

clinssen:
Internal enhancement. Now also returns numerical values for the internals.

This will be used in the SpiNNaker code generator.


pnbabu:
Could you merge master into this?

PR #1278, opened 2025-12-04
Support random functions for neuromodulated spikes

### LINKED ISSUE #1277: Random number functions do not work when used inside an `onReceive` block for neuromodulated spikes
> Hi, I want to generate a random variable 'rand_u' using random.uniform(0,1) inside the onReceive(mod_spikes) which was set up using a volume transmitter. But the module doesn't compile if I add the rng inside the mod_spikes, update, or equation block. It only works under the onReceive pre_spikes or post_spikes. Do you have any suggestions to make it work? Thanks. 

 _Originally posted by @paruljain1358 in [#1016](https://github.com/nest/nestml/issues/1016#issuecomment-3607378629)_

pnbabu:
Fixes #1277


tests/nest_tests/test_random_functions_neuromod_spikes.py
```diff
@@ -0,0 +1,160 @@
...
+            t_hist, w_hist = self.run_synapse_test(neuron_model_name=self.neuron_model_name,
+                                                   synapse_model_name=self.synapse_model_name,
+                                                   resolution=.1,  # [ms]
+                                                   delay=1.,  # [ms]
+                                                   pre_spike_times=pre_spike_times,
+                                                   post_spike_times=post_spike_times,
+                                                   vt_spike_times=vt_spike_times,
+                                                   use_random_func=use_random_func)
+            if use_random_func:
+                assert np.abs(w_hist[-1] - w_hist[0]) > 1E-3, "Weights should change under this protocol!"
```
clinssen: Could you make this ``assert not np.allclose()`` or ``with pytest.raises(AssertionError): np.testing.assert_allclose()``?

clinssen:
Would you be okay changing the implementation to this? https://github.com/pnbabu/nestml/pull/8 I just wanted to see if this would work, but it looks like it indeed worked out of the box!

clinssen (approved):

PR #1287, opened 2026-01-05
Add GL model tutorial notebook

clinssen:
Minimised version of #1005: keeping the neuron model and single-neuron analysis, but eliminating the facilitating synapses and network model because this was still incomplete/needed more work.

The idea of a minimised version of the tutorial notebook is that we can already present the GL model in a setting that is useful for our end users and will help uptake of the model. Future work on replicating network results is not expected to reveal issues with the NESTML GL model or the NESTML code generation tool, but be an interesting computational neuroscience replication study. This tutorial already demonstrates a working model.


clinssen:
Thank you for the careful and thorough review! I have made the requested adjustments:
- changed citation
- updated units in plots and text
- added and explained "frozen" Poisson noise

Could I ask you to give it one final look? Much obliged!

shimoura (commented):
Besides the small corrections I added, the rest looks fine!

clinssen:
Thanks @shimoura!

PR #1288, opened 2026-01-06
Remove event attributes

### LINKED ISSUE #1286: Re-evaluate concept of synaptic delay
In NEST, synapses have a (dendritic) synaptic delay property (C++ class member). In the development of synapse model support in NESTML, this property was faithfully mirrored as a parameter of the NESTML synapse model. However, from the perspective of the NESTML synapse model, this is a property external to the synapse entity, as delays are managed by the simulator—the synapse knows only about spikes coming into its spiking input ports, and can emit spikes on its spiking output port; the delay is not specified mathematically (as e.g. ``f(t - d)`` instead of ``f(t)``) in the model. With spike event attributes (#1137) being abandoned as a concept, leaving only weighted Dirac delta pulses, only the weight remains as a parameter of the spike event.

To achieve this:
- #1137 should be abandoned; third factors from the presynaptic neuronal partner should come into the synapse via a dedicated spiking or continuous-valued input port of the synapse model.
- Attributes should be removed from output ports (reverting #1124) — done in #1288.
- The predefined function ``emit_spike()`` should only take one optional weight attribute (with a default equal to 1) and not a delay — done in #1288.
- Delay member and getter/setter etc. should be added during code generation — done in #1288.
- Option to make delays a CommonSynapseProperty — done in #1288.

Open questions:
- Should the synapse be able to read out and change its own delay? Should this be the same delay as the NEST dendritic delay? In the future, conceivably we could have a function ``emit_spike(w, t_offset)`` that emits a spike at time ``t + t_offset``. However, this still wouldn't necessarily call for having access to the NEST-internal dendritic delay.

clinssen:
Fixes #1286.

- Remove delay parameter from synapses by default. If desired, synaptic (dendritic) delay from NEST can still be accessed via the code generator option ``delay_variable``.
- Remove attributes from spiking output port; spikes are now "cleanly" defined as Dirac delta pulses. The predefined function ``emit_spikes()`` only takes one parameter now (default: 1) corresponding to the weighting of the delta pulse.


doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -897,26 +897,17 @@ Calling the ``emit_spike()`` function in the ``update`` block results in firing
 Event attributes
 ~~~~~~~~~~~~~~~~
 
-Each spiking output event can be parameterised by one or more attributes. For example, a synapse could assign a weight (as a real number) and delay (in milliseconds) to its spike events by including these values in the call to ``emit_spike()``:
+Each spiking output event corresponds to a Dirac delta pulse and can be parameterised by one attributes (the area of the pulse). For example, a synapse could assign a weight (as a real number) to its spike events by including this value in the call to ``emit_spike()``:
```
pnbabu:
```suggestion
Each spiking output event corresponds to a Dirac delta pulse and can be parameterised by one attribute (the area of the pulse). For example, a synapse could assign a weight (as a real number) to its spike events by including this value in the call to ``emit_spike()``:
```

doc/running/running_nest.rst
```diff
@@ -342,28 +342,58 @@ Simulation of volume-transmitted neuromodulation in NEST can be done using "volu
...
+       onReceive(pre_spikes):
+           print("dendritic delay = {d}")
+
+This will print the string ``dendritic_delay = 2.5``.
+
+
+Synaptic weights
+~~~~~~~~~~~~~~~~~
+
+As synaptic weights are hard-wired into the NEST C++ base class for the NESTML synapse class, a special annotations must be made on the NESTML side to indicate which state variable or parameter corresponds to the weight. To indicate the correspondence, use the code generator option ``weight_variable``. For example, given the following model:
```
pnbabu: We actually do not make any annotation on the NESTML end, but only in the code generation options, right?

pnbabu (changes requested):
Thanks for these changes! I have a few minor suggestions.

doc/running/running_nest.rst
clinssen: Thank you for the suggestion, I have updated the text!

pnbabu (approved):
Looks good to me, thanks!

clinssen:
Thank you for the review!

Issue #976, opened 2023-10-31
Take ODE-toolbox solver recommendation into account


clinssen:
Note that ODE-toolbox since release 3.0.0 returns a conditional list of solvers. NESTML should dynamically pick the solver that is appropriate for the condition that holds.

PR #1291, opened 2026-01-11
Add support for ODE-toolbox 3.0.0

clinssen:
TODO: add support to generate code if a list of solvers is returned. --> spun off to #976.


pynestml/codegeneration/nest_code_generator.py
```diff
@@ -890,6 +913,70 @@ def _get_neuron_model_namespace(self,
 
         return namespace
 
+    def create_ode_toolbox_indict(self, neuron: ASTModel, kernel_buffers: Mapping[ASTKernel, ASTInputPort]):
```
pnbabu: Why are these functions added here? Aren't they part of `ode_toolbox_transformer.py`?

clinssen: I think this changed since the creation time of this PR. Thanks for catching this!

.github/workflows/nestml-build.yml
```diff
@@ -317,6 +321,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip pytest jupyterlab matplotlib pycodestyle scipy pandas
           python -m pip install -r requirements.txt
+          python -m pip install --upgrade git+https://github.com/nest/ode-toolbox
```
pnbabu: Shall we test the release version of `ode-toolbox` here?

pynestml/codegeneration/nest_compartmental_code_generator.py
```diff
@@ -402,11 +404,20 @@ def ode_solve_analytically(self, neuron: ASTModel, parameters_block: ASTBlockWit
...
+        extra_kws = {}
+        if ODEToolboxUtils.is_ode_toolbox_v3_or_higher(odetoolbox):
+            extra_kws["disable_singularity_mitigation"] = True    # multiple conditional solvers returned from ODE-toolbox not yet supported by NESTML
+        else:
+            Logger.log_message(None, None, "Old version of ODE-toolbox used; consider upgrading. ``use_alternative_expM`` flags will be ignored.", None, LoggingLevel.WARNING)
+
+        full_solver_result = odetoolbox.analysis(
             odetoolbox_indict,
             disable_stiffness_check=True,
+            disable_singularity_detection=True,
```
pnbabu, line 409: Should we also define this as a code generation option, and set it to `True` by default?

pynestml/codegeneration/nest_compartmental_code_generator.py
```diff
@@ -449,12 +460,21 @@ def ode_toolbox_analysis(self, neuron: ASTModel,
         if numeric_solvers:
             odetoolbox_indict = self.create_ode_indict(
                 neuron, parameters_block, kernel_buffers)
-            solver_result = analysis(
+            extra_kws = {}
+            if ODEToolboxUtils.is_ode_toolbox_v3_or_higher(odetoolbox):
```
pnbabu: What do you think about fixing the ode-toolbox version to `3.0.0` for future releases?

.github/workflows/nestml-build.yml
clinssen: Yes, I would propose to use the latest release version everywhere except in the "main" CI stage that is the most comprehensive and also uses NEST-main.

pynestml/codegeneration/nest_compartmental_code_generator.py
clinssen: It is indeed already a code generation option, where it gets used by ODEToolboxTransformer. But there are several additional calls to ``odetoolbox.analysis()`` inside the NEST-compartmental code generator. I will email @LeanderEwert to see if it helps the processing speed to disable singularity detection there or make it optional. Thanks for pointing this out.

pynestml/codegeneration/nest_compartmental_code_generator.py
clinssen: Sorry, what do you mean exactly by this—where would you like to fix the version?

pnbabu: I meant in the requirements.txt file. I see you already have a line there, which is commented out right now.

clinssen: Done!

pnbabu: If we're fixing the version of ode-toolbox in the requirements file, do we need to check for version `v3` or higher?

clinssen: That is how it's currently written:
```
odetoolbox >= 3.0.0

```

pnbabu: Sorry, I meant if we need to check this:
```
if ODEToolboxUtils.is_ode_toolbox_v3_or_higher(odetoolbox):
```

clinssen: Thanks for the suggestion—it has been removed!

pnbabu (approved):
Looks good to me, thanks!

clinssen:
Awesome, thanks for the review!

PR #1295, opened 2026-01-14
Disambiguate physical units from declared variables

### LINKED ISSUE #1010: Do not allow physical units to appear in expressions without a numeric literal in front
Currently, physical units can appear in expressions without a numerical literal in front, for example:

```
parameters:
    d ms = 1 ms

onReceive(spikes):
    emit_spike(1, d)
```

will emit a spike with a delay of 1 ms. However, if the user forgets to define the variable or parameter $d$, then NESTML assumes this is the physical unit "days", which has the units of time and thus does not raise any errors or warnings.

@heplesser proposes to allow physical units to appear only directly following a numeric literal, for instance, in case "days" is really what is intended, then:

```
onReceive(spikes):
    emit_spike(1, 1 d)
```

would be the correct way to write that.

## Implementation notes

Add a new CoCo that does the following:

1. look for all variables that are units
2. ignore variables that appear on the lhs (they are part of a declaration); if a variable appears on the rhs:
3. iterate up (while parent is expression, and multiplication/division/exponentiation and lhs is unit or expression containing only units and same for rhs) until you find the whole expression (``ms*S/F/mmol``)
4. for each of these groups: have to ensure that for each, there is a numeric literal in front of it (should be ``42 ms*S/F/mmol``)

### LINKED ISSUE #1260: Assignment to physical unit should trigger error message
The following NESTML model should, but does not currently, trigger an informative error message about the left-hand side not being something that can be assigned to:

```
update:
    V = 1000 mV
```

Thanks to @mlober for reporting this!

clinssen:
Fixes #1260.
Fixes #1010.


doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -220,23 +220,25 @@ Physical units can have at most one of the following magnitude prefixes:
...
-For example, the following defines a possible unit:
+For example, the following is a possible unit:
 
 .. code-block:: nestml
 
-   mV*mV*nS**2/(mS*pA)
+   mV*nS**2*MOhm**-1*pA
 
-Units of the form ``<unit>**-1`` can also be expressed as ``1/<unit>``. For example
+Units of the form ``1/<unit>`` should be expressed as ``<unit>**-1``. For example
```
pnbabu, line 229: Why is this changed to `should be`? Does this mean that units of the form `1/<unit>` are not supported at all?

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -794,6 +796,15 @@ For any two valid numeric expressions ``x``, ``y``, boolean expressions ``b``,\
 | ``?:``                                       | Ternary operator (return ``x`` if ``b`` is true, ``y`` otherwise) | ``b ? x : y``            |
 +----------------------------------------------+-------------------------------------------------------------------+--------------------------+
 
+Numeric literals
+~~~~~~~~~~~~~~~~
+
+Numeric literals with a physical type can appear as constants in any expression, in the form of a number, followed by a physical unit. For example:
```
pnbabu: Numeric literals can be without units as well, right? Am I missing something here?

doc/nestml_language/neurons_in_nestml.rst
```diff
@@ -133,7 +133,7 @@ The incoming spikes could have been equivalently handled with an ``onReceive`` e
        V_m' = -V_m / tau_m + I_syn / C_m
 
    onReceive(spikes):
-       I_syn += spikes * pA * s
+       I_syn += 1 pA * spikes * s
```
pnbabu: Should this be with `1 s` instead of `s`?

pynestml/cocos/co_co_no_assignment_to_unit.py
```diff
@@ -0,0 +1,82 @@
...
+
+class CoCoNoAssignmentToUnitVisitor(ASTVisitor):
+    def visit_assignment(self, node: ASTAssignment):
+        """
+        Ensures the coco.
+        :param node: a single node.
+        """
+        symbol_as_type = node.get_scope().resolve_to_symbol(node.get_variable().get_complete_name(), SymbolKind.TYPE)
+        if not symbol_as_type:
+            # the symbol cannot be resolved as a unit type, so the is definitely no assignment to a unit type here
```
pnbabu:
```suggestion
            # the symbol cannot be resolved as a unit type, so there is definitely no assignment to a unit type here
```

pynestml/cocos/co_cos_manager.py
```diff
@@ -137,6 +139,22 @@ def check_variables_defined_before_usage(cls, model: ASTModel) -> None:
...
+    @classmethod
+    def check_no_assignment_to_unit(cls, model: ASTModel) -> None:
+        """
+        Checks that physical units are not assigned to.
+        :param model: a single model.
+        """
+        CoCoNoAssignmentToUnit.check_co_co(model)
+
+    @classmethod
+    def check_v_comp_requirement(cls, neuron: ASTModel):
```
pnbabu: Why is this added here?

pynestml/visitors/unit_type_fixer_visitor.py
```diff
@@ -0,0 +1,213 @@
...
+from pynestml.symbols.symbol import SymbolKind
+from pynestml.utils.ast_utils import ASTUtils
+from pynestml.visitors.ast_parent_visitor import ASTParentVisitor
+from pynestml.visitors.ast_symbol_table_visitor import ASTSymbolTableVisitor
+from pynestml.visitors.ast_visitor import ASTVisitor
+
+
+class UnitTypeFixerVisitor(ASTVisitor):
+    r"""
+    There can be ambiguities during parsing between what is a variable and what is a unit type. E.g. in an expression ``unit*unit**2*var*var/var``. During parsing, this whole expression gets recognised as an ``ASTUnitType``. This visitor takes variables out of ``ASTUnitType``s.
```
pnbabu, line 40: Could you add an example or two in the docs here showing what is the input and output of this visitor?

tests/nest_tests/resources/test_continuous_third_factor_in_modulatory_spikes_handler_neuron.nestml
```diff
@@ -71,10 +71,10 @@ model test_continuous_third_factor_in_modulatory_spikes_handler_neuron:
             integrate_odes(I_syn_exc, I_syn_inh, V_m)
 
     onReceive(exc_spikes):
-        I_syn_exc += exc_spikes * pA * s
+        I_syn_exc += exc_spikes * 1 pA * 1E3 s
```
pnbabu: Is the factor `1E3` required here?

pnbabu (changes requested):
Thanks for this PR! I have a few minor comments.

doc/nestml_language/nestml_language_concepts.rst
clinssen: Yes, that is correct, it's a consequence of the new grammar rules. It's no longer allowed to write things like "1/pA" as this is deemed confusing; instead, it should be "1 / 1 pA" or "1 pA**-1".

doc/nestml_language/nestml_language_concepts.rst
clinssen: Thanks for pointing this out! I renamed the section to "Numeric literals with a physical unit" and placed it after the definitions of physical units and primitive data types.

doc/nestml_language/neurons_in_nestml.rst
clinssen: Thanks for catching this, fixed!

pynestml/cocos/co_cos_manager.py
clinssen: Oops, that must be from a merge conflict; removed!

pynestml/visitors/unit_type_fixer_visitor.py
clinssen: I have expanded the docstring.

tests/nest_tests/resources/test_continuous_third_factor_in_modulatory_spikes_handler_neuron.nestml
clinssen: Oops, it is not—removed!

pynestml/visitors/unit_type_fixer_visitor.py
```diff
@@ -0,0 +1,236 @@
...
+       parameters:
+           foo real = 42
+
+    the expression
+
+    .. code:: nestml
+
+       42 pA/foo
+
+    will initially be parsed as a numeric literal with a unit type of "pA/foo". However, "foo" actually refers to a parameter in the model. This situation is further exacerbated by NESTML allowing variables to be defined with the same name an units; for instance, there could be a variable defined with the name "mV".
```
pnbabu:
```suggestion
    will initially be parsed as a numeric literal with a unit type of "pA/foo". However, "foo" actually refers to a parameter in the model. This situation is further exacerbated by NESTML allowing variables to be defined with the same name as units; for instance, there could be a variable defined with the name "mV".
```

pnbabu (changes requested):

pnbabu (approved):
Looks good to me, thanks!

PR #1297, opened 2026-01-15
Fix sign error in Wang-Buszaki model

clinssen:
Excitatory synaptic current in this model is all negative. This PR fixes the sign errors in this model: I_syn_exc is (mostly) positive and I_syn_inh is negative.

See also #1265. I would defer further discussion on what the sign conventions should be until we merge #1299.


tests/nest_tests/test_wb_cond_exp_neuron.py
```diff
@@ -0,0 +1,155 @@
...
+            ax[1].set_ylabel("Spikes")
+            ax[0].set_ylabel("v [ms]")
+            ax[0].set_ylim(-100, 50)
+
+            for i in ts:
+                ax[0].axvline(x=i, lw=1., ls="--", color="gray")
+
+            plt.savefig("/tmp/wb_cond_exp.png")
+
+        expected_firing_rate = 50    # [Hz]
```
pnbabu, line 157: Where is this value coming from?

clinssen: I ran it once and counted the spikes, then filled in that value here. Shall I add this as a code comment?

pnbabu (approved):
Thank you!

clinssen:
Thank you for the review!

PR #1299, opened 2026-01-17
Revise input and output ports in NESTML

### LINKED ISSUE #944: Remove "inhibitory" and "excitatory" spiking input port specifiers
Depends on #938.

"Excitatory" and "inhibitory" input port specifiers are confusing and should be eliminated.

Current sitation: if specifiers are present, spikes get "routed" to the right buffer depending on weight of the incoming event.
```
exc_spikes <- excitatory spike
inh_spikes <- inhibitory spike
```
(this example: actually only one in-port (one rport)

Behaviour is such that:
```
w < 0:
	gets routed to inh_spikes
w > 0:
	gets routed to exc_spikes
```

Proposal: change to multisynapse model (behavioural/API change), or preserve current situation by putting the check for the sign of the weight inside the onEvent() handler function.

```
in_spikes <- spikes

onEvent(in_spikes):
	if in_spikes < 0:
		g_syn_inh += 1
	else:
		g_syn_exc += 1
```

clinssen:
Introduces the NESTML predefined ``sift()`` function to process spike events in event handlers. See here for documentation on ``sift()``: https://github.com/nest/nestml/pull/1299/files#diff-22c1d3766680291b0aed3cd548fc0980ecc6511ebb5532a8ed2607105e1775e3R1198

Remove the ``excitatory`` and ``inhibitory`` spiking input port qualifiers. Instead, spikes should be routed (a) on the basis of having multiple spiking input ports (like one for excitatory and one for inhibitory input); or (b) on the basis of a condition like 
```py
if weight < 0: 
    # treat as an inhibitory spike
else:
    # treat as an excitatory spike
```

A consequence of this is that spike ports can no longer be assumed to be linear and time-invariant in the general case (as in this case, a spike from the buffer will be processed differently depending on the sign of its weight). On the C++ side, this means that each spiking input port now uses the ``VectorRingBuffer`` class (https://github.com/nest/nestml/pull/1299/files#diff-ee49404cf3290ce0e9ee40e25068b6d123da357aee83794ab89745a7e2aaf988R1) rather than the ``RingBuffer``. The performance penalty of this is approximately 12.5% in a small balanced network example. An extra code generation option, ``linear_time_invariant_spiking_input_ports``, is provided to instruct the NEST code generator that a set of given ports can be considered LTI. With this option enabled, performance is indistinguishable from the current main branch performance.

<s>Should be merged after #1288.</s> Done.
<s>Should be merged after #1349.</s> Done.
<s>Should be merged after #1358.</s> Done.
<s>Should be merged after #1291.</s> Done.
Should be merged after #1422.

Replaces #1137.

Fixes #944.


pynestml/visitors/ast_builder_visitor.py
```diff
@@ -306,6 +307,9 @@ def visitFunctionCall(self, ctx):
                 args.append(self.visit(arg))
         elif ctx.expression() is not None:
             args.append(self.visit(ctx.expression()))
+        if name.upper() == "SIFT":
+            args[0].set_implicit_conversion_factor(1E3)    # XXX: TODO: the factor 1E3 should be removed; see https://github.com/nest/nestml/issues/984
```
clinssen, line 311: To do: add a link from #984 back to this line of code.

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
    kernel g = delta(t)
 
 
+Input
+-----
+
+External input into the model is said to be received through *input ports*. A NESTML model may contain none, or any number of input ports. Models can receive two distinct types of input: spikes and time-continuous functions.
```
tomtetzlaff:
```suggestion
External input to the model is received through *input ports*. A NESTML model may contain none, or any number of input ports. Models can receive two distinct types of input: spikes and time-continuous functions.
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+
+   \int \delta(t) dt = 1
+
+The unit of the Dirac delta function follows from its definition. Consider the sifting property of the delta function:
+
+.. math::
+
+   f(T) = \int f(t) \delta(t - T) dt
+
+Here :math:`f(t)` is a continuous function of :math:`t`, the value of which is selected at time :math:`t = T`. As the unit of :math:`f()` must be the same on both the left and the right-hand side, the unit of :math:`\delta(t) dt` must be equal to 1. Therefore, the unit of :math:`\delta(t)` must be equal to the inverse of the unit of :math:`t`, that is, the delta function has units :math:`\text{s}^{-1}`. Therefore, all the incoming spikes defined in the input block will implicitly have the unit of :math:`\text{s}^{-1}`. (Note that the physical units of a train of spikes, which is a sum of delta functions, is also :math:`\text{s}^{-1}`.)
```
tomtetzlaff:
```suggestion
Here :math:`f(t)` is a continuous function of :math:`t`, the value of which is selected at time :math:`t = T`. As the unit of :math:`f()` must be the same on both the left and the right-hand side, the unit of :math:`\delta(t) dt` must be equal to 1. Therefore, the unit of :math:`\delta(t)` must be equal to the inverse of the unit of :math:`t`, that is, the delta function has units :math:`\text{s}^{-1}`. Therefore, all the incoming spikes defined in the input block will implicitly have the unit :math:`\text{s}^{-1}`. 
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+   \mathrm{spikes\_pre}(t) = \sum_k w_k \delta(t - t_k)
+
+A corresponding spiking input port can be defined in a NESTML model as follows:
+
+.. code-block:: nestml
+
+   input:
+       spikes_pre <- spike
+
+Note that the units of ``spikes_pre`` are in :math:`\text{s}^{-1}`, as ``w`` has been defined as a dimensionless real number.
```
tomtetzlaff:
```suggestion
Note that the unit of ``spikes_pre`` is :math:`\text{s}^{-1}`, as ``w`` has been defined as a dimensionless real number.
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+
+Spiking input can be processed either by referencing the input port in the right-hand side of an equation (see :ref:`Handling spiking input in equations`) or by means of ``onReceive`` event handlers (see :ref:`Handling spiking input by event handlers`).
+
+
+Handling spiking input in equations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The spiking input port name ``spikes_pre`` can be used directly in the right-hand side of ODEs, for instance as follows:
```
tomtetzlaff:
```suggestion
The spiking input port name ``spikes_pre`` can be used directly on the right-hand side of an ODE, for instance as follows:
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+
+.. math::
+
+   \frac{dx}{dt} = -\frac{x}{\tau} + \mathrm{spikes\_pre}(t)
+
+If ``x`` is a real number, then the units here are consistent (in :math:`\text{s}^{-1}`) on left- and right-hand side of the equation. This can be written in NESTML as:
```
tomtetzlaff:
```suggestion
If ``x`` is a real number, then the units here are consistent (in :math:`\text{s}^{-1}`) on the left- and on the right-hand side of the equation. This can be written in NESTML as:
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+
+.. code-block:: nestml
+
+   x' = -x / tau + spikes_pre
+
+The name of the input port can also be used inside of a convolution. For instance, if ``K`` is a :doc:`kernel <Kernel functions>`, then:
+
+.. math::
+
+   \frac{dx}{dt} = -\frac{x}{\tau} + \frac{1}{C} \left(K \ast \mathrm{spikes\_pre}\right)
```
tomtetzlaff:
```suggestion
   \frac{dx}{dt} = -\frac{x}{\tau} + \frac{1}{C} \left(K \ast \mathrm{spikes\_pre}\right)(t)
```

doc/nestml_language/nestml_language_concepts.rst
```diff
@@ -1071,20 +978,251 @@ A Dirac delta impulse kernel can be defined by using the predefined function ``d
...
+
+.. code-block:: nestml
+
+   onReceive(spikes_pre):
+       println("Info: processing a presynaptic spike at time t = {t}")
+       # ... further statements go here ...
+
+The statements in the event handler will be executed when the event occurs and integrate the state of the system from "just before" the event (at :math:`t-\epsilon`, for :math:`\epsilon\rightarrow 0`) to "just after" the event (at :math:`t=t+\epsilon`). Analogous to the ``update`` block, the predefined variable ``t`` indicates the time :math:`t` at the start of the interval (the function ``timestep()`` may not be used inside an ``onReceive`` block to prevent confusion).
+
+Typically, the statements in the ``onReceive`` block integrate the delta function across time, yielding the surface area under the curve, the value of which corresponds to the weight of the spike. To perform this integration in a concise way, we use the sifting proprty of the delta function. The sifting property states that for any continuous function :math:`f(t)`:
```
tomtetzlaff, line 1097:
```suggestion
Typically, the statements in the ``onReceive`` block integrate the delta function across time, yielding the surface area under the curve, the value of which corresponds to the weight of the spike. To perform this integration in a concise way, we use the sifting property of the delta function. The sifting property states that for any continuous function :math:`f(t)`:
```

tomtetzlaff (approved):
Very nice. Apart from the small typo, this looks very good to me. Thanks a lot for this nice solution.

clinssen:
Thank you very much for the review and the kind thumbs up!

clinssen:
Update per 25-08-2026:

In a meeting with @pnbabu and @abigailm, the following was discussed.

Because "sift" (or "sample") is deemed too much of a signal-processing field vernacular, the name "get_weight" is preferred for this function instead. It is parameterised by the name of the spiking input port and the "current" time ``t``, as in, ``get_weight(exc_spikes, t)``. Although "get_area" and "get_amplitude" were also considered as options, "weight" has the advantage that it refers to the area under the Dirac delta pulse in a mathematically sound way, and is also a commonly used term in neuroscience.

An additional clarification would be to explicitly include the parameter ``t`` in the signature of the ``onReceive`` block, as follows:
```
onReceive(exc_spikes, t):
```

A complete onReceive block could then look as follows:
```
onReceive(exc_spikes, t):
    V_m += unit_psp * get_weight(exc_spikes, t)
```

Although the syntax 
```
onReceive(event from exc_spikes):
    V_m += unit_psp * event.weight
```
was also considered, it should not be necessary to have to create a temporary variable "event", and it was also decided to postpone treatment of spike attributes until a later point in time.

The views of an onReceive block as (a) "exc_spikes is a Platonically ideal train of delta pulses from zero to infinity" or (b) "a spike has arrived; now I need to (procedurally) deal with it" were discussed. The current design suggestion (using sift/sample/get_weight) tends more to (a), whereas it might be more intuitive for our end users to reason about this from (b).

clinssen:
Comment from @tomtetzlaff 2026-09-03: The ``onReceive`` can be thought of as a function that is being called. It is then not clear where the ``t`` comes from (who is the caller?) Furthermore, adding ``t`` as extra parameter in the ``onReceive`` block makes it seem that that already accounts for grabbing area under the curve (integrating the Dirac delta from -epsilon to +epsilon). Thus, he suggests dropping the extra ``t`` parameter from the ``onReceive`` block.

PR #1304, opened 2026-01-20
Allow NEST to read internals

clinssen:
Allow NEST to read internal parameters of the neuron. Add a test for this and for internals being properly updated when parameters are changed.


pnbabu (approved):
Thank you for adding this!

Issue #1232, opened 2025-07-09
Units of delta kernel


clinssen:
Kernels are defined to be unitless (dimensionless). However, a kernel $\delta(t)$ clearly has units 1/s. This is inconsistent.

PR #1307, opened 2026-01-21
Eliminate debug messages from matplotlib in tests

clinssen:
Matplotlib produces a huge amount of debug log messages. This PR changes the matplotlib logger level to WARNING internally, and cleans up the imports.

### DIFF
--- a/tests/nest_tests/test_delta_kernel.p_
+++ b/tests/nest_tests/test_delta_kernel.p_
@@ -25,7 +25,6 @@ import pytest
 
 # try to import matplotlib; set the result in the flag TEST_PLOTS
 try:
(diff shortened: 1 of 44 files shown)


tests/nest_tests/test_delta_kernel.p_
```diff
@@ -19,21 +19,24 @@
 # You should have received a copy of the GNU General Public License
```
pnbabu: Why isn't this a `.py` file?

clinssen: Thanks for catching that! I have spun that off to this existing issue, as I think they are related: #1232.

PR #1311, opened 2026-01-26
Change all user message quotation marks to double quotation marks

### LINKED ISSUE #1120: Change all user message quotation marks to double quotation marks
All user-facing messages should use double rather than single quotation marks. This is to prevent confusion with the single quotation mark character used for differential order. For example:

``Unknown variable: 'g_ahp''``

is more legible as:

``Unknown variable: "g_ahp'"``

clinssen:
Fixes #1120.

### DIFF
--- a/tests/nest_tests/wb_cond_exp_test.py
+++ b/tests/nest_tests/wb_cond_exp_test.py
@@ -28,7 +28,7 @@ try:
     import logging
     import matplotlib as mpl
     mpl.use("agg")
-    logging.getLogger('matplotlib').setLevel(logging.WARNING)    # prevent matplotlib from printing a lot of debug messages when NESTML is running in DEBUG logging_level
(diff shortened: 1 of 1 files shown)


tests/nest_tests/test_integrate_odes.py
```diff
@@ -238,7 +238,7 @@ def test_integrate_odes_params2(self):
 
     def test_integrate_odes_higher_order(self):
         r"""
-        Tests for higher-order ODEs of the form F(x'',x',x)=0, integrate_odes(x) integrates the full dynamics of x.
+        Tests for higher-order ODEs of the form F(x"",x',x)=0, integrate_odes(x) integrates the full dynamics of x.
```
pnbabu:
```suggestion
        Tests for higher-order ODEs of the form F(x'',x',x)=0, integrate_odes(x) integrates the full dynamics of x.
```

pnbabu:
How do we ensure that the future python files have double-quoted strings? Should we add a static check for the same in the CI?

clinssen:
Thank you for the suggestion, I have added a check to the CI using flake8.

PR #1314, opened 2026-02-04
Use GitHub Actions CB instead of bencher for continous benchmarking

clinssen:
Bencher has a user interface that is challenging to work with. A new GitHub Actions integration, [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark), provides the same functionality but is more lightweight and self-contained. It posts a message to the pull request thread with benchmark results (see https://github.com/clinssen/nestml/pull/21 for an example) and renders historical benchmark plots to GitHub Pages (see https://clinssen.github.io/nestml/dev/bench/index.html for an example).


pnbabu:
I cannot see the benchmark run in GitHub Actions for this PR. Does it run only after merging the PR?

PR #1341, opened 2026-03-05
Fix recordable inline expressions in case of vectors in recordablesmap

clinssen:


pnbabu (approved):
Thank you for the fix!

clinssen:
Thank you for the review!

PR #1349, opened 2026-03-27
Introduce unit postsynaptic response constant to clean up expressions in models

clinssen:
Spun off from #1299.

Should be merged after #1350.

Clean up postsynaptic response expressions: instead of multiplying by physical units, like "... * pA", instead define a constant postsynaptic response current/voltage/conductance, e.g. "unit_psc", and multiply by this constant instead.

Also removed some irrelevant comments from the model docstrings, e.g. related to numerics, which do not belong in the model documentation but concern implementation details.


pnbabu (approved):
Looks good to me!

clinssen:
Thank you for the review!

PR #1358, opened 2026-04-14
Refactor finding delta factors

clinssen:
<s>Should be merged after #1364.</s> Done.

Cleanup that moves the part of the code generator that creates the "kernel_buffers" and "delta_factors" data structures into the ConvolutionsToBuffersTransformer. For more documentation about what the transformer does, please see its docstring.

Fix a bug in ASTUtils.get_input_port_by_name() pertaining to vector input ports. For this reason, the magic string that is used to separate port name and vector index was changed from ``_`` into ``__VEC_IDX__``. (This should be mentioned in the release notes!)

As part of this PR, the delta kernel test has been re-enabled.


pynestml/symbols/predefined_types.py
```diff
@@ -258,7 +258,6 @@ def register_unit(cls, unit):
         :type unit: astropy.units.core.Unit
         """
         unit_type = UnitType(str(unit), unit)
-        PredefinedUnits.register_unit(unit_type)
```
pnbabu, line 261: Why is this removed?

clinssen: Unfortunately this ends up adding units like "1 / s" to the predefined units, which shouldn't be—it's a composite unit. This messes things up at a later point when using the PredefinedUnits.

pnbabu (approved):
Looks good to me! Please go ahead and merge this after fixing the merge conflicts.

clinssen:
Thanks very much for the review!

PR #1359, opened 2026-04-14
Fix spinnaker installation instructions

clinssen:


doc/running/running_spinnaker.rst
```diff
@@ -15,31 +15,46 @@ As this solution uses a single bus, the scalability of the system is dependent o
...
 ~~~~~~~~~~~~~~~
 
-1. Build the Apptainer image from https://github.com/nest/nestml/blob/main/extras/spinnaker-apptainer.def
+1. Before building the Apptainer image, make sure there is a ``.spynnaker.cfg`` in the current working directory. Make sure logging options (like ``extract_iobuf``) are enabled. See https://spinnakermanchester.github.io/spynnaker/8.0.0/PyNNOnSpinnakerInstall.html for instructions on how to create this file.
+
+2. Build the Apptainer image from https://github.com/nest/nestml/blob/main/extras/spinnaker-apptainer.def
 
    .. code-block:: bash
 
+
```
krishnakantsingh14: Depending on how Apptainer is installed, `sudo` may be required.

doc/running/running_spinnaker.rst
```diff
@@ -15,31 +15,46 @@ As this solution uses a single bus, the scalability of the system is dependent o
...
       apptainer overlay create --size 4096 spinnaker-overlay.img
 
-2. Run the Apptainer image:
+3. Run the Apptainer image:
+
+   .. code-block:: bash
+
+      apptainer shell --overlay spinnaker-overlay.img spinnaker-apptainer.sif
+
+4. Clone the NESTML repository:
```
krishnakantsingh14, line 35: make sure user is inside the apptainer container env.
`Apptainer> 
`

krishnakantsingh14 (approved):
Looks good 👍

clinssen:
Thank you for the review! I have added a clarification about commands expected to be run inside the apptainer shell. As far as I could see, using sudo in combination with apptainer is considered bad practice, so I would not include that text here.

PR #1360, opened 2026-04-15
Check number of arguments in ``emit_spike()`` function

clinssen:
``emit_spike()`` needs a variadic type for its argument, because it needs to support zero and one arguments (default surface area of the delta of 1, or weighted by the value of a variable). However, it may not have more than one argument. This PR adds a coco to check the number of arguments and fixes some leftovers that should have been done in #1288.


pynestml/cocos/co_co.py
```diff
@@ -26,20 +26,19 @@
...
     @abstractmethod
-    def check_co_co(self, node: ASTNode) -> bool:
+    def check_co_co(cls, node: ASTNode):
         """
-        This is an abstract method which should be implemented by all concrete cocos.
-        :param node: a single neuron instance on which the coco will be checked.
-        :return: True, if CoCo holds, otherwise False.
+        Check the coco.
+
+        :param node: a single model instance on which the coco will be checked.
```
pnbabu:
```suggestion
        :param node: a single ASTNode instance on which the coco will be checked.
```

pnbabu (approved):
Looks good to me. I have a minor comment. Please feel free to ignore it and merge this. 🙂

clinssen:
Thank you for the review!

PR #1363, opened 2026-04-20
Test NEST 3.10 release candidate

clinssen:


clinssen:
@pnbabu: in order to save some compute time on our (already quite heavy) CI runs, I would propose to change the v3.9 run into the new 3.10 run, and to drop the 3.0 run. Although we would still support NEST versions back to and including 3.0, they would just not run on CI, just like all the other intermediate versions (3.1, 3.2, ..., 3.8). Perhaps we can set this up as a CI run triggered by a monthly timer or so, but only check NEST's "latest release" and "main branch" on every NESTML repo push. What do you think?

pnbabu (approved):

PR #1364, opened 2026-04-20
Clean up create_integrate_odes_combinations in utils

clinssen:
Small refactoring and cleanup of the code related to ``create_integrate_odes_combinations``.

Previously, the code was setting ``integrate_odes_combinations`` as a private member variable of the ASTModel instance, which is quite dirty and risks having all kinds of unexpected side effects. Although this can now be changed into using ``metadata``, it turned out even this wasn't necessary and the lines in ``nest_code_generator.py`` could be removed.

### DIFF
--- a/pynestml/utils/ast_utils.py
+++ b/pynestml/utils/ast_utils.py
@@ -1995,14 +1995,14 @@ class ASTUtils:
         return [var.get_name() for var in ASTUtils.get_all_variables_in_expression(expr)]
 
     @classmethod
-    def create_integrate_odes_combinations(cls, model: ASTModel) -> None:
+    def create_integrate_odes_combinations(cls, model: ASTModel) -> List[str]:
         r"""
-        Visit all integrate_odes() calls in the model, compose these as a list of strings, and set them as a model private member (``model.integrate_odes_combinations``).
+        Visit all integrate_odes() calls in the model and compose these as a list of strings.
         """
-        model.integrate_odes_combinations = []
+        integrate_odes_combinations: List[str] = []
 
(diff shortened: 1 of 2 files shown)


pynestml/utils/ast_utils.py
```diff
@@ -1995,14 +1995,14 @@ def get_all_variables_names_in_expression(cls, expr: ASTExpression) -> List[str]
         return [var.get_name() for var in ASTUtils.get_all_variables_in_expression(expr)]
 
     @classmethod
-    def create_integrate_odes_combinations(cls, model: ASTModel) -> None:
+    def create_integrate_odes_combinations(cls, model: ASTModel) -> List[str]:
         r"""
-        Visit all integrate_odes() calls in the model, compose these as a list of strings, and set them as a model private member (``model.integrate_odes_combinations``).
+        Visit all integrate_odes() calls in the model and compose these as a list of strings.
```
pnbabu, line 2000: Do we want to keep this function, given that it's not used? Or do we see any potential future use of the function?

clinssen: It's still being used from inside the Jinja templates.

pnbabu: Oh yes, you are right. I didn't check the occurrence of the function outside of `.py` files, my bad.

pnbabu (approved):

pynestml/utils/ast_utils.py
clinssen: Absolutely not, thanks for double-checking this!

PR #1368, opened 2026-04-24
Code generator performance optimisation

clinssen:
This shaves off about 4 s (roughly 10% of the total) from NESTML code generation for a single STDP synapse test. It takes the main CI run down from about 75 minutes to 60 minutes.

Just as an overview of where time goes during code generation: the total time spent on the single synapse test is 32 s. Of that, 2.5 s is for parsing, 6 s are for the model transformers (especially ODE-toolbox), 8 s are for the actual code generation, and 13 s are for the C++ build. NEST version detection takes about 1 s and was being done twice while it only needs to be done once.


pnbabu (approved):
Thanks for this fix!

clinssen:
Thank you for the review!

PR #1372, opened 2026-04-30
Fix `set_verbosity` calls in Jupyter notebooks 

pnbabu:


clinssen (approved):
Thank you very much for this fix!

PR #1376, opened 2026-05-07
Fix typo in stdp_dopa notebook

pnbabu:


clinssen:
Sorry, but this seems to remove the "Pavlov's dog" image rather than fixing it. Would it be okay to replace this PR by #1377?

pnbabu:
> Sorry, but this seems to remove the "Pavlov's dog" image rather than fixing it. Would it be okay to replace this PR by #1377?

I removed it because the image wasn't rendered in the notebook. Do you want to keep it or maybe replace it with a new image?

clinssen:
> I removed it because the image wasn't rendered in the notebook. Do you want to keep it or maybe replace it with a new image?

It is fixed in #1377.

clinssen:
Closing in favour of #1377.

PR #1380, opened 2026-05-18
Fix GetStatus issue in sequence learning notebook

clinssen:
Fix for the sequence learning notebook after https://github.com/nest/nest-simulator/pull/3829 was merged, which fixed an incorrect implementation of GetStatus() that the notebook was being tested against.


pnbabu (approved):
Thanks for the fix!

PR #1381, opened 2026-05-18
Fix vectors with non-linear ODEs

pnbabu:


clinssen (approved):
Great, thank you for the fix!

PR #1382, opened 2026-05-28
Pre post events cleanup

clinssen:
Clean up some internal functions, like eliminating ``add_suffix_to_variable_names2()``. 

Make a clear distinction between pre_spike_updates and post_spike_updates. Store these separately in the synapse model metadata.

Clean up log messages in synapse-post-neuron-transformer.

Move ``get_trace_at()`` to a separate file so it can be re-used in other tests.

Spun off from #1373.


pynestml/frontend/pynestml_frontend.py
```diff
@@ -110,6 +110,12 @@ def transformers_from_target_name(target_name: str, options: Optional[Mapping[st
                                                                                      "xor", "xor_eq"]})
         transformers.append(variable_name_rewriter)
 
+    if target_name.upper() not in ["NEST_COMPARTMENTAL"]:
```
pnbabu, line 113: Why is only `InlineExpressionExpansionTransformer` moved here?

pnbabu (commented):
Could you switch off the debug logging and run the notebooks again? There's also a `%pdb` in one of the cells which needs to be removed.

pynestml/frontend/pynestml_frontend.py
clinssen: It needs to run before the SynapsePostNeuronTransformer in order to not cause issues with inline expressions appearing in the synapse model. This fixes an issue with inline expressions in synapse models.

clinssen:
Oops, thank you for catching the debugging statements in the notebook! I have cleaned things up, could you have another look? Cheers!

pnbabu (approved):
Looks good, thanks!

clinssen:
Thank you for the review!

PR #1319, opened 2026-02-15
Performance optimisations for the compartmental feature. (WIP)

LeanderEwert:
Various performance optimisations for the cm feature, including streamlined data movement, a more SIMD-friendly variable layout, common sub-expression elimination across multiple expressions, and a more efficient synapse implementation.

PR #1388, opened 2026-06-23
Performance optimizations for the compartmental feature

LeanderEwert:
Various optimisations for the compartmental feature, including:
- Optimised vectorisation
- Common subexpression elimination
- Optimised Hines implementation
- Memory optimizations

This PR should be used for merging going forward. We kindly ask that PR #1319 not be touched until this one is merged, as it is referenced by paper code under review!


tests/nest_compartmental_tests/test__compartmental_multi_synapse_codegen.py
```diff
@@ -245,7 +245,7 @@ def assert_generated_synapse_matches_nest(self, synapse_case):
         if TEST_PLOTS:
             self.plot_comparison(synapse_case, rec_nest_runs, rec_nestml_runs, signed_diff_values)
 
-        assert max(abs_diff_values) <= 0.005, (
+        assert max(abs_diff_values) <= 0.01, (
             "the maximum weight difference is too large! ("
             + str(max(abs_diff_values))
             + " > 0.005)")
```
clinssen: Inconsistent value. Why increase it to 0.01, could you say where this number comes from and why it needed to be increased in this PR?

LeanderEwert: It's because the expression integration order between spike readout/discontinuous state updates in response to spiking and continuous kernel buffer integration has changed. This was inconsistent before, and spikes were essentially read out too early. Compared to before, we now have a slightly larger delay, which leads to a larger difference to the point neuron results. Again, this is one of those cases where we still need to think about the interpretation of our explicit delay, but for now I decided to accept this increased difference since it is expected for this change in order.

pynestml/codegeneration/nest_compartmental_code_generator.py
```diff
@@ -140,6 +144,7 @@ class NESTCompartmentalCodeGenerator(CodeGenerator):
         "weight_variable": {}
     }
 
+    _fp_precision = "double"
```
clinssen, line 154: Can this be a code generator option too?

tests/nest_compartmental_tests/test__compartmental_multi_synapse_codegen.py
LeanderEwert: To clarify this a bit more precisely:
The mechanism contributions are evaluated at half-time steps. This means the integration step is from t-h/2 to t+h/2, so the states at the start of the step are at t-h/2. So when we read out the spikes that have arrived at time t as before they should not actually have arrived yet from the perspective of the mechanism. This means that the order had to be changed such that state advancement happens before the discontinuous spike effect is added.

pynestml/codegeneration/nest_compartmental_code_generator.py
```diff
@@ -666,26 +714,58 @@ def _get_neuron_model_namespace(self, neuron: ASTModel, paired_synapses: Optiona
         neuron.accept(rng_visitor)
         namespace["norm_rng"] = rng_visitor._norm_rng_is_used
 
+        def _suffix_float_literals(expr: str) -> str:
```
clinssen, line 739: What does this do? Could you add a docstring and a little example?

pynestml/codegeneration/printers/cpp_expression_printer.py
```diff
@@ -191,6 +191,25 @@ def _print_arithmetic_operator_expression(self, node: ASTExpressionNode) -> str:
...
+                exponent = rhs_node.get_numeric_literal()
+                if isinstance(exponent, float) and exponent.is_integer():
+                    exponent = int(exponent)
+
+                if isinstance(exponent, int) and 0 <= exponent <= 4:
+                    lhs = self.print(node.get_lhs())
+                    if exponent == 0:
+                        return "1.0"
+                    if exponent == 1:
+                        return "(" + lhs + ")"
```
clinssen, line 205: Please add extra parentheses.

pynestml/codegeneration/printers/cpp_function_call_printer.py
```diff
@@ -39,6 +39,10 @@ class CppFunctionCallPrinter(FunctionCallPrinter):
     Printer for ASTFunctionCall in C++ syntax.
     """
 
+    def __init__(self, expression_printer=None, exp_function: str = "std::exp"):
+        super().__init__(expression_printer)
+        self._exp_function = exp_function
```
clinssen, line 45: Could you add an ``assert(self._exp_function, ["std::exp", "fastexp", ...]``

pynestml/codegeneration/resources_nest_compartmental/cm_neuron/@NEURON_NAME@.h.jinja2
```diff
@@ -276,7 +282,11 @@ private:
   double
   get_state_element( size_t elem )
   {
+{%- if fp_precision == "single" %}
```
clinssen, line 285: Please add a raise exception if "single" is used in the code generator.

pynestml/codegeneration/resources_nest_compartmental/cm_neuron/cm_neuroncurrents_@NEURON_NAME@.h.jinja2
```diff
@@ -470,121 +524,143 @@ private:
...
+                        {% endif %}{% for inline in chan_dep %}, const std::vector< {{ cm_scalar_t }} >& shared_{{inline.variable_name}}{% endfor %}{% if con_in_dep|length %}
+                        {% endif %}{% for inline in con_in_dep %}, const std::vector< {{ cm_scalar_t }} >& shared_{{inline.variable_name}}{% endfor %}
+                        , std::vector< {{ cm_scalar_t }} >& i_val, std::vector< {{ cm_scalar_t }} >& g_val);
     {%- endwith %}
 
     void f_update();
     void f_self_spike();
 
-    void postsynaptic_synaptic_processing();
+  void  postsynaptic_synaptic_processing();
```
clinssen: typo

pynestml/codegeneration/resources_nest_compartmental/cm_neuron/fastexp_v4_@NEURON_NAME@.h.jinja2
```diff
@@ -0,0 +1,71 @@
...
+#include <cmath>
+
+namespace nest
+{
+
+#pragma omp declare simd
+__attribute__( ( always_inline ) ) inline float
+cm_fast_propagator_exp( const float x )
+{
+  // Degree-5 polynomial approximation of exp(x) on [-2, 0].
```
clinssen, line 35: Is there any attribution/citation available for this code?

pynestml/frontend/pynestml_frontend.py
```diff
@@ -450,6 +450,11 @@ def generate_nest_compartmental_target(input_path: Union[str, Sequence[str]], ta
         Enable development mode: code generation is attempted even for models that contain errors, and extra information is rendered in the generated code.
     codegen_opts : Optional[Mapping[str, Any]]
         A dictionary containing additional options for the target code generator.
+    Compartmental-specific options:
```
clinssen: This should be in the documentation for the NESTCompartmentalCodeGenerator class. Then it will appear here: https://nestml.readthedocs.io/en/latest/pynestml.codegeneration.html#pynestml.codegeneration.nest_compartmental_code_generator.NESTCompartmentalCodeGenerator

It could be helpful to additionally add a link to that page from the "NEST Simulator compartmental target" (https://nestml.readthedocs.io/en/latest/running/running_nest_compartmental.html) page.

tests/nest_compartmental_tests/resources/valid/CoCoIntegrateOdesIgnoredInCompartmental.nestml
```diff
@@ -0,0 +1,16 @@
+# CoCoIntegrateOdesIgnoredInCompartmental.nestml
```
clinssen, line 1: please add copyright header

pynestml/codegeneration/nest_compartmental_code_generator.py
LeanderEwert: Added a docstring to the small "post-processing" printer class using it.

pynestml/codegeneration/nest_compartmental_code_generator.py
clinssen: After turning this into a code generation option, can this line be removed?

tests/nest_compartmental_tests/test__compartmental_stdp.py
```diff
@@ -310,4 +317,4 @@ def test__compartmental_stdp(self):
 
         plt.savefig("compartmental_stdp.png")
 
-        assert max(abs_diff_values) <= 0.005, ("the maximum weight difference is too large! (" + str(max(abs_diff_values)) + " > 0.005)")
+        assert max(abs_diff_values) <= 0.01, ("the maximum weight difference is too large! (" + str(max(abs_diff_values)) + " > 0.01)")
```
clinssen, line 320: 0.01 seems like a really large value, could you perhaps add a brief comment on why this is the case in the code?

tests/nest_compartmental_tests/test__fastexp_accuracy.py
```diff
@@ -0,0 +1,435 @@
...
+try:
+    import matplotlib as mpl
+    mpl.use("agg")
+    import matplotlib.pyplot as plt
+    TEST_PLOTS = True
+except BaseException:
+    TEST_PLOTS = False
+
+
+DT = 0.1
```
clinssen, line 42: Ideally, these parameters would go inside the test class (but we can probably leave this for now).

tests/nest_compartmental_tests/test__self_spike_convolutions.py
```diff
@@ -154,7 +154,7 @@ def test_self_spike_convolutions(self):
             "chan_primary0": 0.5945205479701962,
             "chan_secondary0": 0.5945205479701962,
             "concentration0": 0.3589924428735023,
-            "v_comp0": 3.0631281139481406,
+            "v_comp0": 3.132492545007508,
```
clinssen, line 157: Could you briefly comment on why this was changed—what changed behaviourally in the model?

tests/nest_compartmental_tests/test__compartmental_stdp.py
LeanderEwert: The further increased difference is explained in the function docstring. Before, it was just the differing interpretation of delay. Now it also reflects the new integration order between spike arrival and internal state forwarding, which effectively adds another half timestep of delay, whereas before the spike was essentially seen half a timestep too early.

tests/nest_compartmental_tests/test__self_spike_convolutions.py
LeanderEwert: This is also related to the change in integration order. Confusingly, the one channel affected by this change wasn't being recorded, so it looks as if only the voltage changed magically. I am adding it to the recorded and asserted values.

pynestml/codegeneration/resources_nest_compartmental/cm_neuron/fastexp_v4_@NEURON_NAME@.h.jinja2
LeanderEwert: I don't think so. It's a known polynomial approximation method with parameters that Willem obtained himself using an optimiser. The idea, I think, is essentially Taylor expansion + Horner's method. As for its suitability in this context, Willem plans to do further research. :)

Just from a quick search, I think they do something very similar here: 
http://www.federicoperini.info/wp-content/uploads/FastExp-Final.pdf

clinssen (approved):
Thanks for all the work and updates!

I am fine merging this after #1291 has been merged.

pnbabu (approved):
Looks good to me too, thanks!

PR #1391, opened 2026-07-07
Update requirements to add `typing_extensions` instead of `typing`

pnbabu:
The `typing_extensions` module is not installed by default in Python versions < 3.12, resulting in an import error. This PR adds it to the `requirements.txt` file in place of `typing`, since the latter is included in Python versions >= 3.5, which is not a concern for us, as the minimum required Python version is `3.9`.


clinssen (approved):
Thank you for catching this!

PR #1399, opened 2026-07-20
Fix spike integration dependence on whether integrate_odes() is called

### LINKED ISSUE #1392: NESTML iaf_psc_delta ignores absolute refractory period under spiking input
NEST 3.9.0
NESTML 8.2.0
Python 3.12.3

I am using the [iaf_psc_delta](https://github.com/nest/nestml/blob/main/models/neurons/iaf_psc_delta_neuron.nestml) example neuron and find that membrane dynamics and spike timings deviate between NEST and NESTML.

Setup: one Poisson neuron -> two parrot neurons -> each parrot is connected either to one NEST `iaf_psc_delta` neuron or one NESTML `iaf_psc_delta_neuron.nestml` neuron.

**See [debug.py](https://github.com/user-attachments/files/29822023/debug.py)**

Expectation: both neurons should have exactly same membrane dynamics and spike timings.
Finding: the NESTML model ignores the absolute refractory period -- V_m still integrates spikes even though this code in `iaf_psc_delta_neuron.nestml` should prevent that:
```
    update:
        if refr_t > 0 ms:
            # neuron is absolute refractory, do not evolve V_m
            integrate_odes(refr_t)
        else:
            # neuron not refractory
            integrate_odes(V_m)
```

The same does not happen if I use a constant current I_stim.

Image

Image

clinssen:
See #1392.

Spikes are no longer (erroneously) added to the membrane potential of the iaf_psc_delta neuron if ``integrade_odes()`` is not called. The model docstring has been corrected.

Updated NEST/NESTML model equivalence test to check for receiving spikes during refractory period for iaf_psc_delta model to prevent regressions.


pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
```diff
@@ -1478,6 +1478,26 @@ inline void
...
+     *
+     * step 1: regardless of whether and how integrate_odes() will be called, update variables due to convolutions
+    **/
+{% if uses_analytic_solver %}
+{%-     for variable_name in analytic_state_variables: %}
+{%-         if "__X__" in variable_name %}
+{%-             set update_expr = update_expressions[variable_name] %}
+{%-             set var_ast = utils.get_variable_by_name(astnode, variable_name)%}
+{%-             set var_symbol = var_ast.get_scope().resolve_to_symbol(variable_name, SymbolKind.VARIABLE) %}
+{%-             if use_gap_junctions %}
```
pnbabu: Is the code related to gap junctions relevant for synapses?

tests/nest_tests/test_integrate_odes.py
```diff
@@ -234,6 +236,50 @@ def test_integrate_odes_params2(self):
...
+            fig, ax = plt.subplots(nrows=1)
+            ax.plot(times, x_actual, label="x")
+            ax.grid(which="major", axis="both")
+            ax.grid(which="minor", axis="x", linestyle=":", alpha=.4)
+            ax.legend()
+
+            fig.savefig("/tmp/test_integrate_odes_spikes.png", dpi=300)
+
+        # verify
+        np.testing.assert_allclose(x_actual[-1], .004)    # XXX desired should be 4, this is due to an issue with the scale of the weight, see https://github.com/nest/nestml/issues/984
```
pnbabu, line 280: What do you mean by the scale of the weight here?

pynestml/codegeneration/resources_nest/point_neuron/common/SynapseHeader.h.jinja2
clinssen: Thanks for catching this; it has been removed.

tests/nest_tests/test_integrate_odes.py
clinssen: We are integrating 4 spikes, each with a weight of 1, so the desired value should be 4, but due to scaling constants in NEST it comes out as 0.004. This should be resolved when we fix #984.

pnbabu (approved):
Looks good to me, thanks!

clinssen:
Awesome, thanks for the review!

PR #1401, opened 2026-07-23
Add integer division operator and fix float/int division

### LINKED ISSUE #902: Divisions of integers not converted to divisions of doubles in generated C++ code of nestml functions
I encountered an issue with expressions in nestml functions. I had expressions which featured a division of integers, i.e. `4 / 7`, which in the C++ code were not converted to divisions of doubles, i.e. `4./7.`. I feel that this mixes abstraction levels; when writing models in a high-level language, I would expect more pythonic behaviour, and `4 / 7` should return `0.571429` and not `0`. 

Furthermore, I think the issue would be fixed by simply calling `sympy.printing.ccode(sympy.sympify(expr))` with `expr` the expression in the function block, unless there are some subtle aspects of the code generation for nestml functions that I am missing.

clinssen:
Fixes #902.

Introduce Python-syntax, C-semantics integer division operator ``//``.

Allow the multimeter to record integer-type state variables by adding an extra typecasting function (from int to double).

Test that results are as expected for NEST and Python-standalone targets.


pynestml/codegeneration/resources_nest/point_neuron/directives_cpp/MemberVariableGetter.jinja2
```diff
@@ -8,4 +8,11 @@ inline {{ declarations.print_variable_type(variable_symbol) }} get_{{ printer_no
 {
   return {{ nest_codegen_utils.print_symbol_origin(variable_symbol, variable) % printer_no_origin.print(variable) }};
 }
+{%-    if isinstance(variable_symbol.get_type_symbol(), IntegerTypeSymbol) %}
```
pnbabu, line 11: Why is a new function required when the return type (`double`) is the same as the normal `get_` function?

clinssen: The existing getter is of type ``int`` for variables declared as integers in NESTML. I needed a getter that always reliably returns a double.

pnbabu (approved):

clinssen:
Thanks for the review!

PR #1403, opened 2026-07-27
Change compiler flag affecting module unloading behaviour

clinssen:
See https://github.com/nest/nest-simulator/issues/3802 for discussion.

### DIFF
--- a/pynestml/codegeneration/resources_nest/point_neuron/setup/CMakeLists.txt.jinja2
+++ b/pynestml/codegeneration/resources_nest/point_neuron/setup/CMakeLists.txt.jinja2
@@ -254,7 +254,7 @@ if ( BUILD_SHARED_LIBS )
     endif ()
     set_target_properties( ${MODULE_NAME}_module
         PROPERTIES
-        COMPILE_FLAGS "${NEST_CXXFLAGS} -DLTX_MODULE"
+        COMPILE_FLAGS "${NEST_CXXFLAGS} -fno-gnu-unique -DLTX_MODULE"
         LINK_FLAGS "${LINKER_FLAGS}"
         PREFIX ""
         OUTPUT_NAME ${MODULE_NAME} )
--- a/pynestml/codegeneration/resources_nest_compartmental/cm_neuron/setup/CMakeLists.txt.jinja2
+++ b/pynestml/codegeneration/resources_nest_compartmental/cm_neuron/setup/CMakeLists.txt.jinja2
@@ -257,7 +257,7 @@ if ( BUILD_SHARED_LIBS )
     add_library( ${MODULE_NAME}_module MODULE ${MODULE_SOURCES} )
     set_target_properties( ${MODULE_NAME}_module
         PROPERTIES
-        COMPILE_FLAGS "${NEST_CXXFLAGS} -DLTX_MODULE -march=native -ffast-math"
+        COMPILE_FLAGS "${NEST_CXXFLAGS} -fno-gnu-unique -DLTX_MODULE -march=native -ffast-math"
         LINK_FLAGS "${NEST_LIBS}"
         PREFIX ""
         OUTPUT_NAME ${MODULE_NAME} )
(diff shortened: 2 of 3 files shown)


heplesser:
Here suggestions from a discussion with Claude for robust checking for support/lack thereof for the flag:

Here's the standard way to do it, using CMake's `CheckCXXCompilerFlag` module:

```cmake
include(CheckCXXCompilerFlag)

# Some compilers only emit a warning (not a hard error) for an
# unrecognized flag, which would give a false positive. Forcing
# -Werror during the check makes sure unsupported flags actually
# fail the check.
set(CMAKE_REQUIRED_FLAGS "-Werror")
check_cxx_compiler_flag("-fno-gnu-unique" HAVE_FNO_GNU_UNIQUE)
unset(CMAKE_REQUIRED_FLAGS)

if(HAVE_FNO_GNU_UNIQUE)
    target_compile_options(mytarget PRIVATE -fno-gnu-unique)
endif()
```

A few notes on why it's built this way:

**Why `-Werror` in the check.** GCC and Clang generally *do* hard-error on an unknown `-f...` option ("unrecognized command-line option"), so in practice a plain `check_cxx_compiler_flag` call would work fine here. But it's a known footgun with `CheckCXXCompilerFlag` in general that some flags only trigger a warning, which the check would otherwise treat as success. Setting `CMAKE_REQUIRED_FLAGS` to include `-Werror` before the check is the documented workaround, and it's cheap insurance here.

**Save/restore `CMAKE_REQUIRED_FLAGS`.** If you're doing several such checks, wrap it so you don't leak `-Werror` into unrelated checks:

```cmake
include(CheckCXXCompilerFlag)

set(_saved_required_flags "${CMAKE_REQUIRED_FLAGS}")
set(CMAKE_REQUIRED_FLAGS "-Werror")
check_cxx_compiler_flag("-fno-gnu-unique" HAVE_FNO_GNU_UNIQUE)
set(CMAKE_REQUIRED_FLAGS "${_saved_required_flags}")
unset(_saved_required_flags)
```

**If you also compile C code**, use `CheckCCompilerFlag` / `check_c_compiler_flag` the same way and gate C sources with a separate cache variable, since C and C++ compiler drivers can differ (and the flag is really only meaningful for the C++ front end's vague-linkage handling, though GCC accepts it as a general codegen option for both).

**Wrap it in a small helper if you use this pattern a lot:**

```cmake
function(add_compile_option_if_supported target lang flag varname)
    include(Check${lang}CompilerFlag)
    set(_saved "${CMAKE_REQUIRED_FLAGS}")
    set(CMAKE_REQUIRED_FLAGS "-Werror")
    if(lang STREQUAL "CXX")
        check_cxx_compiler_flag("${flag}" ${varname})
    else()
        check_c_compiler_flag("${flag}" ${varname})
    endif()
    set(CMAKE_REQUIRED_FLAGS "${_saved}")
    if(${varname})
        target_compile_options(${target} PRIVATE $<$<COMPILE_LANGUAGE:${lang}>:${flag}>)
    endif()
endfunction()

add_compile_option_if_supported(mytarget CXX "-fno-gnu-unique" HAVE_FNO_GNU_UNIQUE)
```

**Alternative: skip the compile check and gate by platform/compiler ID instead**, since this flag is really only relevant on GNU/Linux ELF targets with GCC or Clang:

```cmake
if((CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang") AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
    target_compile_options(mytarget PRIVATE -fno-gnu-unique)
endif()
```

This is faster (no test compile) but less robust — it doesn't account for edge cases like musl-based Linux toolchains or older GCC/Clang versions where the flag might not exist, and it hardcodes the platform/vendor assumptions this flag depends on. The `check_cxx_compiler_flag` approach is the more portable and future-proof option since it asks the actual compiler rather than guessing from `CMAKE_SYSTEM_NAME`/`CMAKE_CXX_COMPILER_ID`, which is generally the preferred CMake idiom for this kind of "add this flag if the toolchain understands it" logic.

PR #1429, opened 2026-09-15
Fix RNG test

clinssen:
This test was effectively not doing anything. Small maintenance PR.


pynestml/cocos/co_cos_manager.py
```diff
@@ -538,5 +538,6 @@ def check_cocos(cls, model: ASTModel, after_ast_rewrite: bool = False):
         cls.check_resolution_func_legally_used(model)
         cls.check_input_port_size_type(model)
         cls.check_timestep_func_legally_used(model)
+        cls.check_co_co_nest_random_functions_legally_used(model)
```
pnbabu, line 541: I think this coco wasn't added here because this was specific to NEST. It is currently in the `run_nest_target_specific_cocos` function.

clinssen: Thanks for pointing this out as it's indeed in the name.

I believe that the restrictions regarding where calls to rng sampling functions may appear, were initially too wide. For instance, these functions should not appear on the right-hand sides of ODEs; this is based purely on mathematics and should thus hold for all models. This test is fixed in this PR.

However, does NEST have additional restrictions? Was it for instance the case that NEST synapses cannot have state variables or parameters that are initialised by a random value?

pnbabu: These functions cannot be used in state and parameter blocks as well, which was discussed a couple of years ago: https://github.com/nest/nestml/issues/1016

